PR #1, opened 2018-07-26
WIP: Import govuk-frontend

adrianclay:
Work in progress

image


app/views/layouts/application.html.erb
```diff
@@ -1,13 +1,103 @@
...
+  <link rel="mask-icon" href="/assets/images/govuk-mask-icon.svg" color="#0b0c0c">
+  <link rel="apple-touch-icon" sizes="180x180" href="/assets/images/govuk-apple-touch-icon-180x180.png">
+  <link rel="apple-touch-icon" sizes="167x167" href="/assets/images/govuk-apple-touch-icon-167x167.png">
+  <link rel="apple-touch-icon" sizes="152x152" href="/assets/images/govuk-apple-touch-icon-152x152.png">
+  <link rel="apple-touch-icon" href="/assets/images/govuk-apple-touch-icon.png">
 
-  <body>
+  <%= csrf_meta_tags %>
+
+  <%= stylesheet_link_tag    'application', media: 'all' %>
```
adrianclay: We're missing the tags which fixes older browsers.

```
  <!--[if !IE 8]><!-->
  <link href="/govuk-frontend/all.css" rel="stylesheet" />
  <!--<![endif]-->

  <!--[if IE 8]>
    <link href="/govuk-frontend/all-ie8.css" rel="stylesheet" />
  <![endif]-->

  <!--[if lt IE 9]>
    <script src="/html5-shiv/html5shiv.js"></script>
  <![endif]-->
```

app/controllers/basic_controller.rb
```diff
@@ -0,0 +1,5 @@
+class BasicController < ApplicationController
```
adrianclay: Should rename this to be something sensible.

app/assets/stylesheets/application.scss
```diff
@@ -1,16 +1,4 @@
...
- * files in this directory. Styles in this file should be added after the last require_* statement.
- * It is generally better to create a new file per style scope.
- *
- *= require_tree .
- *= require_self
- */
+$govuk-image-url-function: 'image-url';
+$govuk-font-url-function: "font-url";
 
+@import "../../../node_modules/govuk-frontend/all";
```
adrianclay: Be worth doing a quick investigation into which imports we actually require, instead of pulling in everything.

app/assets/stylesheets/application.scss
```diff
@@ -1,16 +1,4 @@
...
- *
- * You're free to add application-wide styles to this file and they'll appear at the bottom of the
- * compiled file so the styles you add here take precedence over styles defined in any other CSS/SCSS
- * files in this directory. Styles in this file should be added after the last require_* statement.
- * It is generally better to create a new file per style scope.
- *
- *= require_tree .
- *= require_self
- */
+$govuk-image-url-function: 'image-url';
```
adrianclay: Documentation for these: https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md#alternative-solution

app/views/layouts/application.html.erb
```diff
@@ -1,13 +1,103 @@
...
+    <div class="govuk-footer__meta">
+      <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
+
+        <svg role="presentation" focusable="false" class="govuk-footer__licence-logo" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 483.2 195.7" height="17" width="41">
+          <path
+            fill="currentColor"
+            d="M421.5 142.8V.1l-50.7 32.3v161.1h112.4v-50.7zm-122.3-9.6A47.12 47.12 0 0 1 221 97.8c0-26 21.1-47.1 47.1-47.1 16.7 0 31.4 8.7 39.7 21.8l42.7-27.2A97.63 97.63 0 0 0 268.1 0c-36.5 0-68.3 20.1-85.1 49.7A98 98 0 0 0 97.8 0C43.9 0 0 43.9 0 97.8s43.9 97.8 97.8 97.8c36.5 0 68.3-20.1 85.1-49.7a97.76 97.76 0 0 0 149.6 25.4l19.4 22.2h3v-87.8h-80l24.3 27.5zM97.8 145c-26 0-47.1-21.1-47.1-47.1s21.1-47.1 47.1-47.1 47.2 21 47.2 47S123.8 145 97.8 145"
+            />
+        </svg>
+        <span class="govuk-footer__licence-description">
```
adrianclay: Ugg the indentation in this file is a mess.

app/assets/stylesheets/application.scss
adrianclay: TODO: Does this make sense for production, do we want to use another method for assets?

kjdchapman:
Closing this in favour of #16.

This PR pursued using the `webpacker` gem for all assets - javascript, stylesheets, images, fonts. Turns out that's pretty tricky!

It was actually reasonably successful for javascript and stylesheets, but images and fonts were another matter.

It's not 100% clear, but I believe the root of the problem was mismatched tools and purpose - for the purposes of this change, we wanted assets to be served on a simple path (as that's what govuk-frontend expects), but `webpacker` wants to emit packs with funky file paths. 

These packs are easy to reference inside our own .js files, but that wasn't the need.

PR #10, opened 2018-07-30
Create initial Jenkinsfile

adrianclay:
Have:

* Added RSpec
* Cleaned up the gems
* Added govuk-linter


docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3.4'
+
+services:
+  app:
+    build: .
+    environment:
+      RACK_ENV: development
```
naliwajek: Do we need `RAILS_ENV` or will Rails pick up `RACK_ENV` as well?

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ruby:2.5-alpine
+ARG BUNDLE_INSTALL_CMD
+ENV RACK_ENV=development
+
+WORKDIR /usr/src/app
+
+COPY Gemfile Gemfile.lock .ruby-version ./
+RUN apk --update --upgrade add build-base sqlite-dev tzdata && \
```
naliwajek: Must remember to remove that `sqlite` dependency when we remove the gem 🐡

docker-compose.yml
adrianclay: Looks like Rails will pick it up:
https://github.com/rails/rails/blob/master/railties/lib/rails.rb#L73

adrianclay:
Will merge for now.  Will address any further feedback in later PRs.

PR #11, opened 2018-07-31
Copy over Jenkinsfile deploying mechanism from other apps

adrianclay:
This will allow us to deploy our app to staging/production from Jenkins.


kjdchapman:
This looks sensible to me, but I think someone more familiar with Jenkins should probably take a look and approve.

PR #12, opened 2018-07-31
Add healthchecking route for ECS autoscaling

adrianclay:
AWS target groups use the healthchecking route to determine whether the app is ready for accepting incoming requests or not.  Once the app is ready, it will be registered with the load balancer to start receiving requests.

At the moment this healthcheck is very simple, as our app doesn't have any complex behaviour.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+ENV['RAILS_ENV'] ||= 'test'
+require File.expand_path('../../config/environment', __FILE__)
+
+abort("The Rails environment is running in production mode!") if Rails.env.production?
+require 'rspec/rails'
+
+ActiveRecord::Migration.maintain_test_schema!
+
+RSpec.configure do |config|
+  config.fixture_path = "#{::Rails.root}/spec/fixtures"
+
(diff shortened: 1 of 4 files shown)


spec/rails_helper.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+ENV['RAILS_ENV'] ||= 'test'
+require File.expand_path('../../config/environment', __FILE__)
+# Prevent database truncation if the environment is production
```
naliwajek: I'd clean up the comments later.

app/controllers/autoscaling_controller.rb
```diff
@@ -0,0 +1,5 @@
+class AutoscalingController < ApplicationController
```
kjdchapman: This name feels a little specific to me - health checks are useful for autoscaling, but also other things (alerts, etc). I'd be inclined to try and reference the user in the module names (and maybe the path too?). I guess the user is `monitoring`, or something thereabouts.

I'm not sure how you've done naming/grouping for controllers in your apps so far, seems like a good opportunity to ask!

This isn't serious enough to stop you from merging.

adrianclay: Agreed.

PR #17, opened 2018-08-02
Signup and Login

tommotaylor:


app/assets/javascripts/html5shiv.js
```diff
@@ -0,0 +1,335 @@
+/**
```
adrianclay: Was committing the `html5shiv` intentional?

db/migrate/20180801132811_create_users.rb
```diff
@@ -0,0 +1,51 @@
...
+
+
+      # Uncomment below if timestamps were not included in your original model.
+      t.timestamps null: false
+    end
+
+    add_index :users, :email,                unique: true
+    add_index :users, :reset_password_token, unique: true
+    add_index :users, :confirmation_token,   unique: true
+    # add_index :users, :unlock_token,         unique: true
```
naliwajek: Why keep comments in migration file? 🤔

app/controllers/users/confirmations_controller.rb
```diff
@@ -0,0 +1,74 @@
+class Users::ConfirmationsController < Devise::ConfirmationsController
+  skip_before_action :authenticate_user!
+
+  # PUT /resource/confirmation
+  def update
+    with_unconfirmed_confirmable do
+      if @confirmable.has_no_password?
```
naliwajek: I guess `@confirmable` is an instance variable defined by Devise and not us hence we lack attr_reader 😢

app/controllers/users/confirmations_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+    end
+    unless @confirmable.errors.empty?
+      self.resource = @confirmable
+      render 'users/confirmations/new'
+    end
+  end
+
+  def pending
+  end
```
naliwajek: Do we need to override this method? Is the original implementation from Devise throwing `NotImplemented` exception otherwise? 🤔

app/controllers/users/confirmations_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  # PUT /resource/confirmation
+  def update
+    with_unconfirmed_confirmable do
+      if @confirmable.has_no_password?
+        @confirmable.attempt_set_password(params[:user])
+        if @confirmable.valid? and passwords_match?
+          do_confirm
+        else
+          do_show
+          @confirmable.errors.clear #so that we wont render :new
```
naliwajek: Is this comment a truth?

We use `@confirmable.errors.clear` as some kind of a hack to not render a `:new` view at all? 
Or is it "so that we won't render confirmable errors in the :new view" ?

Looks like the second case to me though I might not understand how this trick works 😉

app/models/user.rb
```diff
@@ -0,0 +1,36 @@
...
+  # new function to return whether a password has been set
+  def has_no_password?
+    self.encrypted_password.blank?
+  end
+
+  # Devise::Models:unless_confirmed` method doesn't exist in Devise 2.0.0 anymore.
+  # Instead you should use `pending_any_confirmation`.
+  def only_if_unconfirmed
+    pending_any_confirmation {yield}
+  end
```
naliwajek: Hope our acceptance tests will detect such breaking changes in the future 🙂

.rspec
```diff
@@ -1 +1 @@
---require spec_helper
+--require rails_helper
```
naliwajek: Loading whole `rails_helper` for all specs might be an overkill but I guess we can back to it once we start seeing performance hit on tests for POROs 🙂

Gemfile
```diff
@@ -12,11 +12,19 @@ gem 'puma', '~> 3.12'
...
+
+group :test do
+  gem 'capybara'
+  gem 'selenium-webdriver'
+end
 
 group :development, :test do
   gem 'byebug', '~> 10'
   gem 'govuk-lint', '~> 3'
+  gem 'database_cleaner'
```
naliwajek: We need this in dev? 🤔

Gemfile
```diff
@@ -12,11 +12,19 @@ gem 'puma', '~> 3.12'
...
   gem 'rspec-rails', '~> 3'
```
naliwajek: Same here, `rspec-rails` has no role in `development` env AFAIR? 🤔

app/models/user.rb
```diff
@@ -0,0 +1,36 @@
+class User < ApplicationRecord
+  devise :confirmable, :database_authenticatable, :registerable, :recoverable,
+    :rememberable, :trackable, :validatable
+
+  after_create :reset_confirmation_token
```
naliwajek: I'd avoid callbacks by encapsulating them in Use Cases when possible.

config/application.rb
```diff
@@ -6,7 +6,7 @@
 require "active_job/railtie"
 require "active_record/railtie"
 require "action_controller/railtie"
-# require "action_mailer/railtie"
+require "action_mailer/railtie"
 require "action_view/railtie"
 # require "action_cable/engine"
```
naliwajek: We can remove that one?

config/environments/development.rb
```diff
@@ -46,4 +47,12 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  # Paperclip gem
```
naliwajek: Do we use `paperclip` gem? 🤔

app/assets/javascripts/html5shiv.js
kjdchapman: See my comment on #18 - `npm install` brings the file in.

config/environments/development.rb
adrianclay: Don't belive so.

app/views/users/registrations/edit.html.erb
```diff
@@ -0,0 +1,43 @@
+<h2>Edit <%= resource_name.to_s.humanize %></h2>
```
adrianclay: Do we need this view? Is this functionality available at all? I'd dump it if not.

app/controllers/users/confirmations_controller.rb
tommotaylor: @naliwajek `@confirmable` is set in at L53 and L62.

app/controllers/users/confirmations_controller.rb
tommotaylor: @naliwajek This is the "pending confirmation page" that is displayed to the user. It basically says "Check your email". We tell devise to redirect the user after registration here: https://github.com/alphagov/govwifi-admin/pull/17/files#diff-107c009feef83ea12f774836bb3e6a4bR8

app/models/user.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def attempt_set_password(params)
+    unless params[:password] == params[:password_confirmation]
```
emileswarts: Need to be careful we're not comparing two empty strings here.

app/views/users/confirmations/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <%= form_for(resource, as: resource_name, url: users_confirmations_path, html: { method: :put }) do |f| %>
+        <%= devise_error_messages! %>
+        <%= f.hidden_field :confirmation_token, value: params[:confirmation_token] %>
+
+        <div class="govuk-form-group ">
+          <%= f.label :password, "New password", class: "govuk-label" %>
+          <% if @minimum_password_length %>
+            <em>(<%= @minimum_password_length %> characters minimum)</em>
+          <% end %>
+          <%= f.password_field :password, autofocus: true, class: "govuk-input", autocomplete: "off" %>
```
emileswarts: required: true

app/models/user.rb
tommotaylor: Devise has a validation on emails being of a certain length so thats handled elsewhere.

PR #18, opened 2018-08-02
Add html5shiv

tommotaylor:
Extracted out of #17 to fix `master`


kjdchapman:
I believe we misunderstood, and this isn't needed.

There's an html5shiv in `package.json`, which `npm install` will grab, and then it's correctly imported - easier to keep up to date this way.

The perils of using `rails s` instead of `make serve`!

PR #22, opened 2018-08-08
WIP: Verify email is gov domain

kjdchapman:
This is branched off #17, so won't be merged until after it.

This PR implements basic whitelisting on sign up emails - .gov.uk and subdomains are allowed, everyone else isn't.

This was the quickest implementation, there are patterns that beat the very simple string test - not sure how much time it's worth spending on a bulletproof regex (should we lift the one we have from PHP? seems horrible.)


kjdchapman:
Closed in favour of #23, since Jenkins apparently doesn't like `proper_branch_names`

PR #23, opened 2018-08-08
Verify email is in gov domain

kjdchapman:
This is branched off #17, so won't be merged until after it.

This PR implements basic whitelisting on sign up emails - .gov.uk and subdomains are allowed, everyone else isn't.

This was the quickest implementation, there are patterns that beat the very simple string test - not sure how much time it's worth spending on a bulletproof regex (should we lift the one we have from PHP? seems horrible.)

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -1,41 +1,8 @@
 class Users::RegistrationsController < Devise::RegistrationsController
-  alias :rebuild_devise_user :build_resource
-
-  def create
-    if email_matches_whitelist
-      do_default_devise_behaviour
-    else
-      rebuild_devise_user
-      add_error_to_devise_user
-      render 'new'
-    end
-  end
-
 protected
 
   # The path used after sign up for inactive accounts.
   def after_inactive_sign_up_path_for(_resource)
     users_confirmations_pending_path
-  end
-
-  def do_default_devise_behaviour
-    Devise::RegistrationsController
-      .instance_method(:create).bind(self).call
-  end
-
-  def email_matches_whitelist
-    email = params.dig(:user, :email)
-    return false if email.blank?
-
-    domain = email.split('@').last
-    domain.starts_with?('gov.uk') || domain.include?('.gov.uk')
-  end
-
-  def add_error_to_devise_user
-    resource.errors.add(
-      :email,
-      :invalid,
-      message: 'only government/whitelisted emails can sign up'
-    )
   end
 end
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Users::RegistrationsController < Devise::RegistrationsController
 protected
 
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -1,10 +1,41 @@
-# frozen_string_literal: true
-
 class Users::RegistrationsController < Devise::RegistrationsController
+  alias :rebuild_devise_user :build_resource
+
+  def create
+    if email_matches_whitelist
+      do_default_devise_behaviour
+    else
+      rebuild_devise_user
+      add_error_to_devise_user
+      render 'new'
+    end
+  end
+
 protected
 
   # The path used after sign up for inactive accounts.
   def after_inactive_sign_up_path_for(_resource)
     users_confirmations_pending_path
+  end
+
+  def do_default_devise_behaviour
+    Devise::RegistrationsController
+      .instance_method(:create).bind(self).call
+  end
+
+  def email_matches_whitelist
+    email = params.dig(:user, :email)
+    return false if email.blank?
+
+    domain = email.split('@').last
+    domain.starts_with?('gov.uk') || domain.include?('.gov.uk')
+  end
+
+  def add_error_to_devise_user
+    resource.errors.add(
+      :email,
+      :invalid,
+      message: 'only government/whitelisted emails can sign up'
+    )
   end
 end
--- a/app/views/users/confirmations/pending.html.erb
+++ b/app/views/users/confirmations/pending.html.erb
@@ -6,7 +6,7 @@
       validate your e-mail address.
     </p>
     <p class="govuk-body">
-      Simply follow the isntructions in the email.
+      Simply follow the instructions in the email.
     </p>
   </div>
 </div>
--- a/spec/features/logging_in_after_signing_up_spec.rb
+++ b/spec/features/logging_in_after_signing_up_spec.rb
@@ -5,21 +5,22 @@ describe 'logging in after signing up' do
   let(:correct_password) { 'f1uffy-bu44ies' }
 
   before do
-    sign_up_for_account
+    sign_up_for_account(email: 'tom@gov.uk')
     create_password_for_account(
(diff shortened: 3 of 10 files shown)


kjdchapman:
Notably this is ugly right now:

image

Scanning the designs I can't see this case, but I can make it a bit better than this at least!

spec/features/sign_up_as_an_organisation_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(page).to have_content 'Congratulations!'
+    end
+  end
+
+  context 'with a notgov.uk email' do
+    let(:email) { 'someone@notgov.uk' }
+
+    it 'tells me my email is not valid' do
+      expect(page).to have_content(
+        'only government/whitelisted emails can sign up'
```
kjdchapman: Like the `congratulations`, very much a placeholder message.

spec/features/support/sign_up_helpers.rb
```diff
@@ -0,0 +1,23 @@
+def sign_up_for_account(email: 'default@gov.uk')
+  visit new_user_registration_path
+  fill_in 'user_email', with: email
+  click_on 'Sign up'
+end
+
+def create_password_for_account(password: 'supersecret')
+  return unless confirmation_email_received?
```
kjdchapman: Did this to make acceptance tests nicer - can share the simple sign up / create password `before` block - but not sure if it's worth it.

Is this too hidden?

app/controllers/users/registrations_controller.rb
```diff
@@ -1,11 +1,28 @@
-# frozen_string_literal: true
-
 class Users::RegistrationsController < Devise::RegistrationsController
 
+  def create
+    if email_matches_whitelist
+      super
```
kjdchapman: This doesn't feel very expressive.

Wrapping the call in a more expressive method (e.g. `do_default_devise_behaviour`) has some awkwardness with `super`, and also feels a little overkill?

app/controllers/users/registrations_controller.rb
```diff
@@ -1,11 +1,28 @@
...
+    else
+      build_resource
```
kjdchapman: This also feels like a devise-specific thing that might not make sense to a reader. More inclined to pull this out and explain it through method naming.

kjdchapman:
Talking to @tommotaylor irl, decided to leave the "style the error messages in a GDS way" functionality out of this.

It's something we'll need to do for both this and #17, so let's get both merged and then worry about it as a separate task.

app/views/users/confirmations/pending.html.erb
```diff
@@ -6,7 +6,7 @@
       validate your e-mail address.
     </p>
     <p class="govuk-body">
-      Simply follow the isntructions in the email.
+      Simply follow the instructions in the email.
```
adrianclay: I wanted to follow the `isn't tructions`.

spec/features/logging_in_after_signing_up_spec.rb
```diff
@@ -5,29 +5,30 @@
   let(:correct_password) { 'f1uffy-bu44ies' }
 
   before do
-    sign_up_for_account
+    sign_up_for_account(email: 'tom@gov.uk')
```
adrianclay: I forsee conflicts with #25

app/controllers/users/registrations_controller.rb
```diff
@@ -1,10 +1,41 @@
...
   def after_inactive_sign_up_path_for(_resource)
     users_confirmations_pending_path
   end
+
+  def do_default_devise_behaviour
+    Devise::RegistrationsController
+      .instance_method(:create).bind(self).call
+  end
+
+  def email_matches_whitelist
```
adrianclay: Wonder if we can evolve this repo to also have Actors?  Be nice to see this behaviour separated from the Devise fun.

spec/features/logging_in_after_signing_up_spec.rb
kjdchapman: Yeah - not sure how to avoid.

app/controllers/users/registrations_controller.rb
kjdchapman: I considered putting this in use case, but wasn't sure if it was worth it - definitely planning to pull it out when whitelist evolves into a more complex test (or regex, ugh).

Could pull it out now if you feel strongly about it - might help to give some shape to the architecture early on.

app/controllers/users/registrations_controller.rb
```diff
@@ -1,10 +1,41 @@
-# frozen_string_literal: true
-
 class Users::RegistrationsController < Devise::RegistrationsController
+  alias :rebuild_devise_user :build_resource
```
tommotaylor: I would personally have moved this to the model, as it is a `User` validation, and used the Rails `validate` method.

You can either use a method to validate: https://guides.rubyonrails.org/active_record_validations.html#custom-methods or a validator class: https://guides.rubyonrails.org/active_record_validations.html#custom-validators

Then we shouldn't need to override this controller at all, and the validator will add the errors to the unsaved `User` object and set the error in the form.

kjdchapman: I thought about this, but I felt like tying us to a ActiveModel-specific way of doing it might be more pain as this whitelist becomes more complicated. 

Simple validations - making sure something isn't blank - seem fine to sit there, it wouldn't be surprising to the next reader to find them there, and . This one seems more like a _business rule_ than a _validation_ to me, and I like pulling it up out of the model to make it more obvious because of that. I envisioned it being a `use case` in the near-future.

I'm still going back and forth on "if we're using Rails, may as well go full Rails" and "must keep as little in the framework as possible". Be interested to hear what rest of team thinks.

app/controllers/users/registrations_controller.rb
```diff
@@ -1,10 +1,41 @@
...
 protected
 
   # The path used after sign up for inactive accounts.
   def after_inactive_sign_up_path_for(_resource)
     users_confirmations_pending_path
   end
+
+  def do_default_devise_behaviour
+    Devise::RegistrationsController
+      .instance_method(:create).bind(self).call
```
emileswarts: This feels a bit like we're forcing our way through this flow.
Happy to discuss

app/controllers/users/registrations_controller.rb
tommotaylor: Hmm, I still think it belongs on the User model. To me its not a business case its Model validation: "a user cannot be created if it doesn't have a suitable email address". The controller is not a great place for this logic if we are sticking to MVP (regardless of the methods/techniques rails offers).

If this does get complicated, you can still use a `UseCase` for sure, but it should still be called from the Model, eg: `validate :confirm_email_user_case`, where the `def confirm_email_user_case` would call the `UseCase` and return `true` or `false`. But importantly, it belongs on the user because it determines if the user can be created or not.

app/controllers/users/registrations_controller.rb
kjdchapman: Yeah, this might be another indicator that I should validate on the model instead (as [@tommotaylor suggested](https://github.com/alphagov/govwifi-admin/pull/23#discussion_r209229910)) - I'm gonna give that a try.

app/controllers/users/registrations_controller.rb
kjdchapman: Talked about this in person - validator sounds good!

PR #24, opened 2018-08-08
Replace Alpine for Debian in our Dockerfile

adrianclay:
The Alpine docker image has a smaller stack size which results in
SystemStackError from being raised in situations where you have a
deep call stack.

This error was being raised in this repository when running
Capybara tests involving rack-test and SASS compilation.

Switching to Debian should fix this as it doesn't have this stack
size problem.

See more information on this issue here:
https://github.com/docker-library/ruby/issues/196

When switching to Debian, we're pulling in the yarn binary via
their apt repository as default Debian doesn't provide an npm
from which we could use `npm install -g yarn`.

An additional problem is that the default Debian install doesn't
support apt repositories on HTTPS urls.  We first have to install
the `apt-transport-https` before being able to use the yarn repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,16 @@
-FROM ruby:2.5-alpine
+FROM ruby:2.5
 ARG BUNDLE_INSTALL_CMD
 ENV RACK_ENV=development
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


tommotaylor:
Looks good and works for me @adrianclay

PR #26, opened 2018-08-09
Fix loading of the crown image in production

adrianclay:
Currently this image is broken when compiled as it is compiled to
the logical_path as "govuk-logotype-crown.png" whereas our app was
fetching it as "assets/images/govuk-logotype-crown.png"

I believe this is missed in development as we include both the
govuk-frontend and govuk-frontend/assets/images folders so
sprokets can find it using the longer path.


adrianclay:
I'm not happy that we've got problems which can arise between development and production, do any of the team have any thoughts on this?

kjdchapman:
Share your concern, @adrianclay. Would like to deal with assets in the same way across environments to avoid things like this.

Not sure how to achieve that, feels like Rails might fight us.

adrianclay:
I've spoken to @tommotaylor in person, his view is the convention is to always use short paths.  This makes me feel sad, but I'll accept it for now.

PR #27, opened 2018-08-09
Connect to db

emileswarts:


Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'puma', '~> 3.12'
 gem 'rails', '~> 5.1.5'
 gem 'sass-rails', '~> 5.0'
 gem 'sqlite3'
```
adrianclay: Can get rid of the `sqlite3` gem too?

Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM ruby:2.5
 ARG BUNDLE_INSTALL_CMD
-ENV RACK_ENV=development
-ENV DEVISE_SECRET_KEY=fake-secret-key
+
+ENV RACK_ENV development
+ENV DB_USER root
+ENV DB_PASS root
+ENV DB_HOST db
+ENV DEVISE_SECRET_KEY fake-secret-key
```
emileswarts: Wasn't aware that OS ENV variables override this.
Good to know

config/database.yml
```diff
@@ -1,25 +1,23 @@
...
 
-# Warning: The database defined as "test" will be erased and
-# re-generated from your development database when you run "rake".
-# Do not set this db to the same as development or production.
-test:
-  <<: *default
-  database: db/test.sqlite3
+staging:
+  <<: *mysql
+  database: govwifi_admin_staging
```
emileswarts: We need to use env variables here no?

docker-compose.yml
```diff
@@ -1,9 +1,21 @@
...
+    ports:
+      - "13306:3306"
+
   app:
     build: .
     environment:
       RACK_ENV: development
+      DB_USER: root
+      DB_PASS: root
+      DB_HOST: db
```
emileswarts: This needed if we have defaults in the docker file itself?

config/database.yml
naliwajek: Why? Will we have many different database names per environment?

emileswarts: Sure ok yup

PR #29, opened 2018-08-10
Allow user to add an IP

tommotaylor:
Does not include styling, validations or user flow, eg: sign in first. These will be a separate PR.


app/controllers/ips_controller.rb
```diff
@@ -0,0 +1,17 @@
+class IpsController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  def new
+    @ip = Ip.new
+  end
+
+  def create
+    Ip.create(ip_params)
```
emileswarts: maybe create! so it blows up loudly

PR #39, opened 2018-08-13
Generate Radius Secret Key

tommotaylor:


db/schema.rb
```diff
@@ -10,17 +10,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180810103847) do
+ActiveRecord::Schema.define(version: 20180813135333) do
 
-  create_table "ips", force: :cascade do |t|
+  create_table "ips", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
tommotaylor: What is this?

emileswarts: Default Rails 5 behaviour as weird as it may be

Makefile
```diff
@@ -32,6 +32,7 @@ test:
 	$(DOCKER_COMPOSE) up -d db
 	./mysql/bin/wait_for_mysql
 	$(DOCKER_COMPOSE) run -e RACK_ENV=test --rm app ./bin/rails db:create db:schema:load
+	$(DOCKER_COMPOSE) run -e RAILS_ENV=test --rm app ./bin/rails db:migrate
```
emileswarts: Small one but is this a line on its own on purpose?

Could just append it at the end of this command like so:
$(DOCKER_COMPOSE) run -e RACK_ENV=test --rm app ./bin/rails db:create db:schema:load db:migrate

spec/features/add_an_ip_address_spec.rb
```diff
@@ -28,8 +28,8 @@
       expect { click_on 'save' }.to change { Ip.count }.by 1
 
       expect(Ip.first.user).to eq(user)
-      expect(page).to have_content('New IP Added!')
-      expect(page).to have_content(Ip.last.address)
+      expect(page).to have_content('IP Added')
+      expect(page).to have_content(user.radius_secret_key)
```
kjdchapman: We could keep `expect(page).to have_content(Ip.last.address)` as well? Or was there a problem with it?

PR #40, opened 2018-08-14
See Instructions on How to Add IP to Access Point

tommotaylor:
- Links on `ip#show` page (currently go nowhere as tech docs are not ready).
- Add test user to seeds, and seed db in `make serve` (test@gov.uk / "password").

screen shot 2018-08-14 at 09 22 43


db/seeds.rb
```diff
@@ -1,7 +1,5 @@
 # This file should contain all the record creation needed to seed the database with its default values.
```
kjdchapman: I haven't used Rails seeds in a while - is there any risk of this account ending up in the production database?

tommotaylor: If its run in production it will, but we don't have it run in production anywhere, its only called in `make serve`. The Makefile isn't used in any of the deployment stuff is it?

kjdchapman: It's not as far as I know - the `Jenkinsfile` defines the deployment process, and although I'm not Groovy-fluent, it looks like it's doing something completely different. 

Might be worth double-checking with @emileswarts / @naliwajek if they're better at reading `Jenkinsfiles`? :D

PR #42, opened 2018-08-14
Send emails through GovNotify

naliwajek:


spec/lib/notifications/use_case/send_confirmation_email_spec.rb
```diff
@@ -0,0 +1,26 @@
+describe Notifications::UseCase::SendConfirmationEmail do
+  class NotificationsGatewayMock
+    def send(opts)
+      raise unless opts[:email] == 'test@example.com'
+      raise unless opts[:locals][:confirmation_url] == 'https://example.com'
+      raise unless opts[:template_id] == '5f42e490-ce5e-44e7-9104-805136961116'
+      raise unless opts[:reference] == 'confirmation_email'
+
+      {}
```
naliwajek: That's to be changed.

spec/support/confirmation_use_case_spy.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def execute(email:, confirmation_url:)
+    @@last_confirmation_url = confirmation_url
+  end
+
+  private
+
+  attr_reader :notifications_gateway
+end
```
naliwajek: This I want to execute on every spec that uses Devise/Rails.

lib/notifications/use_case/send_confirmation_email.rb
```diff
@@ -0,0 +1,28 @@
+module Notifications
+  module UseCase
+    class SendConfirmationEmail
+      TEMPLATE_ID = '5f42e490-ce5e-44e7-9104-805136961116'
```
emileswarts: Probably stick this in a config yaml file but good enough for now.

lib/notifications/use_case/send_confirmation_email.rb
```diff
@@ -0,0 +1,28 @@
+module Notifications
```
kjdchapman: I quite like the `Actor::UseCase::Thingy` naming convention - wondering if there's something clearer than `Notifications` to be the top-level for this one?

I say this, and I couldn't find something better than `Administrator` for a previous feature.

... maybe this one is `Administrator` too? Or maybe it's `Common`, if "notifications" is a thing the app does for multiple actors?

kjdchapman:
We structured our spec folder a little differently, think we should agree on a structure so things don't get confusing.

We did:
`spec/features` for acceptance tests (might want `spec/features/actor` at some point)
`spec/use_cases/actor` for use case tests

You've done:
`spec/lib/actor/use_cases` for use case
`spec/lib/actor/gateway` for gateway

Do you have a preference? I like some bits from both, would probably make the actors top-level if I was doing it from scratch:
- `spec/actor/features` for acceptance tests
- `spec/actor/use_cases` for use cases, repeat this pattern for gateways etc
- `spec/support` if it's general, `spec/actor/support` if it's actor-specific.

Also if there's any strong feelings from @tommotaylor @emileswarts (my main strong feeling is "make it consistent now")?

naliwajek:
@kjdchapman I don't mind different spec structure – it's like that only because when I was branching off master your `spec/use_case` dir wasn't there yet ¯\_(ツ)_/¯

spec/features/support/sign_up_helpers.rb
```diff
@@ -12,7 +13,5 @@ def create_password_for_account(password: 'supersecret', confirmed_password: 'su
 end
 
 def confirmation_email_link
-  confirmation_email = ActionMailer::Base.deliveries.first
-  parsed_email = Nokogiri::HTML(confirmation_email.body.to_s)
-  parsed_email.css('a').first['href']
+  ConfirmationUseCaseSpy.last_confirmation_url
```
naliwajek: Ha.

config/environments/test.rb
```diff
@@ -33,6 +33,8 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  config.action_mailer.default_url_options = { host: "example.com" }
```
naliwajek: What domain should we use for production?

db/schema.rb
```diff
@@ -10,17 +10,17 @@
...
     t.datetime "updated_at", null: false
     t.integer "user_id"
     t.index ["user_id"], name: "index_ips_on_user_id"
   end
 
-  create_table "users", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
+  create_table "users", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
     t.string "email", default: "", null: false
     t.string "encrypted_password", default: "", null: false
     t.string "reset_password_token"
```
naliwajek: Is this only me or is this OK? Not sure what ActiveRecord is doing here.

kjdchapman: Think it's due to this:
https://stackoverflow.com/questions/27697119/what-is-cascade-in-rails-schema-rb-and-where-did-it-come-from

naliwajek: I'm fine with `force: cascade` but why it changed places on `make build`, why the diff...

kjdchapman: Oh I missed it was just a swap!

:tophat:  Rails :tophat:

config/environments/test.rb
emileswarts: We don't use ActionMailer for any of the emails, could we just remove this?

docker-compose.yml
```diff
@@ -11,5 +11,7 @@ services:
 
   app:
     build: .
+    volumes:
+    - ./:/usr/src/app
```
emileswarts: Not sure we want to commit this.

config/environments/test.rb
naliwajek: No, URL generators n AuthenticationMailer are using this...

spec/features/support/sign_up_helpers.rb
```diff
@@ -13,13 +15,11 @@ def create_password_for_account(password: 'supersecret', confirmed_password: 'su
...
 end
 
 def confirmation_email_received?
-  ActionMailer::Base.deliveries.any?
+  !ConfirmationUseCaseSpy.last_confirmation_url.nil?
```
emileswarts: Could just use .present? here

naliwajek: This way it will work in specs which not load `rails_helper`

PR #43, opened 2018-08-14
See a list of my IPs

tommotaylor:
Without IPs:
screen shot 2018-08-14 at 11 58 54

List of IPs:
screen shot 2018-08-14 at 12 26 12

Add back button to Show:
screen shot 2018-08-14 at 12 11 33


spec/factories/ips.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :ip do
+    address "1.12.123.45.78"
```
emileswarts: A very strange IP, think once we add the validation, these specs will start failing.

PR #44, opened 2018-08-14
See that I have to wait 24 hours

tommotaylor:
Add activation notice to New IP form:

screen shot 2018-08-14 at 14 42 19


kjdchapman:
I'd change the text of the subsection - in fact, probably just the header:

**Your IP will not be available immediately**

I like this as a first slice. I think I'd rather we spent time on making these changes near-instant via _awesome infrastructure magic_ rather than putting a fancy timer next to new IPs or anything.

kjdchapman:
(I've approved because the header text is to my taste, not strictly necessary)

PR #45, opened 2018-08-14
Validate IP Address

tommotaylor:


spec/use_cases/administrator/check_if_valid_ip_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      let(:address) { '192.168.1.15/24' }
+
+      it 'returns true' do
+        result = subject.execute(address)
+        expect(result).to eq(success: true)
+      end
+    end
+
+    context 'with a valid IPv6 address' do
+      let(:address) { '2001:db8:0:1234:0:567:8:1' }
```
kjdchapman: Sadly, RADIUS can't accept IPv6 addresses, so this actually needs to fail!

Expect `IPAddr` has some easy config way to do this.

spec/use_cases/administrator/check_if_valid_ip_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      let(:address) { '172.16.254.1' }
+
+      it 'returns true' do
+        result = subject.execute(address)
+        expect(result).to eq(success: true)
+      end
+    end
+
+    context 'with a valid suffixed IPv4 address' do
+      let(:address) { '192.168.1.15/24' }
```
kjdchapman: Same as IPv6, RADIUS can't accept IP ranges.

In the future it'd be nice to have stuff like, and then the app responsible for sending to RADIUS sorts things out, but feels like we should just flat disallow everything but basic IPs for the mo.

spec/use_cases/administrator/check_if_valid_ip_spec.rb
```diff
@@ -0,0 +1,79 @@
+describe CheckIfValidIp do
+  subject { described_class.new }
+
+  context 'invalid' do
```
kjdchapman: Might be good to test a typical "mistyped and invalid IP" here - I'd expect IPAddr to cover it, but it'll make me sleep easier at night.

`let(:address) { '192.618.0.1' }`

tommotaylor: Done

spec/use_cases/administrator/check_if_valid_ip_spec.rb
tommotaylor: Great catch

lib/use_cases/administrator/check_if_valid_ip.rb
```diff
@@ -0,0 +1,26 @@
+class CheckIfValidIp
+  def execute(address)
+    { success: address_valid_for_radius?(address) }
+  end
+
+private
+
+  def address_valid_for_radius?(address)
+    return false if address.nil?
```
emileswarts: Was wondering if this line is required?
Same with the one below actually, is a subnet a valid ipv4?

PR #49, opened 2018-08-15
Use email templates per environment

naliwajek:
GovNotify has two versions: production (live) and staging. They both
define different sets of templates and API keys. We must use a correct
app for specific env.

### DIFF
--- a/app/mailers/authentication_mailer.rb
+++ b/app/mailers/authentication_mailer.rb
@@ -5,9 +5,14 @@ class AuthenticationMailer < ::Devise::Mailer
 
   def confirmation_instructions(record, token, _opts = {})
     confirmation_link = confirmation_url(record, confirmation_token: token)
+    template_id = GOV_NOTIFY_CONFIG['confirmation_email']['template_id']
 
     SendConfirmationEmail.new(
       notifications_gateway: EmailGateway.new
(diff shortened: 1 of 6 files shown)


app/mailers/authentication_mailer.rb
```diff
@@ -5,9 +5,14 @@ class AuthenticationMailer < ::Devise::Mailer
...
+    template_id = GOV_NOTIFY_CONFIG['confirmation_email']['template_id']
 
     SendConfirmationEmail.new(
       notifications_gateway: EmailGateway.new
-    ).execute(email: record.email, confirmation_url: confirmation_link)
+    ).execute(
+      email: record.email,
+      confirmation_url: confirmation_link,
+      template_id: template_id
+    )
```
naliwajek: We could use `GOV_NOTIFY_CONFIG` directly in the Use Case but that would couple it to Rails. So we don't do it.

config/initializers/gov_notify.rb
```diff
@@ -0,0 +1,3 @@
+GOV_NOTIFY_CONFIG = YAML.load_file(
+  Rails.root.join('config/gov_notify.yml')
+)[Rails.env]
```
kjdchapman: missing line break

kjdchapman:
One tweak to make, but lookin' good!

PR #58, opened 2018-08-16
Add ability to send emails with notifications of new users

naliwajek:
Support (for today's date its Steve) wants to be CC'ed when
new user are registered so he can chase org admins to sign up
for MoU.

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -51,7 +51,17 @@ protected
 
   def confirm_user
     @confirmable.confirm
+    notify_support_of_new_user(@confirmable)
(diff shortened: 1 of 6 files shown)


app/controllers/users/confirmations_controller.rb
```diff
@@ -51,7 +51,17 @@ def render_new_page
 
   def confirm_user
     @confirmable.confirm
+    notify_support_of_mew_user(@confirmable)
```
tommotaylor: Spelling of `new`/`mew`

tommotaylor: No need to pass in `@confirmable`, its available to the method below. Or is this a Clean Architecture style preference?

PR #59, opened 2018-08-16
Add guidance to new IP form

tommotaylor:
Improve IP form to be safer for users:

- Add Errors:
screen shot 2018-08-16 at 11 41 08

- Add Guidance
screen shot 2018-08-16 at 11 40 56


naliwajek:
Looks good. I understand it. Normally I'd wonder is user going to understand it but here users are even more technical than we are.

PR #63, opened 2018-08-21
View RADIUS IPs

kjdchapman:
This PR adds displaying Radius IPs to the "view all my IPs" page. I could add the partial to the other view in this PR, or open a different PR. Taking suggestions.

I've used the [details view from the GOV.UK Design System](https://design-system.service.gov.uk/components/details/) to try and give a little protection to them, might need some discussion.

If the environment variables aren't set, these pages blow up - I think it's better to have that loud failure.

I also want to sort out how "IP" is written in some of the stuff added here - I followed the [Ruby style guide](https://github.com/rubocop-hq/ruby-style-guide#camelcase-classes) without really considering it, and I've realised it's now inconsistent. I don't have a strong preference for what we do apart from consistency.


kjdchapman:
... and immediately I see GOV.UK linting is failing on some of the IP method names. That's that decided then!

kjdchapman:
An image of the new functionality:

image

PR #67, opened 2018-08-23
Ask user for name on password confirmation page

emileswarts:
We want to store additional details for users once they follow the
confirmation link from their inbox.

Rename some of the set_password methods to be more general

### DIFF
--- a/app/views/users/confirmations/show.html.erb
+++ b/app/views/users/confirmations/show.html.erb
@@ -8,9 +8,15 @@
       <%= form_for(resource, as: resource_name, url: users_confirmations_path, html: { method: :put }) do |f| %>
         <%= f.hidden_field :confirmation_token, value: params[:confirmation_token] || params[:user][:confirmation_token] %>
 
+         <%= resource.errors.full_messages %>
+        <div class="govuk-form-group">
+          <%= f.label :name, "Name", class: "govuk-label" %>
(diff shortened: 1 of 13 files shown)


app/views/users/confirmations/show.html.erb
```diff
@@ -8,9 +8,14 @@
       <%= form_for(resource, as: resource_name, url: users_confirmations_path, html: { method: :put }) do |f| %>
         <%= f.hidden_field :confirmation_token, value: params[:confirmation_token] || params[:user][:confirmation_token] %>
 
+        <div class="govuk-form-group <%= "govuk-form-group--error" if resource&.errors&.any? %>">
```
tommotaylor: This is going to set the red error flag on the left hand side of both the password and name fields when only one has an error. 

The guidance is that you add the red flag only to the field in question: https://design-system.service.gov.uk/components/error-message/

We probably need to do something like use the `resource.errors` to identify the field by `id` and add the error class.

spec/models/user_spec.rb
```diff
@@ -7,21 +7,21 @@
     it { should have_many(:ips) }
```
tommotaylor: Could add `it { should validate_presence_of(:name).on(:update) }`

emileswarts: 👍

PR #69, opened 2018-08-23
Improve UX of (what will likely be) landing page

kjdchapman:
Main thing is changes to copy.

Also added a password reset link and more hints (for users who aren't organisations, and around which emails are valid)

image

### DIFF
--- a/app/views/users/registrations/new.html.erb
+++ b/app/views/users/registrations/new.html.erb
@@ -2,27 +2,26 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Welcome!</h1>
-    <p class="govuk-body">
-      We are glad you would like to sign up your organisation for GovWifi.
-    </p>
-    <p class="govuk-body">
-      You have the option to sign up for a three month trial. After you have
-      tested GovWifi, you can go live and scale it at your organisation for free!
+    <h1 class="govuk-heading-l">Sign up your Organisation to GovWifi</h1>
+
+    <p class="govuk-!-padding-bottom-1">
+      <%= link_to "If you are trying to connect to GovWifi as an end user, see instructions here", "", class: "govuk-link govuk-!-font-size-19" %>
     </p>
 
     <div class="govuk-grid-column-two-thirds govuk-!-padding-left-0">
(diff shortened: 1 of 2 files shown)


app/views/users/registrations/new.html.erb
```diff
@@ -2,28 +2,27 @@
...
-    <p class="govuk-body">
-      We are glad you would like to sign up your organisation for GovWifi.
-    </p>
-    <p class="govuk-body">
-      You have the option to sign up for a three month trial. After you have
-      tested GovWifi, you can go live and scale it at your organisation for free!
+    <h1 class="govuk-heading-l">Sign up your Organisation to GovWifi</h1>
+
+    <p class="govuk-!-padding-bottom-1">
+      <%= link_to "If you are trying to connect to GovWifi as an end user, see instructions here", "", class: "govuk-link govuk-!-font-size-19" %>
```
tommotaylor: I would suggest just making the end bit the link. More Govuk'y. Eg:
screen shot 2018-08-23 at 11 52 58

app/views/users/registrations/new.html.erb
```diff
@@ -2,28 +2,27 @@
...
+          <div class="govuk-hint"> Must be a valid government email address </div>
           <%= f.email_field :email, autofocus: true, autocomplete: "email", class: "govuk-input", required: true %>
         </div>
-
         <div class="actions">
           <%= f.submit "Sign up", class: "govuk-button govuk-!-margin-top-4" %>
         </div>
       <% end %>
 
+      <%= link_to "Forgot password", new_password_path(resource_name), class: "govuk-link govuk-!-font-size-19" %><br />
```
tommotaylor: This already exists and can be displayed dynamically from the `users/shared/links` partial.

app/views/users/registrations/new.html.erb
kjdchapman: Good shout:

image

Commit coming shortly..

app/views/users/registrations/new.html.erb
kjdchapman: Yeah, I looked in the partial, saw `controller_name !=` and ran away. 

I'm worried about how much time I have for sorting out these pages today, but I can try and fix this if you think it's best?

kjdchapman: Talked about this IRL and Tom pointed out the very easy solution - nice one Tom!

PR #71, opened 2018-08-23
Create Organisations v2

tommotaylor:
When a user signs up they are asked to enter an Organisation Name which has to be unique.

Refactored the error classes into a helper method for each form field in the app.

screen shot 2018-08-23 at 11 35 12
screen shot 2018-08-23 at 11 35 32

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -7,6 +7,7 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
   def update
     with_unconfirmed_confirmable do
       if @confirmable.attempt_set_password(params[:user])
+        Organisation.create(name: params[:organisation_name], users: [@confirmable])
         confirm_user
       else
         render_show_page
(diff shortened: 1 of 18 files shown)


app/controllers/users/confirmations_controller.rb
```diff
@@ -27,6 +27,17 @@ def pending; end
 
 protected
 
+  def update_user
+    @confirmable.update_details(params[:user])
```
emileswarts: Wonder if we changed the shape of the params, we could get organisation creation for free?

params = {
  user: {
    name: 'bob',
      organisation: {
         name: 'bestorg'
      }
  }
}

tommotaylor: Looks like you can but you have to pass an `Organisation` object, rather than a hash. So we still need the organisation creation step.

I got this working:
```
    organisation = @confirmable.build_organisation(name: params[:organisation_name])
    return true if organisation.save && user.save
```

But I don't know if its an improvement on the origin because we now have to call save on the user again. Before we didn't need to:

```
    organisation = Organisation.new(name: params[:organisation_name], users: [@confirmable])
    return true if organisation.save
```
Thoughts?

PR #73, opened 2018-08-23
Invite a user

tommotaylor:
Basic Invite System.

DOES NOT implement the following:

- Styling of invite pages, including errors.
- Assigning invitee to inviter's organisation.
- Validation / correct errors on invites and registrations from invites


db/migrate/20180823150512_remove_index_users_on_confirmation_token.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveIndexUsersOnConfirmationToken < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :users, column: :confirmation_token
```
emileswarts: Is this because the confirmation token will be blank at that time?

spec/features/team/accept_an_invitation_and_sign_up_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    context "signing up as an invited user" do
+      before do
+        fill_in "Name", with: "Ron Swanson"
+        fill_in "Password", with: "password"
+        fill_in "Password confirmation", with: "password"
+        click_on "Set my password"
+      end
+
+      it "confirms the user" do
+        expect(invited_user.confirmed?).to eq(true)
```
emileswarts: Given these specs are very high level, wonder if we should assert that we are on the page we expect to be after successful signup.  Manage IP addresses page?

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      before do
+        fill_in "Email", with: invited_user_email
+      end
+
+      it "creates an unconfirmed user" do
+        expect {
+          click_on "Send an invitation"
+        }.to change { User.count }.by(1)
+        expect(InviteUseCaseSpy.invite_count).to eq(1)
+        expect(invited_user.confirmed?).to eq(false)
```
kjdchapman: Similar thoughts to Emile here - I wouldn't test a model directly from an acceptance spec, as it's testing internals rather than the external behaviour of the system.

If I wanted to continue the test further (making sure the user I'd signed up isn't confirmed yet), I'd check their email for the confirmation / try visiting the site as them and make sure I can't access things yet.

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,11 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   before_action :authenticate_user!
+  before_action :configure_permitted_parameters, if: :devise_controller?
+
+protected
+
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:accept_invitation, keys: [:name])
```
kjdchapman: I worry that it's easy to find out the _what_ of this line, but not the _why_ - might be tricky to change if you're looking at it and new to the project.

First of all - do _I_ understand the "why"? This change is so that when someone accepts an invitation, they can submit a `name` field to be saved on their new user?

I'm not sure on the best approach to make this more expressive, but I'd probably at least want to change the name of this method to something more specific - maybe `allow_name_parameter_for_invitations` if I've understood the reasons correctly.

tommotaylor: Went with `allow_devise_to_set_attributes`

emileswarts: @kjdchapman Just had a quick chat about the naming of this and feel like `configure_devise_permitted_parameters` is a good name.

It describes what this does within the Rails / Devise domain terminology.

Let me know what you think.

emileswarts: Ok propose we go with this for now and can discuss

tommotaylor: I think we are splitting hairs a bit here, everyone has a valid point. Emile's suggestion seems the middle ground between keeping it technically correct and a very descriptive plain english method name.

spec/features/team/invite_a_team_member_spec.rb
emileswarts: Yeah I looked at this and thought the same.
Felt ok with this as it is looking at the state of the email that gets sent out.  Feel like that could the definition of done for this functionality (ensuring the email is correct).

Not sure how else you could assert this using the UI

kjdchapman: Yeah, I'm happy to look at the e-mail - doesn't have to be the UI that we're confirming, can be anything a user has access to.

Happy to explain more about why I think it's best to avoid testing the models in these specs.

emileswarts: Agreed on not testing models in feature specs.
You got anything in mind that we can use as an assertion here instead?

kjdchapman: For this specific line?:
`expect(invited_user.confirmed?).to eq(false)`

That to me says _I want to check this user has to confirm before they can use the site_.

I'd probably break this out into two specs, one that checks what happens when they login, one that checks they get the right email, something like this:

```ruby
context "with correct data" do
  #snip

  it 'sends the invitee a confirmation email' do
    # check this with spies or similar
  end

  context 'then trying to login as the invitee' do
    before do
      login_as "invitee@gov.uk"
      visit homepage
    end

    it 'tells me to confirm my account first' do
      expect(page).to have_content("confirm your account first")
    end
  end
end
```

PR #74, opened 2018-08-23
Tabify logged-in landing, add placeholder content

kjdchapman:
This PR introduces three tabs - home, locations and team - that
will hopefully help to make it clearer to a user what's happening once
they've logged in.

I think this version of the content is too dense to guide a user, and
I'm not happy with a lot of the language, we can refine it from here.

"Home" part of this: 

image

"Locations" takes what the page used to look like (which will soon get updated as well):

image

Team isn't available yet:

image

Notably tabbed view isn't something GOV.UK are confident in yet, but things can be moved around pretty easily.

### DIFF
--- a/app/views/ips/index.html.erb
+++ b/app/views/ips/index.html.erb
@@ -1,14 +1,80 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <div class="govuk-grid-row">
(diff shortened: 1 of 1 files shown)


kjdchapman:
@emileswarts ish. I don't think this is the _correct_ way to guide people through the service, but it is _better_ than it was. 

Suspect Anne will get a bunch of feedback on how to make it right if she goes armed with this tomorrow.

PR #75, opened 2018-08-23
Disable SSL in development

tommotaylor:


config/environments/development.rb
```diff
@@ -52,5 +52,8 @@
   config.action_mailer.perform_deliveries = false
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.default :charset => "utf-8"
+
+  # Set to false so rails server works in development. True in all other envs
```
kjdchapman: I'd drop this comment - strikes me as:
```ruby
# perform the service
service.perform
```

kjdchapman:
From the Dockerfile:
```Dockerfile
RUN RAILS_ENV=production rails assets:precompile
```

config/environments/development.rb
tommotaylor: Fair

kjdchapman:
My mistake - the `production` flag in the Dockerfile is just for asset precompilation.

PR #78, opened 2018-08-28
Improve UX of add IP page

kjdchapman:
## Summary 
This commit changes various markup and messages to better illustrate to users the purpose and functionality of the now "Add Locations" page. it also adds a functioning 'back' link.

It also modifies the feature specs that were checking against these pages.

Notably this changes the language from "adding IPs" to "adding locations", which I'm not sure if correct. Feels like we have divergent language around this concept.

## New

image

## Old (for reference)

image

### DIFF
--- a/app/views/ips/index.html.erb
+++ b/app/views/ips/index.html.erb
@@ -78,12 +78,12 @@
       <p class="govuk-body">
         You currently have no IPs configured for use with GovWifi. Find out <%= link_to "more", "#", class: "govuk-link" %>
       </p>
-      <%= link_to "Add IP", new_ip_path, class: "govuk-button" %>
+      <%= link_to "Add Location", new_ip_path, class: "govuk-button" %>
     <% else %>
       <%= render "radius_details",
         london_ips: @london_radius_ips,
         dublin_ips: @dublin_radius_ips  %>
-      <%= link_to "Add IP", new_ip_path, class: "govuk-button govuk-!-margin-bottom-0" %>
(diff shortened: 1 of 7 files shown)


app/views/ips/index.html.erb
```diff
@@ -78,12 +78,12 @@
       <p class="govuk-body">
         You currently have no IPs configured for use with GovWifi. Find out <%= link_to "more", "#", class: "govuk-link" %>
       </p>
-      <%= link_to "Add IP", new_ip_path, class: "govuk-button" %>
+      <%= link_to "Add Location", new_ip_path, class: "govuk-button" %>
```
emileswarts: Think we will introduce the concept of locations down the line (next sprint I think).
This will be needed to migrate existing organisations to the admin platform.

Think we should keep the IP naming the way it is until we do that.  Thoughts?

kjdchapman: Mmm, I changed this on Anne's recommendation - think we should probably chat to her tomorrow about it? I don't think I'm clear on the location / IP split either.

I'm happy to change this back to IPs, and it's easy to do so - I'll probably do that until we have that chat, so this doesn't have to wait to get merged.

kjdchapman: I've [switched everything back to IPs](https://github.com/alphagov/govwifi-admin/pull/78/commits/abb69b0a95167487daf99d3795bfaf62dbbe46e0) for now.

PR #79, opened 2018-08-28
Assign Invited User to Organisation

tommotaylor:
Assigns the current user's organisation to the invited user.

I realise I am testing model attributes from a feature spec, however we currently have no indication of the user's organisation in the view, so it kind of has to be here for now. (Unless we want to make some controller specs that is 🤢 )

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -1,0 +1,14 @@
+class Users::InvitationsController < Devise::InvitationsController
+  # Overrides https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L12
+  def new
+    self.resource = User.new(organisation_id: current_user.organisation_id)
+    render :new
+  end
+
+private
+
+  # Overrides https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L105
+  def invite_params
+    params.require(:user).permit(:email, :organisation_id)
+  end
+end
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -1,11 +1,11 @@
 class Users::InvitationsController < Devise::InvitationsController
-  # Overrides https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L12
-  def new
-    self.resource = User.new(organisation_id: current_user.organisation_id)
-    render :new
-  end
+  before_action :add_organisation_to_params, only: :create
 
 private
+
+  def add_organisation_to_params
+    params[:user][:organisation_id] = current_user.organisation_id
+  end
 
   # Overrides https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L105
   def invite_params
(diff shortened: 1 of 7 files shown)


app/views/users/invitations/new.html.erb
```diff
@@ -3,6 +3,8 @@
 <%= form_for resource, :as => resource_name, :url => invitation_path(resource_name), :html => {:method => :post} do |f| %>
   <%= devise_error_messages! %>
 
+  <%= f.hidden_field :organisation_id %>
```
emileswarts: Do we need this?
Given we are newing up the User with the organisation id already assigned.

Just worried that having this in the markup opens up the possibility of someone modifying it.

tommotaylor: Thats a valid concern. We could assign it in the create action, but its not as clean. It would require either implementing a large part of the controller from the gem https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L18

We could do it as a before create action and manually update the params, then call super.... Let me see.

spec/features/team/accept_an_invitation_and_sign_up_spec.rb
```diff
@@ -39,6 +39,7 @@
       it "confirms the user" do
         expect(invited_user.confirmed?).to eq(true)
         expect(invited_user.name).to eq("Ron Swanson")
+        expect(invited_user.organisation).to eq(user.organisation)
```
kjdchapman: This looks like another acceptance test where we're testing internals... but that's not new to this PR, so I'm happy for us to address it all in one go later. I need to look at how these spies are working.

app/controllers/users/invitations_controller.rb
```diff
@@ -0,0 +1,14 @@
+class Users::InvitationsController < Devise::InvitationsController
+  before_action :add_organisation_to_params, only: :create
+
+private
+
+  def add_organisation_to_params
+    params[:user][:organisation_id] = current_user.organisation_id
+  end
+
+  # Overrides https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L105
```
kjdchapman: I wouldn't include this comment - I think the pattern of overriding Devise controllers is common enough that a reader will know what's going on.

Also, unlike before, I'm happy with this override - it's more clear that the intent of this class is to `add_organisation` at the point of inviting.

spec/features/team/accept_an_invitation_and_sign_up_spec.rb
tommotaylor: Yeah I put a little disclaimer in the description above. I assume that soon we'll have an "Organisation Name" displayed in the main view, but until then, the only way to test that this gets assigned would be to implement controller specs, which seems a bit pointless just for this one assertion. Also controller specs generally smell.

kjdchapman: Oh, sorry, I missed that. Or you edited after the fact? Either way, s'cool.

Down with controller specs for sure. Even the rails team doesn't like 'em.

app/controllers/users/invitations_controller.rb
kjdchapman: No worries, I'm happy if you want to include it.

PR #82, opened 2018-08-28
Route53 Healthchecks

emileswarts:
This introduces the Aws SDK for Route53.
The process of establishing whether an IP is healthy requires 2 steps:
  - Get healthcheck Id for that IP, excluding the latency health check
  - Do another API request to establish whether the health check is
  healthy based on the 6 health checkers associated with it

This doesn't display it on the view yet.
There is a failing feature spec for this that isn't a part of this PR.

This creates the use cases and wraps tests around them with fake AWS
gateways.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
+gem 'aws-sdk-route53', '~> 1.0.0.rc13'
 gem 'devise', '~> 4.5.0'
 gem 'devise_invitable', '~> 1.7.0'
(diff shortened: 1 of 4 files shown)


kjdchapman:
At a scan, I like this, but I'm finding the use case itself a little hard to parse. 

I'd like to check it out and play with a little bit to get my head round it, but I'm not sure if I'll get to that today.

I don't have any major complaints if it'd block you to wait until I can get to it (I'm certainly happy if someone else reads and approves).

tommotaylor:
LGTM but like Kyle this is very new code and domain to me.

PR #83, opened 2018-08-28
Redirect back to "List" page when IPs created

kjdchapman:
This PR removes the "show IP" page entirely.

It also does some work on the associated acceptance spec. Pleasingly, without much testing of internals, they didn't require many changes to fit this new flow - good sign!

After I add an IP, it shows me the list page and a success message:

image

Notably, there's no "govuk"-style success message yet, this is a placeholder as @stephenjoe1 has one in the works.

One outstanding issue: Having to link to this specific tab by anchor means the browser window moves down to that anchor, hiding the message. Suggestions welcome!

This also has a the knock-on effect of showing the Devise  "login success" message. Not sure if we want this or not:

image


emileswarts:
Is the radius secret key displayed in the "Show connection details" dropdown?

app/views/ips/_activation_notice.html.erb
```diff
@@ -1,9 +1,6 @@
-<div class="govuk-grid-column-two-thirds">
-  <div class="govuk-inset-text">
-    <h4 class="govuk-heading-s">24 Hours to Activate</h4>
-    <p class="govuk-body">
-      You can configure your access points immedately but it will take 24 hours
-      for your new IP to be activated.
-    </p>
-  </div>
+<div class="govuk-inset-text">
+  <h4 class="govuk-heading-s">New locations are not immediately available</h4>
```
emileswarts: We should stick with IP terminology until we actually introduce locations / sites.

app/views/ips/index.html.erb
```diff
@@ -78,12 +78,12 @@
       <p class="govuk-body">
         You currently have no IPs configured for use with GovWifi. Find out <%= link_to "more", "#", class: "govuk-link" %>
       </p>
-      <%= link_to "Add IP", new_ip_path, class: "govuk-button" %>
+      <%= link_to "Add Location", new_ip_path, class: "govuk-button" %>
```
emileswarts: As above, would call this IP

app/views/layouts/_flash_messages.html.erb
```diff
@@ -0,0 +1,7 @@
+<% flash.each do |name, msg| -%>
+  <div class='flash-holder'>
+    <% if name != 'alert' %>
+      <%= content_tag :p, msg, class: 'govuk-heading-m flash-message' %>
```
emileswarts: Could we derive this class name from the flash 'name'?
Worried that this will be hard to find if we do create an alert flash message

kjdchapman: Yeah, I wasn't sure how to lay this out, felt a bit hokey. 

There's already an error messages partial that Tom put together, I could bring together the two to make things a bit clearer.

I'll take a look at it, but I'm not sure if I should do that as part of a separate PR?

app/views/ips/_activation_notice.html.erb
kjdchapman: Sorry, had branched off the other branch before these changes - fixed it now.

spec/features/add_an_ip_address_spec.rb
```diff
@@ -9,61 +9,64 @@
...
+      end
+
+      it 'shows a success message' do
+        expect(page).to have_content('IP Added')
+      end
 
-      it 'displays the form' do
-        expect(page).to have_content('Enter IP Address (IPv4 only)')
+      it 'shows me a list of IPs including the new IP' do
+        expect(page).to have_content('10.0.0.1')
```
emileswarts: I usually also assert on the current path, could get false positives as the show page would contain this as well

spec/features/add_an_ip_address_spec.rb
```diff
@@ -9,61 +9,64 @@
...
+      context 'when looking back at the list' do
+        before { visit ips_path }
 
-      it 'shows the successful confirmation page' do
-        expect(Ip.first.user).to eq(user)
-        expect(page).to have_content('IP Added')
-        expect(page).to have_content(Ip.first.address)
-        expect(page).to have_content(user.radius_secret_key)
+        it 'has not added the invalid IP' do
+          expect(page).to_not have_content('InvalidIP')
```
emileswarts: Is "InvalidIP" an error that we display to the users?

kjdchapman: No, it's the string we entered in the previous step as an example of an invalid IP. 

Maybe some junk data makes this clearer (e.g. "121.141.121.aaa")?

app/assets/stylesheets/utilities/layout.scss
```diff
@@ -9,3 +9,16 @@
...
   justify-content: space-between;
 }
+
+.flash-holder {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.flash-message {
```
emileswarts: If we modified this class to ".flash-message-notice", we could interpolate the flash :name in the view and easily extend this to have "flash-message-alert" or "flash-message-error" down the line.

Not vital for now but would remove the need to rework it later, your call.

spec/features/add_an_ip_address_spec.rb
kjdchapman: Hmm, I actually don't like to do this - I think the path is an implementation detail, and what a user is really interested in is checking the content on the page.

Checking paths has the same problem as any other internals - can end up brittle, or fooling you into thinking something is safe under test when the _actual_ functionality is missing.

app/assets/stylesheets/utilities/layout.scss
```diff
@@ -9,3 +9,16 @@
...
+  width: 100%;
+  padding: 1em;
+  color: #ffffff;
```
emileswarts: Do we have a css linter?
If not, we should introduce one soon.

Think it fail on this as it can be just "#fff"

spec/features/add_an_ip_address_spec.rb
kjdchapman: If you want to check they're really seeing all the IPs, I happily could add another test for that.

app/assets/stylesheets/utilities/layout.scss
kjdchapman: I'm hoping we don't have to write much custom CSS - I'm not great at it, a linter would probably help!

app/assets/stylesheets/utilities/layout.scss
kjdchapman: I'm happy to change this, I think it makes it clearer that it only handles notices. 

I'm going to look again at how this interacts with the other partial as well, see if I can make something nicer (maybe as a separate PR)

app/assets/stylesheets/utilities/layout.scss
kjdchapman: Looks like we can do it as part of govuk-lint:
https://github.com/alphagov/govuk-lint#sass

spec/features/add_an_ip_address_spec.rb
kjdchapman: Changed this in [this commit](https://github.com/alphagov/govwifi-admin/pull/83/commits/ded50e6ad8e765d89fb5601b7f55e15718564723).

spec/features/add_an_ip_address_spec.rb
kjdchapman: This was simple, so added a new spec [in this commit](https://github.com/alphagov/govwifi-admin/pull/83/commits/912474e69814186219d6755c8b7537dcf5391706).

PR #84, opened 2018-08-28
Style invite pages with errors

tommotaylor:
Validates the user object before it is created and invitation link sent. Devise invitable has a config flag that is meant to validate the user before create, but it doesn't work. Instead I've used a before action to hook into the `Users::Invitations#create` action to call `validate` in the user and pass it back to the view with errors if its invalid.

When resending an invite, the create action is fired again, so we have to check for existing user and skip validation if it exists.

screen shot 2018-08-29 at 12 15 09
screen shot 2018-08-29 at 12 14 58


app/models/user.rb
```diff
@@ -4,7 +4,7 @@ class User < ApplicationRecord
   accepts_nested_attributes_for :organisation
 
   devise :invitable, :confirmable, :database_authenticatable, :registerable, :recoverable,
-    :rememberable, :trackable, :validatable
+    :rememberable, :trackable, :validatable, :validate_on_invite => true
```
emileswarts: Will get linted on this hashrocket

app/controllers/users/invitations_controller.rb
```diff
@@ -1,8 +1,15 @@
 class Users::InvitationsController < Devise::InvitationsController
-  before_action :add_organisation_to_params, only: :create
+  before_action :add_organisation_to_params, :validate_invited_user, only: :create
 
 private
 
+  def validate_invited_user
+    @user = User.new(invite_params)
+    unless @user.validate
+      render :new, resource: @user
```
emileswarts: Is this to ensure that the email is valid?

emileswarts:
LGTM

tommotaylor:
🚢 it

app/controllers/users/invitations_controller.rb
```diff
@@ -1,12 +1,30 @@
...
     params[:user][:organisation_id] = current_user.organisation_id
   end
 
+  def validate_invited_user
+    return if invited_user_already_exists?
+    render_invite_user_page if user_is_invalid?
+  end
+
+  def user_is_invalid?
+    @user = User.new(invite_params)
```
tommotaylor: `@user = User.new(invite_params).valid?`

PR #85, opened 2018-08-29
Allow users to resend invites on pending teammates

naliwajek:
screen shot 2018-08-29 at 11 59 45

### DIFF
--- a/app/assets/stylesheets/utilities/layout.scss
+++ b/app/assets/stylesheets/utilities/layout.scss
@@ -9,3 +9,7 @@
 .justify-content-between {
   justify-content: space-between;
 }
+
+.small-table-button {
+  margin-bottom: 0px;
+}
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/utilities/layout.scss
```diff
@@ -9,3 +9,7 @@
 .justify-content-between {
   justify-content: space-between;
 }
+
+.small-table-button {
+  margin-bottom: 0px;
```
kjdchapman: There are govuk design system classes if you prefer to not write a custom one - have a look [here](https://design-system.service.gov.uk/styles/spacing/).

app/views/team_members/_table.html.erb
```diff
@@ -3,6 +3,21 @@
     <% team_members.each do |member| %>
       <tr class="govuk-table__row">
         <td class="govuk-table__cell "><%= member.email %></td>
+        <td class="govuk-table__cell ">
+	  <% if !member.invitation_sent_at.nil? && !member.invitation_accepted? %>
```
kjdchapman: I wonder if this should be a method on the `user` model instead, feels like it'd be more expressive there.

PR #86, opened 2018-08-29
Get Route53 Healthchecks

emileswarts:
This introduces the Aws SDK for Route53.
The process of establishing whether an IP is healthy requires 2 steps:
  - Get healthcheck Id for that IP, excluding the latency health check
  - Do another API request to establish whether the health check is
  healthy based on the 6 health checkers associated with it

This doesn't display it on the view yet.
There is a failing feature spec for this that isn't a part of this PR.

This creates the use cases and wraps tests around them with fake AWS
gateways.

### DIFF
--- a/app/controllers/health_checks_controller.rb
+++ b/app/controllers/health_checks_controller.rb
@@ -1,0 +1,12 @@
+class HealthChecksController < ApplicationController
+  def index
+    london_ips = ENV.fetch('LONDON_RADIUS_IPS').split(',')
+    dublin_ips = ENV.fetch('DUBLIN_RADIUS_IPS').split(',')
+
+    health_checks_fetcher_use_case = UseCases::Administrator::HealthChecks::Fetcher.new(ips: london_ips + dublin_ips)
+
+    @health_checks = UseCases::Administrator::HealthChecks::CalculateHealth.new(
+      health_checks_fetcher: health_checks_fetcher_use_case
+    ).execute
+  end
+end
--- a/lib/use_cases/administrator/health_checks/fetcher.rb
+++ b/lib/use_cases/administrator/health_checks/fetcher.rb
@@ -1,0 +1,39 @@
+module UseCases
+  module Administrator
+    module HealthChecks
+      class Fetcher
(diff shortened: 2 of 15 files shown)


app/controllers/health_checks_controller.rb
```diff
@@ -0,0 +1,12 @@
+class HealthChecksController < ApplicationController
+  def index
+    london_ips = ENV.fetch('LONDON_RADIUS_IPS').split(',')
+    dublin_ips = ENV.fetch('DUBLIN_RADIUS_IPS').split(',')
+
+    health_checks_fetcher_use_case = UseCases::Administrator::HealthChecks::Fetcher.new(ips: london_ips + dublin_ips)
+
+    @health_checks = UseCases::Administrator::HealthChecks::CalculateHealth.new(
+      health_checks_fetcher: health_checks_fetcher_use_case
+    ).execute
```
naliwajek: This second Use Case seems like a presenter to me i.e its only job is to prepare previously fetched data for display on the page. Could we pass it in as a dependency to the first use case `.execute` method? On the presenter itself you can just have a method like `.present` to get the data in a format you want.

Example: https://github.com/alphagov/govwifi-user-signup-api/blob/master/tasks/publish_statistics.rb#L13

lib/use_cases/administrator/health_checks/fetcher.rb
```diff
@@ -0,0 +1,39 @@
+module UseCases
+  module Administrator
+    module HealthChecks
+      class Fetcher
+        def initialize(route53_gateway: Gateways::Route53.new, ips:)
```
naliwajek: I'd switch those 2 arguments around.

https://github.com/rubocop-hq/ruby-style-guide#optional-arguments

In case of kwargs, `ips:` is required and `route53_gateway:` is our optional argument.

app/controllers/health_checks_controller.rb
naliwajek: In general, second Use Case does not look like a specific business need for me.

PR #88, opened 2018-08-29
Display health checks to the user

emileswarts:
This is unstyled, but uses the use cases created in the previous pr.
We won't be displaying actual health checks in development, so this is
stubbed out in config/environments/development.rb

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -54,5 +54,55 @@ Rails.application.configure do
   config.action_mailer.default :charset => "utf-8"
 
   config.force_ssl = false
+  config.aws_config = {
+    stub_responses: {
+      get_health_check_status: {
+        health_check_observations: [
+          {
+            region: 'ap-southeast-2',
+            ip_address: '39.239.222.111',
+            status_report: {
+              status: 'Success: HTTP Status Code 200, OK'
+            }
+          }
+        ]
+      }, list_health_checks: {
+          max_items: 10,
+          marker: 'PageMarker',
+          is_truncated: false,
+          health_checks: [
+            {
+              caller_reference: 'AdminMonitoring',
+              id: 'abc123',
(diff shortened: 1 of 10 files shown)


app/controllers/monitoring_controller.rb
```diff
@@ -1,7 +0,0 @@
-class MonitoringController < ApplicationController
```
kjdchapman: Whoops, don't think this was meant to go!

config/environments/development.rb
```diff
@@ -54,5 +54,56 @@
   config.action_mailer.default :charset => "utf-8"
 
   config.force_ssl = false
+  config.aws_config = {
+    stub_responses: {
+      get_health_check_status:
+      {
```
kjdchapman: pedantry: hash on the previous line?

spec/gateways/route53_spec.rb
```diff
@@ -0,0 +1,9 @@
+describe Gateways::Route53 do
```
kjdchapman: Are you happy with just these specs? 

It looks like we're just testing the empty case, but not when there's actually stuff for it to retrieve. Or are you happy covering that through the acceptance specs?

emileswarts: Yeah these tests do more than it looks like.
As you can see we reference 'subject.list_health_checks.health_checks', and expect this to be empty.

This is a stubbed out AWS response object that we are asserting on, meaning that it successfully called the API.

Same with 'subject.get_health_check_status(health_check_id: 'abc').health_check_observations', if the delegation didn't happen correctly, it would blow up.

emileswarts:
@kjdchapman Next PR will be around styling of this exclusively.
Might make sense to show what it looks like then.

Currently it's just a lot of text on one line.

PR #93, opened 2018-08-30
Don't filter out latency health checks

emileswarts:
There is no need to filter out latency health checks as they are
actually just an attribute on a regular health check.

This was missed due to changes between staging and production route53
health checks setup.


kjdchapman:
Does the different setup between environments have any impact on viewing these statuses in staging?

emileswarts:
@kjdchapman No the results will be the same on staging and production

PR #98, opened 2018-08-30
Reword healthy/unhealthy to operational/offline

emileswarts:
Improve wording on the GovWifi status page.


kjdchapman:
Might be worth being consistent with other GOV platforms and using **Operational**?
(I need to find out what the opposite is - maybe @stephenjoe1 knows?)

e.g. Notify:
image

GovUK:
image

(sorry, I should have mentioned this earlier!)

PR #100, opened 2018-08-30
Validate password_confirmation at the model

tommotaylor:
Validates the password confirmation correctly at the model. Wasn't working correctly before so we were forcing the controllers to check the two fields matched.

### DIFF
--- a/spec/features/sign_up_as_an_organisation_spec.rb
+++ b/spec/features/sign_up_as_an_organisation_spec.rb
@@ -75,7 +75,7 @@ describe 'Sign up as an organisation' do
     it_behaves_like 'errors in form'
 
     it 'tells the user that the passwords do not match' do
-      expect(page).to have_content 'Password must match confirmation'
+      expect(page).to have_content "Password confirmation doesn't match Password"
     end
   end
 
@@ -155,7 +155,7 @@ describe 'Sign up as an organisation' do
     it_behaves_like 'errors in form'
 
     it 'correctly sets the second error' do
-      expect(page).to have_content 'Password must match confirmation'
+      expect(page).to have_content "Password confirmation doesn't match Password"
     end
   end
 end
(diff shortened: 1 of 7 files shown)


spec/models/user_spec.rb
```diff
@@ -1,47 +1,58 @@
...
-      it 'should not set the users password' do
-        expect { user.update_details(params) }.to_not change(user, :password)
-        expect(user.errors.empty?).to eq(false)
-        expect(user.errors.full_messages).to eq(['Password must match confirmation'])
+      context 'when passwords do not match' do
+        let(:params) { { password: 'new_password', password_confirmation: 'other_password', name: "New name" } }
+
+        it 'should not set the users password' do
+          expect(user.reload.name).to eq("Old name")
+          expect(user.errors.empty?).to eq(false)
```
kjdchapman: I would get rid of these `.empty?` checks - think the lines below cover it enough.

spec/features/sign_up_as_an_organisation_spec.rb
```diff
@@ -75,7 +75,7 @@
     it_behaves_like 'errors in form'
 
     it 'tells the user that the passwords do not match' do
-      expect(page).to have_content 'Password must match confirmation'
+      expect(page).to have_content "Password confirmation doesn't match Password"
```
kjdchapman: The previous text feels more "human" to me - is it easy to tweak the devise translations .yml to change the message?

spec/models/user_spec.rb
```diff
@@ -1,47 +1,58 @@
...
-    let(:user) { create(:user, password: 'password') }
-    context 'when passwords match' do
-      let(:params) { { password: '123456', password_confirmation: '123456', name: 'bob' } }
+  context 'validations' do
+    it { should validate_presence_of(:name).on(:update) }
 
-      it 'should set the users password' do
-        expect { user.update_details(params) }.to change(user, :password)
-        expect(user.errors.empty?).to eq(true)
+    describe "password and password confirmation must match" do
```
kjdchapman: I'd like to change these descriptions slightly, as the nesting means they read a little weirdly to me. I could get back to you with the suggested edit?

It's not a huge deal if you want to press on, though.

PR #105, opened 2018-09-03
Add help page with a contact form

naliwajek:


app/controllers/help_controller.rb
```diff
@@ -0,0 +1,23 @@
+class HelpController < ApplicationController
+  def index
+    @end_user_docs_link = SITE_CONFIG['end_user_docs_link']
+    @organisation_docs_link = SITE_CONFIG['organisation_docs_link']
```
tommotaylor: Shouldn't these just be called directly from the view? Seems strange to set static links in a controller.

app/views/help/index.html.erb
```diff
@@ -0,0 +1,63 @@
...
+        Contact number (optional)
+      </label>
+      <%= text_field_tag 'contact_number', nil, class: 'govuk-input', type: 'tel' %>
+    </div>
+
+    <div class="govuk-form-group">
+      <label class="govuk-label" for="subject">
+        Subject
+      </label>
+      <%= text_field_tag 'subject', nil, class: 'govuk-input', required: true %>
```
tommotaylor: These should be in a `two-thirds` or `three-quarters` width column. The guidance recommends against full width text inputs.

app/controllers/help_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  def create
+    template_id = GOV_NOTIFY_CONFIG['help_email']['template_id']
+
+    SendHelpEmail.new(
+      notifications_gateway: EmailGateway.new
+    ).execute(
+      email: GOV_NOTIFY_CONFIG['support_email'],
+      sender_email: current_user.email,
+      name: current_user.name,
+      organisation: current_user.organisation.name,
```
tommotaylor: Can use `current_organisation` here.

app/views/help/index.html.erb
```diff
@@ -0,0 +1,63 @@
...
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li><%= link_to 'Check servers status', status_index_path %></li>
+      <li><%= link_to 'Browse technical documentation', @organisation_docs_link %></li>
+    </ul>
+  <p>
+<div>
+
+<%= form_tag('/help') do %>
+  <div>
```
tommotaylor: Could probably do with a little margin above here to separate the "Contact Us" title from the content above. Perhaps a `govuk-grid-row`? https://design-system.service.gov.uk/styles/layout/

app/views/help/index.html.erb
tommotaylor: https://design-system.service.gov.uk/styles/layout/

app/controllers/help_controller.rb
naliwajek: `SITE_CONFIG` a constant created during initialization of Rails. Feels decoupling it from what is passed to a view here is a right thing to do. ERB doesn't need to know Rails constants.

app/controllers/help_controller.rb
naliwajek: Ah, cool lil' helper 👍

PR #106, opened 2018-09-03
Improve layout/messaging of IP Addresses page (+ other minor copy/layout tweaks across app)

kjdchapman:
image

image

This PR separates out the "add an IP" and "configure your access points" tasks a little better. Planning to move them to separate screens in the near-future, based on Anna/Constance's feedback.

Adds a warning about IPs not being immediately available to the messaging feedback and some other scattered bits of copy.

### DIFF
--- a/app/controllers/ips_controller.rb
+++ b/app/controllers/ips_controller.rb
@@ -6,7 +6,7 @@ class IpsController < ApplicationController
   def create
     @ip = current_organisation.ips.new(ip_params)
     if @ip.save
-      flash[:notice] = 'IP Added: ' + @ip.address
+      flash[:notice] = "#{@ip.address} added, it will be active starting tomorrow"
       redirect_to ips_path(anchor: 'ips')
     else
       render :new
--- a/app/controllers/ips_controller.rb
+++ b/app/controllers/ips_controller.rb
@@ -6,8 +6,11 @@ class IpsController < ApplicationController
   def create
     @ip = current_organisation.ips.new(ip_params)
     if @ip.save
-      flash[:notice] = "#{@ip.address} added, it will be active starting tomorrow"
-      redirect_to ips_path(anchor: 'ips')
+      redirect_to(
+        ips_path,
+        anchor: 'ips',
+        notice: "#{@ip.address} added, it will be active starting tomorrow"
+      )
     else
       render :new
     end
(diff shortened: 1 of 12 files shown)


app/views/team_members/_table.html.erb
```diff
@@ -4,20 +4,19 @@
       <tr class="govuk-table__row">
         <td class="govuk-table__cell "><%= member.email %></td>
         <td class="govuk-table__cell ">
-	  <% if !member.invitation_sent_at.nil? && !member.invitation_accepted? %>
-            invitation pending
-	  <% end %>
-	</td>
+          <% if !member.invitation_sent_at.nil? && !member.invitation_accepted? %>
```
tommotaylor: `!smthng.nil?` is better expressed as `smthng.present?`or just `if member.invitation_sent_at`. Although in this case I would almost be tempted to drop a method on the `User` model:
```
  def invitation_pending?
    invitation_sent_at && !invitation_accepted?
  end
```
Then you just use `if member.invitation_pending?`

app/controllers/ips_controller.rb
```diff
@@ -6,7 +6,7 @@ def new
   def create
     @ip = current_organisation.ips.new(ip_params)
     if @ip.save
-      flash[:notice] = 'IP Added: ' + @ip.address
+      flash[:notice] = "#{@ip.address} added, it will be active starting tomorrow"
       redirect_to ips_path(anchor: 'ips')
```
tommotaylor: You can one line this if it will fit, but it may be too long a line. 
```
  redirect_to ips_path(anchor: 'ips'), notice: "#{@ip.address} added, it will be active starting tomorrow"
```

app/views/team_members/_table.html.erb
kjdchapman: I actually just changed the tabbing of this bit! 

I'm happy to sort this out whilst I'm here, if it's not making the PR too big.

app/controllers/ips_controller.rb
kjdchapman: Same as [above](https://github.com/alphagov/govwifi-admin/pull/106#discussion_r214708023).

app/views/team_members/_table.html.erb
kjdchapman: Fixed in https://github.com/alphagov/govwifi-admin/pull/106/commits/ce4e5cbcce83542aa7c33a86fec12b705a0b636d

app/controllers/ips_controller.rb
kjdchapman: Fixed in https://github.com/alphagov/govwifi-admin/pull/106/commits/aea325075babddb8d5918ceecc43df08484f8839

PR #108, opened 2018-09-03
Give users a way to escape invitations flow

kjdchapman:
image

Makes this consistent with our other form-style pages (and GOV in general)


tommotaylor:
LGTM, I think maybe just "back" would be better than "back to list" as there will likely be a lot of accounts that don't have team members, so the list never exists. But just a thought, no problem with how it is.

kjdchapman:
I agree with you - Back is better. Just copied it without engaging my brain.

PR #110, opened 2018-09-03
Give users a way to escape invitations flow

kjdchapman:
(Renamed branch due to jenkins tyranny.)


kjdchapman:
(and it didn't even work.)

PR #121, opened 2018-09-07
WIP: Add locations to db

tommotaylor:
This is a spike to add locations to the admin portal. I have approached this from the angle of trying to get the schema ready to accept the old system data, without changing the frontend user flow. Its the smallest slice I could think of. 

Mainly this PR is about the schema and association changes.

But theres some obvious hacks that I will have to remove in the next iteration:
- For a new user, I'm creating a location when they create their organisation.
- Every IP they add is added to this location.
- I've done this so everything works until we modify the frontend flow.

OLD SYSTEM DATA: We need to import the existing data, which has locations, to this DB. So I've looked at the following schema to bring things over. Some names will need to be changed. I think its better to have nice names than follow the old conventions: https://github.com/alphagov/govwifi-allowed-sites-api/blob/master/mysql/schema.sql#L310

ALLOWED SITES: Will need changing as the radius secret key is now on location. Just realised I've gone with "locations" over "sites" as well.

WHAT TO DO WITH THE USER FLOW ON THE ADMIN PORTAL???

I think we are better off with locations than without. We've heard this from Steve, Milan and user testing. I think the following basic location => IP flow satisfies our needs...

State: New user, logged in, no IPs

Clicks on: "Add an IP"

Sees screen: "Where is this IP address located?"
Fills in: "address" and "postcode"
Clicks "next"

Sees screen "Add your IP"
Fills in: "123.12.123.11"
Clicks: "Add IP"

Sees IPs screen: "Location: XXXXX, IPs: "123.12.123.11""
Sees buttons "Add IP to location" - nested under the above IP
Sees button "Add IP in new location" further down page

Clicks "Add IP in new location" takes you to the above flow.
OR
Clicks "Add IP to location" and goes through the regular IP creation flow.

### DIFF
--- a/app/controllers/ips_controller.rb
+++ b/app/controllers/ips_controller.rb
@@ -4,7 +4,7 @@ class IpsController < ApplicationController
   end
 
   def create
-    @ip = current_organisation.ips.new(ip_params)
+    @ip = current_organisation.locations.first.ips.new(ip_params)
     if @ip.save
       redirect_to(
         ips_path,
--- a/app/controllers/ips_controller.rb
+++ b/app/controllers/ips_controller.rb
@@ -4,7 +4,8 @@ class IpsController < ApplicationController
   end
 
   def create
-    @ip = current_organisation.locations.first.ips.new(ip_params)
+    default_location = current_organisation.locations.first
+    @ip = default_location.ips.new(ip_params)
     if @ip.save
       redirect_to(
         ips_path,
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -5,14 +5,9 @@ class Organisation < ApplicationRecord
 
   validates :name, presence: true, uniqueness: true
 
-  before_create :set_radius_secret_key
   after_create :create_location
 
 private
-
-  def set_radius_secret_key
-    self.radius_secret_key = RadiusSecretKeyGenerator.new.execute
-  end
 
   def create_location
     self.locations.create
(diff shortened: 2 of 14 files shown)


app/controllers/ips_controller.rb
```diff
@@ -4,7 +4,7 @@ def new
   end
 
   def create
-    @ip = current_organisation.ips.new(ip_params)
+    @ip = current_organisation.locations.first.ips.new(ip_params)
```
tommotaylor: hack

app/models/organisation.rb
```diff
@@ -1,14 +1,20 @@
 class Organisation < ApplicationRecord
   has_many :users, inverse_of: :organisation
-  has_many :ips
+  has_many :locations
+  has_many :ips, through: :locations
```
tommotaylor: hack

app/models/organisation.rb
```diff
@@ -1,14 +1,20 @@
...
 
   validates :name, presence: true, uniqueness: true
 
   before_create :set_radius_secret_key
+  after_create :create_location
```
tommotaylor: hack hack hack

kjdchapman: Might be good to call this `create_default_location` or something like that, try and make it a bit clearer that it's the default bucket things fall into.

app/controllers/ips_controller.rb
kjdchapman: is it worth explicitly calling out what this location is at the moment?

```
default_location = current_organisation.locations.first
@ip = default_location.ips.new(ip_params)
```

kjdchapman:
As this breaks allowed-sites, should we mark it as WIP for now? I don't want it to accidentally get merged.

PR #124, opened 2018-09-11
Publish IPs and Locations to S3

tommotaylor:
Suggestion from @emileswarts to rename all instances of "Statistics" to "LocationsIps"


spec/features/add_an_ip_address_spec.rb
```diff
@@ -24,16 +24,31 @@
 
     context 'and that IP is valid' do
       before do
+        ENV["S3_STATISTICS_BUCKET"] = "StubBucket"
+        ENV["S3_STATISTICS_OBJECT_KEY"] = "StubKey"
+
+        allow_any_instance_of(Gateways::S3).to receive(:upload)
```
tommotaylor: May not be needed

spec/features/add_an_ip_address_spec.rb
```diff
@@ -24,16 +24,31 @@
 
     context 'and that IP is valid' do
       before do
+        ENV["S3_STATISTICS_BUCKET"] = "StubBucket"
+        ENV["S3_STATISTICS_OBJECT_KEY"] = "StubKey"
```
tommotaylor: Check if these are really needed

app/controllers/ips_controller.rb
```diff
@@ -7,6 +7,7 @@ def create
     default_location = current_organisation.locations.first
     @ip = default_location.ips.new(ip_params)
     if @ip.save
+      publish_locations_and_ips_to_s3
```
kjdchapman: I'd remove the implementation-specific part of this naming - the `to_s3` part.

This controller doesn't need visibility of how the publishing is working (and it can be discovered by a reader from how the use case is composed with `Gateways::S3`).

spec/features/add_an_ip_address_spec.rb
```diff
@@ -24,16 +24,31 @@
 
     context 'and that IP is valid' do
       before do
+        ENV['S3_PUBLISHED_LOCATIONS_IPS_BUCKET'] = 'StubBucket'
+        ENV['S3_PUBLISHED_LOCATIONS_IPS_OBJECT_KEY'] = "StubKey"
```
kjdchapman: Not sure if double quotes here will fail gov-lint?

spec/features/add_an_ip_address_spec.rb
```diff
@@ -24,16 +24,31 @@
...
+
+        allow_any_instance_of(Gateways::S3).to receive(:upload)
+      end
+
+      it 'calls locations_ips publishing' do
+        expect_any_instance_of(PublishLocationsIps).to receive(:execute)
```
kjdchapman: I wouldn't add this sort of mock to an acceptance spec - you're checking the internals of the system rather than external behaviour, which will make these tests brittle.

If I wanted to test this in an acceptance spec, I'd try to get as close to the external behaviour of the system as possible with what I'm checking.

Checking something is in a bucket seems impractical, so I'd probably check the right endpoints were called with the right data (I'm guessing adding something to a bucket is an API call across HTTP?).

spec/gateways/ips_spec.rb
```diff
@@ -0,0 +1,24 @@
+describe Gateways::Ips do
```
kjdchapman: We've tended to call the namespace `Gateway` rather than `Gateways`. I'm not really worried so long as it's consistent across the application, though.

app/controllers/ips_controller.rb
tommotaylor: Perhaps `publish_locations_and_ips_for_performance_platform`? I think I reader should know what this line does without having to look much further.

spec/features/add_an_ip_address_spec.rb
tommotaylor: They don't but good catch, will amend.

lib/use_cases/performance_platform/publish_locations_ips.rb
```diff
@@ -0,0 +1,15 @@
+class PublishLocationsIps
+  def initialize(upload_gateway:, ips_gateway:)
```
kjdchapman: We called these `destination gateway` and `source_gateway` on the other half of this work, and I like that better than calling either of these `ip_gateway`, as it might imply they're not dealing with the same data.

lib/gateways/s3.rb
```diff
@@ -0,0 +1,29 @@
+module Gateways
+  class S3
```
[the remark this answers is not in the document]
tommotaylor: Yeah the name spacing is all off across the app. Gateways are namespaced, UseCases are not. I think this requires a separate PR. I also prefer the `Something::Something::Something` structure, rather than nesting on new lines, as it makes it easier to search for in the codebase, especially when projects become big.

lib/use_cases/performance_platform/publish_locations_ips.rb
tommotaylor: makes sense to have them the same

app/controllers/ips_controller.rb
kjdchapman: I like that, though I did overhear someone the other day speculating about other clients for this data. Happy for you to decide, naming is hard.

spec/features/add_an_ip_address_spec.rb
kjdchapman: That is a positive move, but not sure if changing it is worth it (feels like we end up with a spec that's closer to the edge, but still suffers from the same brittleness as before).

This might be a little tough to do "correctly" if the gem is hooked-in system-wide, don't mind leaving it as is for speed's sake as it sounds like some work to sort out.

lib/gateways/s3.rb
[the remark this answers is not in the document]
kjdchapman: Agreed that this one is for another PR.

PR #137, opened 2018-09-19
Display the location of an IP on the IP index page

emileswarts:
Also style the "Add IP Address" button to be consistent with the button
on the team page.

screen shot 2018-09-19 at 15 51 39

### DIFF
--- a/spec/features/view_ip_addresses_spec.rb
+++ b/spec/features/view_ip_addresses_spec.rb
@@ -31,18 +31,29 @@ describe 'View all my IP addresses' do
     end
 
     context 'when user has IPs' do
-      let(:location) { user.organisation.locations.first }
-      let!(:ip_1) { create(:ip, location: location) }
-      let!(:ip_2) { create(:ip, location: location) }
+      let(:ip_1) { '10.0.0.1' }
+      let(:ip_2) { '10.0.0.2' }
+      let(:address_1) { '179 Southern Street, Southwark' }
+      let(:address_2) { '123 Northern Street, Whitechapel' }
+
+      let(:org) { user.organisation }
+      let(:location_1) { org.locations.create!(address: address_1) }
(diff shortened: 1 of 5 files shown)


spec/features/view_ip_addresses_spec.rb
```diff
@@ -31,16 +31,30 @@
     end
 
     context 'when user has IPs' do
-      let(:location) { user.organisation.locations.first }
-      let!(:ip_1) { create(:ip, location: location) }
-      let!(:ip_2) { create(:ip, location: location) }
+      let(:location_1) do
```
kjdchapman: I've struggled with trying to make tests like this 'nice' a few times - I hate `let!` because it feels like I'm going to trip up the next person, but if I put things in the before block I can't easily access them in the later assertions.

I've found _a_ way to do it that I think works better - pulling out the _data_ you're checking for and doing the creations based on that data in the before block:

```ruby
context 'when user has IPs' do
  let(:ip_1) { '10.0.0.1' }
  let(:ip_2) { '10.0.0.2' }
  let(:address_1) { '179 Southern Street, Southwark' }
  let(:address_2) { '123 Northern Street, Whitechapel' }
  
  let(:org) { user.organisation }
  let(:location_1) { org.locations.create!(address: address_1) }
  let(:location_2) { org.locations.create!(address: address_2) }

  before do
    create(:ip, address: ip_1, location: location_1)
    create(:ip, address: ip_2, location: location_2)

    sign_in_user user
    visit ips_path
  end

  it 'displays the list of my allowed IPs' do
    expect(page).to have_content(address_1)
    expect(page).to have_content(address_2)

    expect(page).to have_content(ip_1)
    expect(page).to have_content(ip_2)
  end
end
```

(That `let(:org)` might be pushing it, but it kept the other stuff off multiple lines, which I think is worth it.)

PR #141, opened 2018-09-20
Fix resend confirmation bug

kjdchapman:
Resetting the token doesn't seem to be necessary - looks like we were digesting the token before comparing, which might have confused things.

Still not super-clear on how this Devise functionality is working, but we removed some code and got the test passing, so it's a step in the right direction.


tommotaylor:
NICE!!!!!

PR #144, opened 2018-09-20
Radius IP whitelist functionality

emileswarts:
We are decommissioning the Allowed sites API and having the admin create
the Radius configuration instead.  It will be uploaded to S3 (later PR),
which is more reliable than a running EC2 instance.

This introduces an unused class which returns the configuration needed
for the radius servers.  We will generate this list each time a new IP
  is saved.


spec/unit/radius_ip_whitelist_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  ipaddr = 2.2.2.2
+  secret = radkey2
+}
+')
+    end
+  end
+
+  context 'with duplicates in the database' do
+    before do
+      create(:ip, address: '2.2.2.2', location: create(:location, organisation: create(:organisation)))
```
kjdchapman: I had expected this to be silently failing to create because of our "unique IP" restriction, and that the test was producing a false positive based on that.

Turns out... we don't actually validate on unique IPs in this app? That seems like a mistake on our part!

spec/unit/radius_ip_whitelist_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  let(:configuration_result) { subject.execute }
+
+  describe '#execute' do
+    it 'returns the correct configuration' do
+      expect(configuration_result).to eq('client 1-1-1-1 {
+  ipaddr = 1.1.1.1
+  secret = radkey1
+}
+client 1-2-2-1 {
```
kjdchapman: Thought about whether we should pull out these `client` blocks into some structure to reduce duplication in this suite, and maybe making the fact the spacing is necessary clearer.

Don't think this test file is large enough yet for it to be a problem though.

spec/unit/radius_ip_whitelist_spec.rb
emileswarts: Yes needs fixing at application database level.
Given this is of crucial importance, I'm happy with a guard here too

kjdchapman: I am very happy to have the test here as well!

Dockerfile
```diff
@@ -12,6 +12,7 @@ ENV LONDON_RADIUS_IPS '111.111.111.111,121.121.121.121'
 ENV DUBLIN_RADIUS_IPS '222.222.222.222,232.232.232.232'
 ENV S3_PUBLISHED_LOCATIONS_IPS_BUCKET 'StubBucket'
 ENV S3_PUBLISHED_LOCATIONS_IPS_OBJECT_KEY 'StubKey'
+ENV S3_WHITELIST_OBJECT_KEY 'WhitelistStubKey'
```
kjdchapman: Are we putting multiple objects in the same bucket?

kjdchapman:
I'm still good for this to go in post-changes :+1:

PR #145, opened 2018-09-20
Create new or select existing location when adding an IP address OLD

naliwajek:


spec/features/guidance_after_sign_in_spec.rb
```diff
@@ -16,7 +17,7 @@
 
   it 'displays radius connection details' do
     expect(page).to have_content(
-      "Your RADIUS secret is #{user.organisation.locations.first.radius_secret_key}"
+      user.organisation.locations.first.radius_secret_key
```
emileswarts: This could probably just be:
location.radius_secret_key ?

app/controllers/ips_controller.rb
```diff
@@ -14,6 +16,8 @@ def create
         notice: "#{@ip.address} added, it will be active starting tomorrow"
       )
     else
+      @ip.destroy if @ip.location.ips.count.zero?
+      @locations = current_locations
```
emileswarts: Where is this instance variable used?

app/controllers/ips_controller.rb
```diff
@@ -14,6 +16,8 @@ def create
         notice: "#{@ip.address} added, it will be active starting tomorrow"
       )
     else
+      @ip.destroy if @ip.location.ips.count.zero?
```
emileswarts: Not sure I follow this behaviour.
If @ip failed to save, we call .destroy on it?

app/controllers/ips_controller.rb
```diff
@@ -24,6 +28,12 @@ def index
 
 private
 
+  def current_locations
```
emileswarts: Not sure this should be in the controller, also the name suggests to me that it is as useful as current_user or current organisation

app/controllers/ips_controller.rb
naliwajek: in `render :new`

app/controllers/ips_controller.rb
naliwajek: Right, should be on `location`

app/controllers/ips_controller.rb
naliwajek: I'm no longer sure what we put in use cases and what not looking at our current state of controllers.

spec/features/guidance_after_sign_in_spec.rb
naliwajek: 👍

app/controllers/ips_controller.rb
```diff
@@ -1,11 +1,13 @@
...
   def new
     @ip = Ip.new
+    @locations = current_locations
   end
 
   def create
-    default_location = current_organisation.locations.first
-    @ip = default_location.ips.new(ip_params)
+    @ip = Ip.new(ip_params)
+    @ip.location = get_location || create_location
```
tommotaylor: Can this all be rewritten as: 
```
  @ip = Ip.new(params)
  if @ip.save
    publish_for_performance_platform
    redirect_to somewhere
  else
    render new
end


Where params are either:
`{ address: "1.1.1.1", location_attributes: { address: "Made Towers", postcode: "SW12 7YC", organisation_id: 1 } }`
or
`{ address: "1.1.1.1", location_id: 1 }`

Comment (tommotaylor) [reply app/controllers/ips_controller.rb]: Also need to add the following to `ip.rb` model:

```
  accepts_nested_attributes_for :location
```

File: app/controllers/ips_controller.rb (hunk 4 above)
Comment (emileswarts) [reply app/controllers/ips_controller.rb]: This is a very view specific data transformation.
Not suggesting we put it in a use case.

Isn't there a more elegant way to do this without this intermediate transformation?
https://apidock.com/rails/v4.2.1/ActionView/Helpers/FormOptionsHelper/select

File: app/views/ips/new.html.erb [hunk 6]
```diff
@@ -10,6 +10,25 @@
...
       <%= form_with model: @ip do |form| %>
+        <div class="govuk-form-group <%= field_error(@ip, :location_id) %>">
+          <%= form.label :location_id, "Select your location", class: "govuk-label"  %><br />
+          <%= form.select :location_id, @locations, { prompt: '-- select --' }, { class: "govuk-select" } %>
+        </div>
+
+        <p class="govuk-body"><strong>OR</strong></p>
+
+        <div class="govuk-form-group <%= field_error(@ip, :location_address) %>">
+          <%= form.label :location_address, "Enter new location address", class: "govuk-label"  %><br />
```
tommotaylor: This needs to be `f.fields_for :location` in order to pass the correct params.

See https://github.com/alphagov/govwifi-admin/blob/master/app/views/users/confirmations/show.html.erb#L12

app/controllers/ips_controller.rb
naliwajek: What will happen if params are not either but rather combined?

Because without JS, there's no way to stop someone from selecting a location but then still putting stuff in textbox.

`{ address: "1.1.1.1", location_id: 1, location_attributes: { address: "Made Towers", postcode: "SW12 7YC", organisation_id: 1 } }`

naliwajek:
```irb(main):008:0> Location.last
  Location Load (1.2ms)  SELECT  `locations`.* FROM `locations` ORDER BY `locations`.`id` DESC LIMIT 1
=> #<Location id: 3, radius_secret_key: "C043947A5C5C504A8486", address: "Made Towers", postcode: "SW12 7YC", organisation_id: 1, created_at: "2018-09-20 16:04:30", updated_at: "2018-09-20 16:04:30">
irb(main):009:0> @ip = Ip.new({ address: "1.1.1.1", location_id: 3, location_attributes: { address: "Made Towers", postcode: "SW12 7YC", organisation_id: 1 } })
  Location Load (1.3ms)  SELECT  `locations`.* FROM `locations` WHERE `locations`.`id` = 3 LIMIT 1
=> #<Ip id: nil, address: "1.1.1.1", created_at: nil, updated_at: nil, location_id: nil>
irb(main):010:0> @ip.save
   (0.8ms)  BEGIN
  Organisation Load (1.1ms)  SELECT  `organisations`.* FROM `organisations` WHERE `organisations`.`id` = 1 LIMIT 1
  Location Create (1.0ms)  INSERT INTO `locations` (`radius_secret_key`, `address`, `postcode`, `organisation_id`, `created_at`, `updated_at`) VALUES ('2EF7E0831A2CA97D195D', 'Made Towers', 'SW12 7YC', 1, '2018-09-20 16:05:16', '2018-09-20 16:05:16')
  Ip Create (0.8ms)  INSERT INTO `ips` (`address`, `created_at`, `updated_at`, `location_id`) VALUES ('1.1.1.1', '2018-09-20 16:05:16', '2018-09-20 16:05:16', 4)
   (1.6ms)  COMMIT
=> true
irb(main):011:0> @ip.location
=> #<Location id: 4, radius_secret_key: "2EF7E0831A2CA97D195D", address: "Made Towers", postcode: "SW12 7YC", organisation_id: 1, created_at: "2018-09-20 16:05:16", updated_at: "2018-09-20 16:05:16">```

PR #152, opened 2018-09-24
Remove remember me checkbox

georgeschena:
Removed "remember me" checkbox as the user details are being remembered regardless of ticking the box.


adrianclay:
Hey @georgeschena looks like you included a `package-lock.json` as part of this PR.  The project uses `yarn` instead of `npm`, so having both lockfiles will be confusing.

kjdchapman:
I managed to skim-miss this on review, thanks @adrianclay .

https://github.com/alphagov/govwifi-admin/pull/153 fixes, also a suggestion from @craigjbass there.

PR #153, opened 2018-09-24
Remove npm lockfile

kjdchapman:
This project uses yarn, so we don't want an npm-driven lockfile - it'd be confusing to future readers.

Suggestion from Craig to prevent this happening again, think that's one for the team to decide on:
> I would usually make this kinda thing a test
> `it 'does not have an npm lock file because we use yarn'`
> If it's easy to get wrong - test


kjdchapman:
Got a card on the backlog for the "other half" of this - testing to make sure it doesn't happen again.

PR #154, opened 2018-09-24
Change IP list to show availability, not dates

kjdchapman:
I took the liberty of removing the 'date added' field, as I believe it was trying to achieve the same thing as this.

This is WIP as I'm not happy with the state of the associated acceptance spec.

image


app/assets/stylesheets/utilities/layout.scss
```diff
@@ -42,3 +42,7 @@
 .health-check-unhealthy {
   background: red;
 }
+
+.unavailable-ip {
+  color: $govuk-error-colour;
+}
```
tommotaylor: The css is currently split between `/utilities` and `/components`. I would suggest creating css classes such as `error` and `red` in a file `utilities/colours.scss`

app/views/ips/_table.html.erb
```diff
@@ -3,14 +3,19 @@
...
   <tbody class="govuk-table__body">
     <% location.ips.each do |ip| %>
       <tr class="govuk-table__row">
-        <th class="govuk-table__header" scope="row"><%= ip.address %></th>
-        <td class="govuk-table__cell "><%= ip.created_at.strftime("%d %b %Y") %></td>
+        <td class="govuk-table__cell" scope="row"><%= ip.address %></td>
+        <% if ip.available? %>
+          <td class="govuk-table__cell">Available</td>
+        <% else %>
+          <td class="govuk-table__cell unavailable-ip">Activating tomorrow
```
tommotaylor: "Activating tomorrow" made me wonder what time tomorrow it will activate. Perhaps "Active tomorrow" or "Activating overnight" is more precise?

spec/features/view_ip_address_readiness_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+require 'features/support/not_signed_in'
+require 'features/support/sign_up_helpers'
+require 'features/support/errors_in_form'
+require 'features/support/activation_notice'
+require 'support/notifications_service'
+require 'support/confirmation_use_case'
+require 'timecop'
+
+describe 'even more requires' do
+  include_examples 'confirmation use case spy'
```
tommotaylor: You don't need this if you create users already confirmed. Eg: `create(:user, :confirmed, :with_organisation)`. If they are already confirmed when created, then the trigger to send a confirmation email is not called. Then you can remove:
```
require 'support/notifications_service'
require 'support/confirmation_use_case'
```

spec/features/view_ip_address_readiness_spec.rb
```diff
@@ -0,0 +1,52 @@
+# TODO: don't require eveerything ever to fill in a form
+require 'features/support/not_signed_in'
+require 'features/support/sign_up_helpers'
+require 'features/support/errors_in_form'
+require 'features/support/activation_notice'
+require 'support/notifications_service'
+require 'support/confirmation_use_case'
```
tommotaylor: Of the above 7 lines of `require`s, I think you only need `require 'features/support/sign_up_helpers'`. You're not using `it_behaves_like` on any of the others.

spec/features/view_ip_address_readiness_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+require 'timecop'
+
+describe 'even more requires' do
+  include_examples 'confirmation use case spy'
+  include_examples 'notifications service'
+
+  after { Timecop.return }
```
tommotaylor: I have always found not returning `Timecop` to be one of the biggest causes of random spec failures in a test suite. I wonder how much slower the specs would be if we threw this in `rails_helper.rb`?

app/assets/stylesheets/utilities/layout.scss
kjdchapman: Yeah, this was very much a chucked-in-not-sure-how-to-arrange-this-css job... might take the time to quickly clear this up, think this particular one should be in `/components` (since it's describing a particular part of the `ip/table` "component")

app/views/ips/_table.html.erb
kjdchapman: Nice catch, I think this can be clearer - "Available from 6am tomorrow"? Doesn't tie us to some time in the night to restart that way.

kjdchapman: I can hear designers in my head saying "make it a negative to distinguish it!", so maybe _Not available until 6am tomorrow_.

spec/features/view_ip_address_readiness_spec.rb
kjdchapman: Yep, just chucked these all in, going to remove them one-by-one as I clear it up.

spec/features/view_ip_address_readiness_spec.rb
kjdchapman: Inclined to do this when we have `Timecop` in multiple places. Then I'd try it out to see the speed change.

kjdchapman:
IP table is consistent within one location, but not across locations:
image

app/assets/stylesheets/utilities/layout.scss
kjdchapman: Think 4be35c makes this a lot better.

PR #155, opened 2018-09-24
Add redirect and flash message implementation to support page submit button

georgeschena:
Implemented a redirect and flash message feature to the submit button on the the support page, which when clicked redirects the user to the root path and presents the user with a flash message stating 'Your support request has been submitted' to inform the user that their support request was sent.


spec/use_cases/administrator/send_confirmation_email_spec.rb
```diff
@@ -1,33 +0,0 @@
-describe SendConfirmationEmail do
-  class SendConfirmationEmailGatewayMock
-    def send(opts)
-      raise unless opts[:email] == 'test@example.com'
-      raise unless opts[:locals][:confirmation_url] == 'https://example.com'
-      raise unless opts[:template_id] == 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
```
emileswarts: Is this functionality covered by a test elsewhere?

PR #157, opened 2018-09-25
Allow Jenkins cleanup to retry upon failure

tommotaylor:


Jenkinsfile
```diff
@@ -51,7 +51,9 @@ pipeline {
 
   post {
     always {
-      sh 'make stop'
+      retry(3) {
+        sh 'make stop'
```
emileswarts: Think 'make stop' fails?

kjdchapman:
Had a conversation IRL about this and realised there's a better diagnosis of the problem - not `make stop` failing on this build, but a different project failing to clean up after itself.

Going to change those other projects first, making sure they're all killing their docker containers properly on exit.

PR #158, opened 2018-09-25
Link new user to add IP's and locations

georgeschena:
Add test to inform new users to add IP's and locations.
Add conditional to display guidance message to guide the user to add IP's and locations.

Before: 
no-radius-secret-key

After: 
screen shot 2018-09-25 at 15 18 11

### DIFF
--- a/app/views/application/_radius_details.html.erb
+++ b/app/views/application/_radius_details.html.erb
@@ -7,8 +7,12 @@
   <div class="govuk-details__text">
     <p> Your RADIUS secrets are:</p>
     <ul class="govuk-list">
+    <% if (current_organisation.locations).empty? %>
+        <p>Your Radius secret keys will be generated when you <%=link_to('add your first IP address', new_ip_path, class: "govuk-link")%></p>
+    <% else %>
       <% current_organisation.locations.each do |loc| %>
-        <li><%= "#{loc.address}, #{loc.postcode}" %> - <strong><%= loc.radius_secret_key %></strong></li>
+          <li><%= "#{loc.address}, #{loc.postcode}" %> - <strong><%= loc.radius_secret_key %></strong></li>
+        <% end %>
       <% end %>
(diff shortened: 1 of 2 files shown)


app/views/application/_radius_details.html.erb
```diff
@@ -7,8 +7,12 @@
   <div class="govuk-details__text">
     <p> Your RADIUS secrets are:</p>
```
kjdchapman: I'd sink lines 8 and 9 into the `if` block - don't need the header or the list unless there are actually secret keys.

app/views/application/_radius_details.html.erb
```diff
@@ -7,8 +7,12 @@
...
     <ul class="govuk-list">
+    <% if (current_organisation.locations).empty? %>
+        <p>Your Radius secret keys will be generated when you <%=link_to('add your first IP address', new_ip_path, class: "govuk-link")%></p>
```
kjdchapman: `RADIUS` not `Radius` - it's an acronym.

spec/features/guidance_after_sign_in_spec.rb
```diff
@@ -3,44 +3,52 @@
...
-    expect(page).to have_content 'Get GovWifi'
-    expect(page).to have_content 'Getting help'
-  end
+  context 'without locations' do
+    before { visit root_path }
 
-  it 'displays radius connection details' do
-    expect(page).to have_content(location.radius_secret_key)
+    it 'displays message to inform user to add IPs and locations' do
+      expect(page).to have_content 'Your Radius secret keys will be generated when you add your first IP address'
```
kjdchapman: I worry that testing against this specific message is brittle - I've tended to try to and check against a short part of these descriptions, e.g. `RADIUS secret keys will be generated`, or maybe not even testing the message at all.

Happy for you to make a judgement call on this.

PR #160, opened 2018-09-26
Lock yarn and nodejs versions

craigjbass:
**Why:** We were seeing a flaky bug introduced by a new version of yarn released

* Yarn 1.9.4
* Nodejs 8~


craigjbass:
See https://github.com/yarnpkg/yarn/pull/6302 for more information on the upstream yarn bug

PR #162, opened 2018-09-26
Split column spacing evenly for ip table columns

kjdchapman:
image


app/assets/stylesheets/components/ips.scss
```diff
@@ -1,3 +1,7 @@
 .unavailable-ip {
   color: $govuk-error-colour;
 }
+
+.half-column {
```
emileswarts: There are no existing styles we can leverage?
I've seen things like "govuk-grid-column-two-thirds".

PR #165, opened 2018-09-27
Order radius IPs based on organisation ID

georgeschena:
For example, for our test organisation, they're now not in 'natural' order:

screen shot 2018-09-27 at 11 44 01

(We checked a few more and they're in different orders too)


kjdchapman:
Decided it was best to not try and test the ordering through UI-level testing, as that was pretty fragile and complicated. Instead, the unit test covers it.

PR #169, opened 2018-10-02
Get logs for a username

tommotaylor:
Still to do in another PR:
- Filter results based on organisations IPs
- Link to this search page in the navigation somewhere

Unknowns:
- How do we ensure this isn't always a broken page in development???? For instance, with the status page, the AWS gem stubs a response in development.

screenshot 2018-10-05 at 12 37 04


lib/use_cases/administrator/get_logs_for_username.rb
```diff
@@ -0,0 +1,13 @@
+module UseCases
+  module Administrator
+    class GetLogsForUsername
+      def initialize(gateway:)
```
kjdchapman: Even when there's only one `gateway` I'd use a more descriptive name for the named parameter - maybe `log_gateway`, or even `auth_request_gateway`.

Been going back and forth on this naming in `logging` as well, feels like it's more important to talk about WHAT we're logging, `log` is very generic.

Maybe this reflects on the name for the whole use case and it should be `ViewAuthRequestsForUsername`?

spec/features/logging/view_logs_for_a_username_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        "start" => "2018-10-01 18:18:09 +0000",
+        "stop" => nil,
+        "success" => true,
+        "username" => username
+      }
+    ]
+  end
+
+  before do
+    stub_request(:get, "http://govwifi-logging-api.com/authentication/events/search?username=#{username}").
```
kjdchapman: This looks like a good approach to me, webmocking the API in the acceptance spec.

Is it not working? Or were you concerned this isn't the right approach?

spec/gateways/logging_api_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+            "start" => "2018-10-01 18:18:09 +0000",
+            "stop" => nil,
+            "success" => true,
+            "username" => username
+          }
+        ]
+      end
+
+      before do
+        stub_request(:get, "http://govwifi-logging-api.com/authentication/events/search?username=#{username}").
```
kjdchapman: This seems good too - happy to use webmock when testing a "remote service"-style gateway.

spec/use_cases/administrator/get_logs_for_username_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        "success" => true,
+        "username" => username
+      }
+    ]
+  end
+
+  subject { described_class.new(gateway: logging_api_gateway ) }
+
+  before do
+    allow(logging_api_gateway).to receive(:search)
```
kjdchapman: I'd fully mock out the gateway here - it's not necessary to instantiate the "real" version at all.

I'd then use two separate specs for the two different bits of behaviour:
Testing that the use case passes the correct parameter to the gateway:
```ruby
let(:spy_gateway) { spy(search: [])
expect(spy_gateway).to have_received(:search).with('some expected term')
```

Testing that it uses the results from the gateway appropriately.
```ruby
let(:stub_gateway) { double(search: results) 
expect(subject.execute).to eq(something_like_results)
```

(I can spin this out into a more full example if it's not clear)

config/site.yml
```diff
@@ -4,6 +4,7 @@ default: &default
   organisation_docs_link: 'https://docs.wifi.service.gov.uk'
   product_page_link: 'https://wifi.service.gov.uk'
   organisation_troubleshooting_link: 'https://docs.wifi.service.gov.uk/troubleshooting.html#troubleshooting'
+  logging_api_search_endpoint: 'http://api-elb.london.wifi.service.gov.uk:8443/logging/authentication/events/search/'
```
tommotaylor: is this sensitive?

app/controllers/logs_controller.rb
```diff
@@ -0,0 +1,9 @@
+class LogsController < ApplicationController
+  def index
+    result = UseCases::Administrator::GetAuthRequestsForUsername.new(
```
emileswarts: Intermediate `result` variable provide any value, over just assigning to `@logs` ?

config/site.yml
emileswarts: Yeah would stick this in an ENV variable, it also needs to be https

lib/gateways/logging_api.rb
```diff
@@ -0,0 +1,14 @@
+module Gateways
+  class LoggingApi
+    LOGGING_API_SEARCH_ENDPOINT = SITE_CONFIG['logging_api_search_endpoint'].freeze
+
+    def search(username:)
+      return [] if username.nil? || username.empty?
```
emileswarts: Probably also validate beyond nil, and ensure a username is 6 characters long before we send this request over the wire.

spec/gateways/logging_api_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      before do
+        endpoint = SITE_CONFIG['logging_api_search_endpoint'] + username
+        stub_request(:get, endpoint)
+          .with(headers: {
+              'Accept' => '*/*',
+              'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
+              'Host' => 'api-elb.london.wifi.service.gov.uk:8443',
+              'User-Agent' => 'Ruby'
+            })
+          .to_return(status: 200, body: logs.to_json, headers: {})
```
emileswarts: More of a general testing comment, but feels like the feature spec covers this pretty well.  Not sure if there is a general consensus around testing feature vs gateway, without too much overlap.  Keen to hear your thoughts

spec/use_cases/administrator/get_auth_requests_for_username_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        "siteIP" => "",
+        "start" => "2018-10-01 18:18:09 +0000",
+        "stop" => nil,
+        "success" => true,
+        "username" => username
+      }
+    ]
+  end
+
+  subject { described_class.new(authentication_logs_gateway: authentication_logs_gateway) }
```
emileswarts: Wonder if we should send in a dummy gateway, stubbing the real gateway feels like we are not reaping the benefits of the Clean Architecture approach.

app/views/logs/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Displaying logs for...</h1>
+
+<ul>
+  <% @logs.each do |log| %>
+    <li><%= log["username"] %></li>
```
emileswarts: Small but I always prefer hashes to blow up with .fetch() so pesky nils don't percolate through the system

spec/gateways/logging_api_spec.rb
tommotaylor: I'm rather new to Clean Architecture so I'm happy to follow advice. If its not needed, then I'm happy to remove it. Currently we test all gateways.

spec/use_cases/administrator/get_auth_requests_for_username_spec.rb
tommotaylor: 🐺

PR #170, opened 2018-10-02
'pending invite' text is now responsive on small device screens

georgeschena:


kjdchapman:
I'd name PRs in the imperative like git commits - so this'd be
> Style 'pending invite' text to be responsive on small device screens

Also, if the PR has a visual element to explain, I'd include screenshots.

app/assets/stylesheets/components/team_members.scss
```diff
@@ -1,3 +1,9 @@
 .small-table-button {
   margin-bottom: 0px;
 }
+
+@media only screen and (max-width: 400px) {
+  #pending-invite {
+    display: none;
```
kjdchapman: Won't this hide the element on small screens? Is that the desired behaviour?

From the PR/commit titles, I'd guessed this would shrink or move the text rather than hide it.

PR #175, opened 2018-10-08
Add user feedback banner 

georgeschena:
Add user feedback banner with a link to the users emailer to allow users to provide us with feedback on the GovWifi system.

### DIFF
--- a/app/helpers/list_helper.rb
+++ b/app/helpers/list_helper.rb
@@ -1,0 +1,2 @@
+module ListHelper
+end
--- a/app/views/list/_table.html.erb
+++ b/app/views/list/_table.html.erb
@@ -8,11 +8,11 @@
     </tr>
   </thead>
   <tbody class="govuk-table__body">
-      <tr class="govuk-table__row">
-        <td class="govuk-table__cell one-quarter" scope="row">Parks & Recreation Dept</td>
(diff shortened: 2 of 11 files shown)


app/helpers/list_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ListHelper
+end
```
tommotaylor: Did you use a generator George? Seems like this helper, `list.js` and `list.scss` are unnecessary.

app/views/list/_table.html.erb
```diff
@@ -0,0 +1,18 @@
+<table class="govuk-table">
```
tommotaylor: this is all unrelated code, I think you've branched off a feature branch rather than master?

app/views/list/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="govuk-grid-row">
```
tommotaylor: Also unrelated code.

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  get 'list/index'
```
tommotaylor: unrelated code.

PR #176, opened 2018-10-08
Display a list of signed up organisations

Renny4Real:
Add authentication for super admin to see the list of organisations.

Show only organisation tab to admin.

Show the other tabs to regular users.


app/views/list/_table.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <th class="govuk-table__header">Created At</th>
+      <th class="govuk-table__header">Updated At</th>
+      <th class="govuk-table__header">Service Email</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+  <% @all_organisations.each do | organisation| %>
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.name %></td>
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.created_at.strftime('%e %b %Y %H:%M %P') %></td>
```
georgeschena: What does the  `strftime` mean and what does the `('%e %b %Y %H:%M %P')` do?

Renny4Real: It's a method that formats the date and time. The arguments are what determine how it looks.

app/helpers/list_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ListHelper
+end
```
naliwajek: We can safely remove helper files if they're empty.

app/views/list/_table.html.erb
```diff
@@ -0,0 +1,21 @@
...
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.updated_at.strftime('%e %b %Y %H:%M %P') %></td>
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.service_email %></td>
```
naliwajek: Do we want this to be a link like `<a href="mailto:service@email.com>service@email.com</a>` ?

if yes, then this may help: https://apidock.com/rails/ActionView/Helpers/UrlHelper/mail_to

app/views/list/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Signed Up Organisations</h1>
+  </div>
+  <%  @all_organisations.each do |organisation| %>
+    <%= render "list/table", organisation: organisation %>
```
naliwajek: I'm confused. Why are we passing local variable `organisation` here? In `list/table` template we seem to be using only a global variable called `@all_organisation`, am I right?

If we use `@all_organisations.each` here, will it display N tables where N is the number of organisations we have? Is that we want? 🤔

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  get 'list/index'
```
naliwajek: I'd stick to the way of defining routes we already have, like in:
https://github.com/alphagov/govwifi-admin/pull/176/files#diff-21497849d8f00507c9c8dcaf6288b136R16

app/views/list/_table.html.erb
Renny4Real: Will look into it, not sure how admin wants it to be yet.

app/views/list/index.html.erb
Renny4Real: One table with N rows for each organisation.

spec/features/view_organisation_list_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    let!(:user) { create(:user, :confirmed, :with_organisation) }
+
+    context 'when admin navigates to the list page' do
+      it 'renders the names of the signed up organisations' do
+        sign_in_user user
+        visit organisations_path
+
+        Organisation.all.each do |organisation|
+          expect(page).to have_content(organisation.name)
```
tommotaylor: Nice!

app/views/application/_subnavigation.html.erb
```diff
@@ -7,6 +7,7 @@
 
   <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-0 govuk-!-padding-right-0">
     <nav class="govuk-body govuk-!-margin-bottom-0 ">
+      <%= link_to "Organisations", organisations_path, class: active_tab(organisations_path) %>
```
tommotaylor: I don't think we want to show a link to this page. Only GDS are going to be able to get access to this page so we can just remember the url.

naliwajek: 👍 for this idea.

naliwajek:
Remember to squash when merging and add a description to commit itself.

app/views/application/_subnavigation.html.erb
```diff
@@ -7,9 +7,13 @@
 
   <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-0 govuk-!-padding-right-0">
     <nav class="govuk-body govuk-!-margin-bottom-0 ">
+    <% if current_user.admin? %>
+      <%= link_to "Organisations", organisations_path, class: active_tab(organisations_path) %>
+    <% else %>
```
tommotaylor: This `else` means that if you are an admin you don't see any of the other features. Is that intended?

app/views/organisations/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        <th class="govuk-table__header">Updated At</th>
+        <th class="govuk-table__header">Service Email</th>
+      </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+      <% @all_organisations.each do | organisation| %>
+        <tr class="govuk-table__row">
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.name %></td>
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.created_at.strftime('%e %b %Y %H:%M %P') %></td>
+        <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.updated_at.strftime('%e %b %Y %H:%M %P') %></td>
```
tommotaylor: I don't think we need updated at. You can't currently update an organisation.

config/routes.rb
```diff
@@ -13,6 +13,7 @@
 
   get '/healthcheck', to: 'monitoring#healthcheck'
   resources :status, only: [:index]
+  resources :organisations, only: [:index]
```
tommotaylor: Should use the `%i[` style here. All arrays in here should be in that style.

db/migrate/20181009090135_add_admin_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddAdminToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :admin, :bool, default: false
```
tommotaylor: this should be `:boolean` https://api.rubyonrails.org/classes/ActiveRecord/Migration.html

tommotaylor:
I've made a few suggestions but also approved. Code all looks good just wondering about a few small things that I made comments on.

app/views/application/_subnavigation.html.erb
Renny4Real: Yeah. Jakub suggested that as admins we may not need the other features.

Renny4Real:
screen shot 2018-10-09 at 16 10 49

Just realised that as the admin, there organisation also appears on the list. Should it be left like that or should we remove the admin's organisations from the list?

PR #177, opened 2018-10-08
Add user feedback banner

georgeschena:
Add user feedback banner with a link to the users emailer to allow users to provide us with feedback on the GovWifi system.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -98,6 +98,15 @@
     <%= render "subnavigation" if user_signed_in? %>
 
     <main class="govuk-main-wrapper " id="main-content" role="main">
+
+      <div class="govuk-phase-banner feedback-banner">
+        <p class="govuk-phase-banner__content">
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -98,6 +98,15 @@
     <%= render "subnavigation" if user_signed_in? %>
 
     <main class="govuk-main-wrapper " id="main-content" role="main">
+
+      <div class="govuk-phase-banner feedback-banner">
+        <p class="govuk-phase-banner__content">
+         <span class="govuk-phase-banner__text">
+           This is a new service – your
+           <a class="govuk-link" href="mailto:govwifi-support@digital.cabinet-office.gov.uk?subject=User Feedback Suggestion">feedback</a> will help us to improve it.
```
naliwajek: Do we have any config file in which we could keep this mailto link?

Though not necessarily, I can accept the way it's now as well.

PR #183, opened 2018-10-10
Restructure dockerfile + make

cybojenix:
The current flow appears to be:
- stop services
- remove anonymous volumes (database)
- build app image
- create + seed database
- serve/test

this process takes a while to run, and there are some unnecessary steps in the mix.
This is initiated by running `make serve`.

as precompiling assets isn't needed for development, we have implemented a multi-stage dockerfile, so we don't precompile assets.
This also reduces the docker context to only needing to know about Gemfiles, which reduces the amount of false positive detections for a rebuild.

We will also avoid destroying volumes. This involves 2 changes:
- introducing a `make setup` command, which currently creates + seeds the database
- introducing a `make clean` command, which will remove docker volumes for you.

The new process will be running `make setup` once, then being able to run `make serve` going forward


Makefile
```diff
@@ -9,38 +9,43 @@ ifndef JENKINS_URL
...
 
-serve:
-	$(MAKE) build
+.PHONY: launch_db
+launch_db:
 	$(DOCKER_COMPOSE) up -d db
 	$(DOCKER_COMPOSE) up -d rr_db
 	./mysql/bin/wait_for_mysql
-	./mysql/bin/wait_for_rr_db
+  ./mysql/bin/wait_for_rr_db
```
cybojenix: this has a space

emileswarts:
Nice, thanks for looking into this.
Read through the PR and there is a lot of syntax I'm unfamiliar with.

How much of a performance impact does this have?

Dockerfile
```diff
@@ -31,8 +31,11 @@ ENV PATH "$PATH:/root/.yarn/bin:/root/.config/yarn/global/node_modules/.bin"
 COPY Gemfile Gemfile.lock .ruby-version ./
 RUN ${BUNDLE_INSTALL_CMD}
 
+CMD ["bundle", "exec", "rails", "server"]
+
+
+FROM base as production
```
emileswarts: Does aliasing this have a performance impact?  Excuse my ignorance here.

cybojenix: Looks like it just loads the previous layer while building, so there will be no meaningful impact

PR #186, opened 2018-10-11
Add MOU page and table cells.

Renny4Real:
Add MOU page for organisations to download and upload MOUs.

Add MOU page for admins to upload MOU template for users to engage with.

Add MOU upload and download cells in the organisations table for admins.
screen shot 2018-10-11 at 13 48 05

The information in the table is rendered in alphabetical order.

Any suggestions on how to get the table header to span two columns would be appreciated. Using `colspan-2` and removing the class on the row header tag works but removes the border underneath MOU.

Use staging s3 bucket to try out implementation.

After deploying to staging, the environment that stores the s3 bucket name will need to be renamed.


Dockerfile
```diff
@@ -28,6 +28,6 @@ RUN ${BUNDLE_INSTALL_CMD}
 
 COPY . .
 
-RUN RAILS_ENV=production bundle exec rails assets:precompile
+RUN ASSET_PRECOMPILATION_ONLY=true RAILS_ENV=production bundle exec rails assets:precompile
```
georgeschena: What does this do?

app/controllers/mou_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    Organisation.find(params[:id]).signed_mou.attach(params[:signed_mou])
+    redirect_to organisations_path
+  end
+
+  def upload
+    current_organisation.signed_mou.attach(params[:signed_mou])
+    redirect_to mou_index_path
+  end
+
+  def admin
```
georgeschena: Could this be named better?

app/controllers/organisations_controller.rb
```diff
@@ -2,7 +2,7 @@ class OrganisationsController < ApplicationController
   before_action :index, :authorise_admin
 
   def index
-    @all_organisations = Organisation.all
+    @all_organisations = Organisation.order('name ASC').all
```
kjdchapman: I would write this as `.order(:name)`, iirc the other syntax can cause issues.

I'm happy to lose the explicit callout that it's `ASC` order , as it's rails convention that `.order` will produce an alphabetical list in this case.

Dockerfile
Renny4Real: When using active storage's engine, booting the app requires a secret key base variable to be set and this variable is set through environment variables on production machines only. But the production mode is used to precompile assets. So instead of sliming the secret key base here, we use this variable as a flag that won't require active storage engine during precomplilation.

app/controllers/mou_controller.rb
kjdchapman: Maybe it should be a different controller? Think it'll make it clearer how separate it is, and separate out the access logic (when we come to add it in) nicely.

app/views/mou/admin.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-l">Memorandum of Understanding</h2>
+      <p class="govuk-body">Admins can upload a copy of a GDS signed MOU.</p>
+      <p class="govuk-body">Be aware that any version of the MOU template that is uploaded will be visible to all organisations requesting to sign up to GovWifi.</p>
+      <hr class="govuk-section-break--m govuk-section-break--visible">
+      <% if Mou.first.unsigned_template.attached? %>
```
kjdchapman: I'm not sure why this is `Mou.first`. Is the first record some special case?

spec/features/upload_download_mous_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    it ' uploads and downloads the mou template' do
+      sign_in_user user
+      visit admin_mou_index_path
+
+      attach_file("unsigned_template", Rails.root + "spec/fixtures/mou.pdf")
+      click_on 'Upload'
+
+      click_on 'Download current template'
+
+      expect(page.body).to eq("12334567 signed mou with content\n")
```
kjdchapman: Is this the content of the test pdf?

spec/features/upload_download_mous_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'features/support/not_signed_in'
+require 'features/support/sign_up_helpers'
+
+describe 'the upload and download of MOUs', focus: true do
```
kjdchapman: I know this is WIP, but want to drop this `focus: true` before merging.

app/views/organisations/index.html.erb
```diff
@@ -17,7 +18,17 @@
         <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.name %></td>
         <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.created_at.strftime('%e %b %Y %H:%M %P') %></td>
         <td class="govuk-table__cell quarter-column" scope="row"><%=  organisation.service_email %></td>
-        <td class="govuk-table__cell quarter-column" scope="row"></td>
+        <td class="govuk-table__cell quarter-column" scope="row">
```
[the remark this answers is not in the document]
Renny4Real: I don't think so, even after removing `quarter-column`, the table is still not spaced evenly.
screen shot 2018-10-11 at 14 30 38

spec/features/upload_download_mous_spec.rb
Renny4Real: Yes 😄

spec/features/upload_download_mous_spec.rb
Renny4Real: Noticed and removed 👍

app/models/mou.rb
```diff
@@ -0,0 +1,3 @@
+class Mou < ApplicationRecord
+  has_one_attached :unsigned_template
+end
```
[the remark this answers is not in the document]
Renny4Real: @tommotaylor 

Class can be renamed to reflect its use – it's going to be a singleton (so `Mou.first` will be gone) holding whatever MOU template is at the moment. 

What structure of Rails controllers you propose to handle:
- show page with MOU for a normal user
- action to upload an MOU for a normal user
- show page with MOU temp[late for an admin
- action to upload an MOU template for an admin

Would you keep MOU template here? And have a corresponding controller? And then have another model for organisations' MOUs? Or keep them on organisations? What about controller for them?

I admit, I struggle with creating any sensible logic using Rails convention.

– Jakub

Dockerfile
Renny4Real: Yep 😄

app/controllers/admins/mou_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Admins::MouController < ApplicationController
```
kjdchapman: Can I do things with this controller if I'm not a super-admin?

spec/features/upload_download_mous_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        expect(page.body).to eq("12334567 signed mou with content\n")
+      end
+    end
+  end
+
+  context 'As a super admin' do
+    let!(:organisation) { create(:organisation) }
+    let!(:user) { create(:user, :confirmed, email: 'me@example.gov.uk', organisation: organisation, admin: true) }
+
+    it ' uploads and downloads the mou template' do
```
kjdchapman: typo - extra space between ' and uploads

app/controllers/admins/mou_controller.rb
naliwajek: 👍

PR #187, opened 2018-10-11
Error handling for searching logs

tommotaylor:
- Better error messages for incorrect usernames
- Improve column table titles
- Add back button
- Add govuk classes to text
- Remove milliseconds from timestamp
- Sort results by latest first

screenshot 2018-10-11 at 14 07 54
screenshot 2018-10-11 at 14 07 44

### DIFF
--- a/app/views/logs/index.html.erb
+++ b/app/views/logs/index.html.erb
@@ -12,7 +12,7 @@
         <th class="govuk-table__header" scope="col">Access Point</th>
         <th class="govuk-table__header" scope="col">MAC Address</th>
         <th class="govuk-table__header" scope="col">IP</th>
-        <th class="govuk-table__header" scope="col">Start</th>
+        <th class="govuk-table__header" scope="col">Time</th>
         <th class="govuk-table__header" scope="col">Status</th>
       </tr>
     </thead>
--- a/app/views/logs/index.html.erb
+++ b/app/views/logs/index.html.erb
@@ -1,29 +1,30 @@
-<% if @logs.empty? %>
-  <h3>No results found for that username, please try again</h3>
-<% else %>
+<%= link_to "Back to search", search_logs_path, class: "govuk-back-link" %>
 
-  <h1>Displaying logs for <%= @logs.first.fetch("username") %></h1>
+<% if @logs.empty? %>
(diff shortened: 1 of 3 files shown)


app/views/logs/index.html.erb
```diff
@@ -1,29 +1,30 @@
+<%= link_to "Back to search", search_logs_path, class: "govuk-back-link" %>
+
 <% if @logs.empty? %>
-  <h3>No results found for that username, please try again</h3>
+  <h3 class="govuk-heading-m">No results found for that username, please try again</h3>
```
kjdchapman: I feel like this language suggests our search can fail, and that being more direct will cause less people to contact us thinking it's 

I'm thinking something more like:
> No authentication attempts found.

> If this user has tried to use GovWifi, there is a problem with a local network forwarding authentication requests.

app/views/logs/index.html.erb
```diff
@@ -1,29 +1,30 @@
...
-        <th class="govuk-table__header" scope="col">AP</th>
-        <th class="govuk-table__header" scope="col">MAC</th>
-        <th class="govuk-table__header" scope="col">siteIP</th>
-        <th class="govuk-table__header" scope="col">Start</th>
-        <th class="govuk-table__header" scope="col">Success</th>
+        <th class="govuk-table__header" scope="col">Access Point</th>
+        <th class="govuk-table__header" scope="col">MAC Address</th>
+        <th class="govuk-table__header" scope="col">IP</th>
+        <th class="govuk-table__header" scope="col">Time</th>
+        <th class="govuk-table__header" scope="col">Status</th>
```
kjdchapman: I like `Result` better than `Status` - the latter implies that this is something still in-progress, could be confusing?

spec/features/logging/view_auth_requests_for_a_username_spec.rb
```diff
@@ -3,7 +3,6 @@
 describe "View authentication requests for a username" do
```
kjdchapman: I worried that this PR only covers the validation and empty result set, but then I realised George is writing some more tests around this - that's how you decided to split it up?

app/views/logs/index.html.erb
tommotaylor: Added to icebox.

PR #188, opened 2018-10-11
Filter authorisation requests

georgeschena:
Before: When searching the authorisation requests the logs showed ALL of the IP's, even if  that organisation did not own it.

After: When searching the authorisation request logs it now only displays the results for IP's that the organisation owns.

### DIFF
--- a/app/controllers/logs_controller.rb
+++ b/app/controllers/logs_controller.rb
@@ -4,7 +4,11 @@ class LogsController < ApplicationController
       authentication_logs_gateway: Gateways::LoggingApi.new
     ).execute(username: params[:username])
 
+    filter_auth_requests_by_organisation(auth_requests)
+  end
+
+  def filter_auth_requests_by_organisation(auth_requests)
     ips = current_organisation.ips.map(&:address)
-    @logs = auth_requests.select { |auth_request| ips.include? (auth_request["siteIP"]) }
+    @logs = auth_requests.select { |auth_request| ips.include?(auth_request["siteIP"]) }
   end
 end
--- a/app/controllers/logs_controller.rb
+++ b/app/controllers/logs_controller.rb
@@ -3,5 +3,7 @@ class LogsController < ApplicationController
     @logs = UseCases::Administrator::GetAuthRequestsForUsername.new(
       authentication_logs_gateway: Gateways::LoggingApi.new
     ).execute(username: params[:username])
+
+    # @raw_logs.filter{|log| log is from an ip that is in the current_organisation.ips array}
   end
 end
--- a/app/controllers/logs_controller.rb
+++ b/app/controllers/logs_controller.rb
@@ -1,9 +1,10 @@
 class LogsController < ApplicationController
   def index
-    @logs = UseCases::Administrator::GetAuthRequestsForUsername.new(
+    auth_requests = UseCases::Administrator::GetAuthRequestsForUsername.new(
       authentication_logs_gateway: Gateways::LoggingApi.new
     ).execute(username: params[:username])
 
-    # @raw_logs.filter{|log| log is from an ip that is in the current_organisation.ips array}
+    ips = current_organisation.ips.map(&:address)
+    @logs = auth_requests.select { |auth_request| ips.include? (auth_request["siteIP"]) }
   end
 end
--- a/app/controllers/logs_controller.rb
+++ b/app/controllers/logs_controller.rb
@@ -4,11 +4,11 @@ class LogsController < ApplicationController
       authentication_logs_gateway: Gateways::LoggingApi.new
     ).execute(username: params[:username])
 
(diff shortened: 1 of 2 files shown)


app/controllers/logs_controller.rb
```diff
@@ -1,7 +1,14 @@
...
+    auth_requests = UseCases::Administrator::GetAuthRequestsForUsername.new(
       authentication_logs_gateway: Gateways::LoggingApi.new
     ).execute(username: params[:username])
+
+    filter_auth_requests_by_organisation(auth_requests)
+  end
+
+  def filter_auth_requests_by_organisation(auth_requests)
+    ips = current_organisation.ips.map(&:address)
+    @logs = auth_requests.select { |auth_request| ips.include?(auth_request["siteIP"]) }
```
kjdchapman: I don't like setting the instance variables in a method like this - both from a clarity and "rails convention" point-of-view.

I'd just rejig things slightly so the `index` method doesn't have any side-effects:

```ruby
def index
    auth_requests = UseCases::Administrator::GetAuthRequestsForUsername.new(
      authentication_logs_gateway: Gateways::LoggingApi.new
    ).execute(username: params[:username])

    @logs = filter_by_organisation(auth_requests)
  end

  def filter_by_organisation(auth_requests)
    ips = current_organisation.ips.map(&:address)
    auth_requests.select { |auth_request| ips.include?(auth_request["siteIP"]) }
  end
end
```
(didn't test this)

This whole controller is very small at the moment, so this is more of a future concern, but I'd fix it now.

spec/features/logging/view_auth_requests_for_a_username_spec.rb
```diff
@@ -2,16 +2,18 @@
 
 describe "View authentication requests for a username" do
   context "with results" do
-    let(:user) { create(:user, :confirmed, :with_organisation) }
     let(:username) { "AAAAAA" }
+    let(:organisation) { create(:organisation) }
```
kjdchapman: This spec feels like it still has "too much" setup, but I think that's going to be pretty tricky to fix. 

I'm happy to move forward with it like this for the moment, but I want to do some work on it - possibly these specs in general, as our models are getting complex - if we can find time.

kjdchapman:
Looks good to me, but I'll wait 'til the merge conflicts are fixed and just skim it again, before giving it the thumbs-up.

spec/features/logging/view_auth_requests_for_a_username_spec.rb
georgeschena: Yeah I agree. Coming from the outside and learning all this from scratch I feel it was pretty complex to set up so maybe in the future we can refactor the specs to be more 'test friendly' @kjdchapman

spec/features/logging/view_auth_requests_for_a_username_spec.rb
```diff
@@ -42,7 +55,13 @@
           })
         .to_return(status: 200, body: logs.to_json, headers: {})
 
-      sign_in_user user
+      organisation = create(:organisation)
+      location_two = create(:location, organisation: organisation)
+
+      create(:ip, location_id: location.id, address: "1.1.1.1")
+      create(:ip, location: location_two, address: "2.2.2.2")
```
tommotaylor: I don't think you need any of lines 58 - 62 here. You already have an organisation (named `organisation`) from line. 6. And you don't actually use these IPs after you've created them as we are just stubbing the response to have a log with the text "2.2.2.2" we are not actually looking for objects in the db.

spec/features/logging/view_auth_requests_for_a_username_spec.rb
```diff
@@ -51,13 +70,23 @@
...
+    end
+
+    it "does not display the logs of the ip the organisation does not own" do
+      expect(page).to_not have_content("2.2.2.2")
+    end
   end
 
   context "without results" do
     let(:user) { create(:user, :confirmed, :with_organisation) }
+    let(:admin_user) { create(:user, :confirmed, organisation_id: organisation.id) }
```
tommotaylor: Do we need both these users?

app/controllers/logs_controller.rb
tommotaylor: Yeah I tend to agree, if you are setting a crucial instance variable for the view eg `@logs` then they should be set in the index action.

But I think you'll fix this in the merge conflict anyway George.

spec/features/logging/view_auth_requests_for_a_username_spec.rb
Renny4Real: If it's to create a super admin, you should be able to just add `admin: true`
e.g. `let(:admin_user) { create(:user, :confirmed, organisation_id: organisation.id, admin: true) }`
Adding an admin has been pushed into master, so it should work here?

PR #189, opened 2018-10-11
add coverage reports to the tests

cybojenix:


kjdchapman:
Be interested to hear about LS's feeling on code coverage tools - I'm guessing from this PR at least some of you are fans! - because I'm not generally a fan.

That said, if it's helpful for you and the rest of the team are fine with it, I'm happy for us to add this on.

cybojenix:
While I agree seeking 100% coverage for the sake of coverage is bad, it's useful for discovering what parts of the code are not tested.
image

PR #190, opened 2018-10-11
Allow admin users to view logs

emileswarts:
We are pulling logs from the read replica now.
We wanted to pull logs from an endpoint exposed on the Logging API but
due to our loadbalancer being public, we are unable to whitelist the
admin security group.


lib/use_cases/administrator/get_auth_requests_for_username.rb
```diff
@@ -6,8 +6,16 @@ def initialize(authentication_logs_gateway:)
...
       def execute(username:)
+        return [] if username_invalid?(username)
+
         @authentication_logs_gateway.search(username: username)
       end
+
+    private
+
+      def username_invalid?(username)
+        username.nil? || username.empty? || username.length != 6
```
Renny4Real: I mentioned in Tom's PR that it would be nice to extract the username validator out of a similar piece of code #dry

PR #191, opened 2018-10-11
Improve styling, spacing, wording 

kjdchapman:
These changes is based on some of the feedback in the design review we had recently. There are many more small pieces to do.

These are small enough changes that doing a PR for each feels excessive, but I want to merge them in at semi-regular intervals to avoid them getting stale - especially as they're tending to be changes cutting across multiple parts of the app.


kjdchapman:
(I can add some screenshots if it helps to explain exactly what these changes achieve)

app/views/ips/index.html.erb
```diff
@@ -1,34 +1,37 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">IP Addresses</h1>
+    <h1 class="govuk-heading-l">IP addresses</h1>
   </div>
   <div class="govuk-grid-column-one-third">
-    <%= link_to "Add IP Address", new_ip_path, class: "govuk-button" %>
+    <%= link_to "Add IP address", new_ip_path, class: "govuk-button" %>
```
Renny4Real: Would it read better as `Add an IP address` ?

PR #196, opened 2018-10-15
Show location postcode in list of IPs, if available

kjdchapman:
image

Added a method to the model to provide this `full address`, didn't feel like enough logic to justify more components than that.

### DIFF
--- a/spec/features/view_ip_addresses_spec.rb
+++ b/spec/features/view_ip_addresses_spec.rb
@@ -31,46 +31,44 @@ describe 'View IP addresses' do
     end
 
     context 'with IPs' do
-      let(:ip_1) { '10.0.0.1' }
-      let(:ip_2) { '10.0.0.2' }
-      let(:address_1) { '179 Southern Street, Southwark' }
-      let(:address_2) { '123 Northern Street, Whitechapel' }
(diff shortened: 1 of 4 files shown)


spec/features/view_ip_addresses_spec.rb
```diff
@@ -18,42 +18,59 @@
...
+      end
+
+      it 'shows the related addresses' do
         expect(page).to have_content(address_1)
         expect(page).to have_content(address_2)
+      end
 
-        expect(page).to have_content(ip_1)
-        expect(page).to have_content(ip_2)
+      it 'shows the related postcoes' do
```
emileswarts: postcoes

spec/models/location_spec.rb
```diff
@@ -1,18 +1,47 @@
...
+      subject.postcode = postcode
+    end
+
+    context 'with address' do
+      let(:address) { "121 Fictional Street" }
+
+      context 'and postcode' do
+        let(:postcode) { "FI5 S67" }
+
+        it "combines the two" do
```
emileswarts: Would stick to single quotes for non interpolated strings for consistency.
Pedantic sorry!

kjdchapman: Not at all, consistency is <3

PR #198, opened 2018-10-15
When adding IPs and this is the first location, hide locations dropdown (+minor styling fix)

kjdchapman:
I don't think this needs any new tests, but I'm happy to hear opinions.

(Cheeky to put aligning in the same PR, can rebase rather than squash to keep them separate)

The form without locations, and an example of the right-aligned button looking nice:

image

image

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -17,3 +17,7 @@ $govuk-font-url-function: "font-url";
 .govuk-phase-banner{
   margin-bottom: 25px;
 }
+
+.text-right{
+  text-align: right;
+}
--- a/app/views/ips/new.html.erb
+++ b/app/views/ips/new.html.erb
@@ -9,12 +9,14 @@
     <%= render "activation_notice" %>
 
     <%= form_with model: @ip do |form| %>
-      <div class="govuk-form-group <%= field_error(@ip, :location_id) %>">
-        <%= form.label :location_id, "Add an IP to an existing location", class: "govuk-label" %>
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -17,3 +17,7 @@ $govuk-font-url-function: "font-url";
 .govuk-phase-banner{
   margin-bottom: 25px;
 }
+
+.text-right{
+  text-align: right;
+}
```
emileswarts: Definitely sure nothing in the frontend styling we import that gives us this?

app/views/ips/new.html.erb
```diff
@@ -9,12 +9,14 @@
...
-      </div>
+      <% if @locations.present? %>
+        <div class="govuk-form-group <%= field_error(@ip, :location_id) %>">
+          <%= form.label :location_id, "Add an IP to an existing location", class: "govuk-label" %>
+          <%= form.select :location_id, @locations, { prompt: '-- select --' }, { class: "govuk-select" } %>
+        </div>
 
-      <p class="govuk-body"><strong>OR</strong></p>
+        <p class="govuk-body"><strong>OR</strong></p>
+      <% end %>
```
emileswarts: Nice fix 👍

app/assets/stylesheets/application.scss
kjdchapman: About 80% - had a look at some of the other admin tools and this seems to be what they were doing. Had a quick look through the Design System too, no mention of it.

emileswarts: Okay I see

PR #200, opened 2018-10-15
Revert "ensure the database is created before setting the environment"

cybojenix:
Reverts alphagov/govwifi-admin#197

Jenkins Docker plugin has a bug that breaks for multi-stage dockerfiles. You need to change how you reference the stages.
Reverting until time is made for implementing this workaround.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ lint: build
 
 .PHONY: test
 test: build launch_db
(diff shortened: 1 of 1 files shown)


cybojenix:
merging this PR into here:
https://github.com/alphagov/govwifi-admin/pull/201/

PR #205, opened 2018-10-15
Extract log search query validator

emileswarts:
This will be used to ensure that search queries are valid for viewing
logs.

### DIFF
--- a/spec/use_cases/administrator/validate_log_search_query_spec.rb
+++ b/spec/use_cases/administrator/validate_log_search_query_spec.rb
@@ -1,0 +1,35 @@
+describe UseCases::Administrator::ValidateLogSearchQuery do
+  let(:email) { 'test@example.com' }
+
+  context 'Valid search params' do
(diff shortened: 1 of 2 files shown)


spec/use_cases/administrator/validate_log_search_query_spec.rb
```diff
@@ -0,0 +1,35 @@
+describe UseCases::Administrator::ValidateLogSearchQuery do
+  let(:email) { 'test@example.com' }
+
+  context 'Valid search params' do
+    context 'valid username' do
+      it 'returns true' do
+        valid_params = [
+          { username: 'ABCDEF', ip: nil },
```
kjdchapman: Can I search for both at once?

emileswarts: The UI won't allow you to search both at once.
You will either search username by input field, or ip by link from the settings page.

PR #209, opened 2018-10-15
Search ip

emileswarts:
Organisation admins can search by IP and username.
Refactor the code to allow this and add a link next to the ips on the settings page, which will take admins to the logs for that IP address.


spec/use_cases/administrator/get_auth_requests_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  subject do
+    described_class.new(
+      authentication_logs_gateway: authentication_logs_gateway,
+      search_validator: search_validator
+    )
+  end
+
+  before do
+    allow(authentication_logs_gateway).to receive(:search)
```
kjdchapman: I think this line isn't needed - RSpec doubles are permissive by default. Just testing it out locally.

spec/use_cases/administrator/get_auth_requests_spec.rb
```diff
@@ -0,0 +1,51 @@
+describe UseCases::Administrator::GetAuthRequests do
+  let(:authentication_logs_gateway) { double }
```
kjdchapman: I would call this out as a `spy`, makes it a bit more obvious what's going on.
```ruby
let(:authentication_logs_gateway) { spy }
```

spec/use_cases/administrator/get_auth_requests_spec.rb
kjdchapman: Okay, they're _strict_ by default, so I'm _100% wrong_.

I would still not use this, I would put it on the spy declaration above:
```ruby
let(:authentication_logs_gateway) { spy(search: nil) }
```

(did test this out now, and it works)

naliwajek: `instance_spy` of that gateway would be even more strict.

PR #210, opened 2018-10-15
WIP: non-HTML5 validation for support requests

kjdchapman:
This page currently uses client-side HTML5 validation only, has no helpful error messages if they're disabled and the request is blank.


kjdchapman:
Only going to remove the validation for now, iceboxing the new validation.

PR #217, opened 2018-10-16
Link to the logs search path, not the results page

emileswarts:


app/views/application/_subnavigation.html.erb
```diff
@@ -15,7 +15,7 @@
       <%= link_to "MOU", organisations_mou_index_path, class: active_tab(organisations_mou_index_path) %>
       <%= link_to "Team members", team_members_path, class: active_tab(team_members_path) %>
       <%= link_to "Settings", ips_path, class: active_tab(ips_path) %>
-      <%= link_to 'Logs', logs_path, class: active_tab(logs_path) %>
+      <%= link_to 'Logs', logs_path, class: active_tab(search_logs_path) %>
```
kjdchapman: I think this should be
`<%= link_to 'Logs', search_logs_path, class: active_tab(search_logs_path) %>`

As-is only changes the css class, not the path

PR #220, opened 2018-10-16
Set service email on Organisation signup

tommotaylor:
screenshot 2018-10-16 at 14 39 55


emileswarts:
I wonder if we should put more emphasis on the fact that this inbox has to be shared and monitored.

PR #225, opened 2018-10-16
Redirect admin based on type of the MOU upload

naliwajek:


kjdchapman:
I'm going to be a pain.

Can we improve the specs somehow so they'll fail if this isn't working?

PR #226, opened 2018-10-16
Make IPs unique on model and db

tommotaylor:


spec/use_cases/radius/generate_ip_whitelist_spec.rb
```diff
@@ -32,28 +32,6 @@
     end
   end
 
-  context 'with duplicates in the database' do
```
kjdchapman: If you want to maintain testing this (because it's a pretty critical break), you could pass a stub gateway to this use case that spits out duplicates.

tommotaylor: Just had a look and it requires refactoring as this use case doesn't take a gateway.

PR #229, opened 2018-10-16
Full-width IPs table

kjdchapman:
image

(Also as a bonus tweak, this PR limits the length of service e-mails so that table doesn't look bad)

### DIFF
--- a/app/views/ips/_table.html.erb
+++ b/app/views/ips/_table.html.erb
@@ -1,23 +1,18 @@
-<h3 class="govuk-heading-m"><%= location.full_address %></h2>
 <table class="govuk-table">
-  <thead class="govuk-table__head">
-    <tr class="govuk-table__row">
-      <th class="govuk-table__header" scope="col">IP Address</th>
-      <th class="govuk-table__header"></th>
-      <th class="govuk-table__header"></th>
-    </tr>
-  </thead>
+  <caption class="govuk-table__caption">
+    <%= location.full_address %>
+  </caption>
   <tbody class="govuk-table__body" id="ips-table">
     <% location.ips.each do |ip| %>
       <tr class="govuk-table__row">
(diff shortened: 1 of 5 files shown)


app/views/ips/_table.html.erb
```diff
@@ -1,23 +1,18 @@
...
-    </tr>
-  </thead>
+  <caption class="govuk-table__caption">
+    <%= location.full_address %>
+  </caption>
   <tbody class="govuk-table__body" id="ips-table">
     <% location.ips.each do |ip| %>
       <tr class="govuk-table__row">
-        <td class="govuk-table__cell half-column" scope="row"><%= ip.address %></td>
+        <td class="govuk-table__cell""><%= ip.address %></td>
```
emileswarts: Rogue quote alert

app/views/ips/_table.html.erb
```diff
@@ -1,23 +1,18 @@
...
         <% if ip.available? %>
-          <td class="govuk-table__cell half-column">Available</td>
+          <td class="govuk-table__cell">Available</td>
         <% else %>
-          <td class="govuk-table__cell half-column unavailable-ip">Not available until 6am tomorrow
+          <td class="govuk-table__cell unavailable-ip">Not available until 6am tomorrow
           </td>
         <% end %>
-        <td class="govuk-table__cell half-column"><%= link_to 'Logs', logs_path(ip: ip.address) %></td>
+        <td class="govuk-table__cell"><%= link_to 'View activity', logs_path(ip: ip.address) %></td>
```
emileswarts: In my mind "Logs" is still more descriptive than "View activity"

kjdchapman: Mm, I think I should have tried "view logs" and got the best of both worlds

app/views/ips/_table.html.erb
kjdchapman: Sneaky quotes

PR #230, opened 2018-10-16
Force third-width columns, pull link right

kjdchapman:
Things don't look right on staging:

image

With this PR, they can look right:

image

### DIFF
--- a/app/assets/stylesheets/utilities/layout.scss
+++ b/app/assets/stylesheets/utilities/layout.scss
@@ -1,3 +1,7 @@
 .text-right {
   text-align: right;
 }
+
+.column-third {
+  width: 33%;
+}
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/utilities/layout.scss
```diff
@@ -1,3 +1,7 @@
 .text-right {
   text-align: right;
 }
+
+.column-third {
+  width: 33%;
+}
```
emileswarts: https://github.com/alphagov/govuk-frontend/blob/ad596de756ef51619c6e5bc8f22f3b93479773dd/package/objects/_grid.scss#L5

Does ".govuk-grid-column-one-third" do exactly this?

kjdchapman: I haven't got the grid system and tables to play nicely together yet - I'd like to sort out these tables "properly" in the near-future, get some text wrapping etc. working, but for now this at least repairs some of the damage in #229

PR #244, opened 2018-10-23
Add google analytics

georgeschena:
Add script to allow google analytics to collect user data

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,6 +1,17 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
+
+  <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -30,6 +30,17 @@
     <meta property="og:image" content="/assets/images/govuk-opengraph-image.png">
   </head>
 
+  <!-- Global site tag (gtag.js) - Google Analytics -->
```
tommotaylor: Needs to be within `<head>` just after the opening `<head>` tag: https://support.google.com/analytics/answer/1008080

config/site.yml
```diff
@@ -6,13 +6,17 @@ default: &default
   organisation_troubleshooting_link: 'https://docs.wifi.service.gov.uk/troubleshooting.html#troubleshooting'
 
 development:
+  google_analytics_script_id: 'UA-127779891-3'
```
tommotaylor: this ID doesn't exist, I think the staging code would be best.

PR #275, opened 2018-10-31
Fix layout of main navigation to be consistent

emileswarts:
The admin main navigation needs to match the product page and the tech
docs.  We are unable to use exactly the same styles as the tech docs and
product page have different stylesheets baked in.

This overwrites some of the default styles for the gov-uk frontend to
match these other 2 apps.


kjdchapman:
Oof, pixel-specific stuff. Does it work OK on mobile?

Or maybe with GA now, we can see if we even need to worry about that.

emileswarts:
@kjdchapman Can confirm that this still works on mobile, have tested it

PR #276, opened 2018-10-31
Lock user account after 10 failed login attempts

georgeschena:


tommotaylor:
Code looks good but my concern is that we are locking users out for an hour without telling them anything other than that their account is locked. 

We need something that tells them that they are locked out for an hour.

This actually makes me think that unlocking the account via email is better as it means they won't contact us to fast forward the one hour lockout.

We should also be reminding them that they can reset their password when they are getting close to getting locked out.

spec/features/authentication/lock_user_account_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  include_examples 'confirmation use case spy'
+  include_examples 'notifications service'
+
+  after { Timecop.return }
+
+  let(:correct_password) { 'password' }
+  let(:incorrect_password) { 'incorrectpassword' }
+
+  before do
+    sign_up_for_account(email: 'george@gov.uk')
```
tommotaylor: Better to use `sign_in_user(user)` here as it bypasses the fields and also means we don't need to user the notification spies.

Eg:
```
let(:user) { create(:user, :confirmed, password: "whatever", password_confirmation: "whatever") }

before do
  sign_in_user(user)
end
```

PR #277, opened 2018-10-31
Auto-log out user after period of inactivity

Renny4Real:
Add feature that logs out a user after one hour of inactivity.


spec/features/user_session_timeout_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    it_behaves_like 'not signed in'
+  end
+
+  context 'when user has signed in' do
+    before { sign_in_user user }
+
+    context 'and they have been inactive for 60 minutes' do
+      before do
+        Timecop.freeze(Time.now + (60 * 60))
```
craigjbass: you can also do

```ruby
  around(:each) do |example|
    Timecop.freeze(Time.now + (60 * 60)) { example.run }
  end
```

which lets you avoid having an after { } block

Renny4Real:
The spec specific to user inactivity passes, another completely separate test which uses timecop now fails.

spec/features/user_session_timeout_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'features/support/sign_up_helpers'
+require 'features/support/not_signed_in'
+require 'timecop'
+
+describe 'with devise timeoutable enabled' do
```
kjdchapman: I wouldn't describe the technology used in the top-level part of these specs - I'd talk in an imperative way like I do with git commits , e.g. `describe 'logout users after long inactivity'`. 

Useful way to remember this is
> If I get this test passing, the application can/will now _do the described thing_

> If I get this test passing, the application will now _logout users after long inactivity_

i'd use the same (or very similar) description for the file name, too.

In general, if I see a `*with* some context` block, I expect to see that context being set up right there in the test - e.g.
```ruby
describe 'with a user' do
  let(:user) { create(:user) }
  # some expectations
end
```

spec/features/user_session_timeout_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  after { Timecop.return }
+
+  context 'when user has not signed in' do
```
kjdchapman: Is this covered by other tests well enough? Don't think we're providing a unique context here.

spec/features/user_session_timeout_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    before { visit root_path }
+
+    it_behaves_like 'not signed in'
+  end
+
+  describe 'user sessions timeout' do
```
kjdchapman: I wouldn't include this wrapping describe block, I don't think it helps to describe what we're testing better.

spec/features/user_session_timeout_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    context 'when user has signed in' do
```
kjdchapman: Given you don't test this context without combining it with the one on [line 23](https://github.com/alphagov/govwifi-admin/pull/277/files#diff-fa5dbd2027b8fe84348baeaa0a15255bR23), I would just have one describe block for the stuff between line 15 and line 37, something like
```ruby
context 'when a logged in user has been inactive for 60 minutes'
```

spec/features/user_session_timeout_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      let(:user) { create(:user, :confirmed, :with_organisation) }
+
+      before do
+        sign_in_user user
+        visit root_path
+      end
+
+      context 'and they have been inactive for 60 minutes' do
```
kjdchapman: Does it make sense to test the negative case? What happens if a user has been inactive for 59 minutes?

I often like to express the "edges" in acceptance tests - 59 minutes they're still logged in, 60 minutes they're logged out - as I find it explains the feature more clearly.

Not sure if it's worth it here, just a thought.

app/views/layouts/_flash_notices.html.erb
```diff
@@ -1,4 +1,5 @@
 <% flash.each do |name, msg| -%>
+  <% next if name == 'timedout'%>
```
kjdchapman: nitpick, spacing:
```suggestion
  <% next if name == 'timedout' %>
```

kjdchapman:
:+1: for the renaming!

PR #280, opened 2018-11-05
Lock user account after 10 failed login attempts

georgeschena:


app/mailers/authentication_mailer.rb
```diff
@@ -32,6 +33,19 @@ def reset_password_instructions(record, token, opts = {})
     )
   end
 
+  def unlock_instructions(record, token, _opts = {})
+    unlock_link = unlock_url(record, unlock_token: token)
+    template_id = GOV_NOTIFY_CONFIG['unlock_account']['template_id']
+
+    UseCases::Administrator::UnlockAccountEmail.new(
```
kjdchapman: I would call this `SendUnlockEmail` - I think this can be confused with a use case that unlocks an account.

config/initializers/devise.rb
```diff
@@ -227,41 +227,41 @@
...
+  config.maximum_attempts = 10
 
   # Time interval to unlock the account if :time is enabled as unlock_strategy.
   # config.unlock_in = 1.hour
 
   # Warn on the last attempt before the account is locked.
-  # config.last_attempt_warning = true
+  config.last_attempt_warning = true
 
   # ==> Configuration for :recoverable
```
kjdchapman: I wouldn't change these config settings (assuming all your tests still pass with them off?). Don't want to tweak `recoverable` as part of this PR if possible.

config/locales/devise.en.yml
```diff
@@ -10,7 +10,7 @@ en:
       already_authenticated: ""
       inactive: "Your account is not activated yet."
       invalid: "Invalid %{authentication_keys} or password."
-      locked: "Your account is locked."
+      locked: "Your account has been locked. Reset your password"
```
kjdchapman: This messaging is probably wrong now we're not guiding people to resetting their password.

Not sure what the right messaging is - "Your account is locked." might be a little unhelpful, but "Your account is locked, check your e-mail" gives a hint to the attacker that they can get access by e-mail.

spec/features/authentication/lock_user_account_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'support/unlock_account_use_case_spy'
+require 'support/unlock_account_use_case'
+
+describe "Locking a users account after ten failed login attempts" do
```
kjdchapman: I would tweak some of these context blocks / it descriptions - might be easier to talk through it together?

spec/features/authentication/lock_user_account_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'support/unlock_account_use_case_spy'
+require 'support/unlock_account_use_case'
```
kjdchapman: Can we make the naming of these imports more expressive?

PR #282, opened 2018-11-06
Fix large organisation name disrupting the sub nav

georgeschena:
When a user has a large organisation name it affects the alignment of the sub navigation bar. This change allows an organisation to have a large name and maintain a professional look for the sub navigation bar.

Before: 
screenshot 2018-11-06 at 15 43 10

After: 
screenshot 2018-11-06 at 15 42 57


tommotaylor:
Doesn't really work:

screenshot 2018-11-06 at 16 14 03

tommotaylor:
I would suggest something like flex box that will always hold that right hand content bottom aligned.

PR #262, opened 2018-10-29
Add Google Tag Manager Scripts

Renny4Real:
Add tag manager script code within `<head> `and `<body>` tags.

PR #283, opened 2018-11-06
Implement suggested changes to the Google Tag Manager code

Renny4Real:
Add some changes to the codebase surrounding google tag manager such as extracting the javascript tags into a separate file.

Changes were based on suggestions in PR #262 

Purpose: To make GTM implementation easier to understand and change.

Suggestions/feedback on changes would be appreciated.

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -7,6 +7,7 @@ Rails.application.config.assets.precompile += %w(
   govuk-logotype-crown.png
   application_ie8.css
   html5shiv.js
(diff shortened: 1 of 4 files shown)


config/initializers/assets.rb
```diff
@@ -7,6 +7,7 @@
   govuk-logotype-crown.png
   application_ie8.css
   html5shiv.js
+  googletagmanager.js
```
tommotaylor: Is this needed?

PR #286, opened 2018-11-07
Make the root page for super admins the list of organisations

Renny4Real:
Homepage for 'Super' admins would now be the list of organisations page.
screenshot 2018-11-07 at 13 36 21

UPDATE: Organisation tab has been put back:
screenshot 2018-11-07 at 14 12 31


app/views/application/_subnavigation.html.erb
```diff
@@ -8,7 +8,6 @@
   <div class="govuk-grid-column-two-thirds govuk-!-margin-bottom-0 govuk-!-padding-right-0 flex-end">
     <nav class="govuk-body govuk-!-margin-bottom-0 ">
     <% if current_user.admin? %>
-      <%= link_to "Organisations", organisations_path, class: active_tab(organisations_path) %>
```
tommotaylor: Why have you removed this?

spec/features/admin/sign_in_as_admin_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'features/support/sign_up_helpers'
+require 'features/support/not_signed_in'
+
+describe 'signing as an admin user' do
+  let!(:user) { create(:user, :confirmed, admin: true) }
+
+  context 'when visiting the home page' do
+    before do
+      5.times { create(:organisation) }
```
tommotaylor: I don't think you need to test it with 5 different organisations. You get the same test coverage out of one.

spec/features/admin/view_organisation_tab_spec.rb
```diff
@@ -12,11 +12,11 @@
     let(:organisation) { create(:organisation) }
     let(:user) { create(:user, :confirmed, email: 'me@example.gov.uk', organisation: organisation, admin: true) }
 
-    it 'displays the organisation tab' do
+    it 'displays not the organisation tab' do
```
kjdchapman: This reads weirdly to me.

```suggestion
    it 'does not display the organisation tab' do
```

app/views/application/_subnavigation.html.erb
Renny4Real: Figured since the homepage redirects to the organisation path, that tab would not be needed.

spec/features/admin/sign_in_as_admin_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      sign_in_user user
+      visit root_path
+    end
+
+    it 'redirects to the organisations page' do
+      expect(page.current_path).to eq(organisations_path)
```
kjdchapman: I wouldn't test this, I'd be happy relying on the test below (checking the content of the page) as sufficient for proving this feature is working.

This test feels brittle to me, as it's tied to an implementation detail (the content we're looking for being at a particular URL).

spec/features/admin/view_organisation_tab_spec.rb
Renny4Real: Yeah, forgot to read over 😁

spec/features/admin/sign_in_as_admin_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    it 'shows list of signed organisations on the home page' do
+      Organisation.all.each do |organisation|
```
tommotaylor: This is also checking for the organisation that the super admin user belongs to, which only passes as its displayed top left in the subnav. This spec is specifically looking at the table though.

app/views/application/_subnavigation.html.erb
tommotaylor: But then you wouldn't be able to get back to it. That aside I think its also important enough to be a subnav item.

spec/features/admin/sign_in_as_admin_spec.rb
tommotaylor: Actually I'm wrong, this will show all organisations as the super admins organisation will be one of the signed up organisations.

PR #293, opened 2018-11-07
Improve organisations table for super-admins

kjdchapman:
screenshot 2018-11-08 at 11 04 07


app/assets/stylesheets/components/organisations_table.scss
```diff
@@ -0,0 +1,19 @@
+.name-col {
+  width: 35%;
```
kjdchapman: Last time we talked about this, Emile had some feelings on not using custom styles - not sure what the best approach is if we want to tweak this table.

app/views/organisations/index.html.erb
```diff
@@ -1,33 +1,35 @@
-<h1 class="govuk-heading-l">Signed Up Organisations: <%= @all_organisations.count %></h1>
+<h1 class="govuk-heading-l">All organisations</h1>
+<h3 class="govuk-heading-m">
+  <%= pluralize(@all_organisations.count, 'organisation') %> are signed up to GovWifi
```
kjdchapman: Might not even be necessary to pluralize here, given I've assumed "are" rather than "is". There'll always be more than one!

app/views/organisations/index.html.erb
```diff
@@ -1,33 +1,35 @@
...
-            <% end %>
-          </div>
+        <td class="govuk-table__cell name-col" scope="row">
+          <%= organisation.name %>
+        </td>
+        <td class="govuk-table__cell created-date-col" scope="row">
+          <%= organisation.created_at.strftime('%e %b %Y') %>
+        </td>
+        <td class="govuk-table__cell govuk-table__cell mou-col" scope="row">
+          <%= organisation.signed_mou.present? ? 'uploaded' : 'none' %>
```
kjdchapman: I don't think `.present?` works how I expect on an `activestorage` method.

spec/features/admin/view_organisation_list_spec.rb
```diff
@@ -2,26 +2,66 @@
...
+        expect(page).to have_content 'Fake Org'
+      end
 
-    let!(:user) { create(:user, :confirmed, admin: true) }
+      it 'shows when they signed up' do
+        expect(page).to have_content('1 Feb 2014')
+      end
+
+      it 'shows they have 10 locations' do
+        expect(page).to have_content('10')
```
kjdchapman: Testing this on lower numbers gives false positives, as the page has numbers all over it in dates etc.

I'd switch this to checking more specific parts of the page (e.g. a cell with some class), even if it's a bit brittle, just didn't get to it.

spec/features/admin/view_organisation_list_spec.rb
```diff
@@ -2,26 +2,66 @@
...
+      end
+
+      it 'shows they have 11 IPs' do
+        expect(page).to have_content('11')
+      end
 
-    context 'when admin navigates to the list page' do
-      it 'renders the names of the signed up organisations' do
-        sign_in_user user
+      xit 'shows they have an MOU' do
```
kjdchapman: Didn't implement this yet. Probably want to check the negative as well.

app/views/organisations/index.html.erb
tommotaylor: Just FYI, the method is `.attached?`

PR #295, opened 2018-11-09
Sort the table view by name or creation date by clicking table header

Renny4Real:
Headers become links that when clicked cause the table to sort in ascending or descending order. The sort direction is toggled each time the same header is clicked.

screenshot 2018-11-09 at 11 32 48
^^^ The list of organisations above is in descending alphabetical order and the clickable links can be seen.


app/controllers/admin/organisations_controller.rb
```diff
@@ -1,5 +1,26 @@
 class Admin::OrganisationsController < AdminController
+  helper_method :sort_column, :sort_direction
+  before_action :index, :authorise_admin
```
kjdchapman: Think you won't need this now this inherits from `AdminController`

kjdchapman:
```suggestion

```

app/controllers/admin/organisations_controller.rb
```diff
@@ -1,5 +1,26 @@
...
   def index
-    @organisations = Organisation.order(:name).all
+    @organisations = Organisation.order("#{sort_column} #{sort_direction}").all
+  end
+
+private
+
+  def authorise_admin
+    redirect_to root_path unless current_user.admin?
+  end
```
emileswarts: Does the AdminController which we inherit from here do this authorisation already?

spec/features/admin/sort_view_in_organisation_list_spec.rb
```diff
@@ -0,0 +1,34 @@
+describe 'sorting the columns in the organisation list' do
+  context 'when admin views the names of organisations' do
+    let!(:organisation_1) { create(:organisation, name: "ABC", created_at: Date.yesterday) }
```
kjdchapman: I've found `let!` to be less clear than the alternatives, most of the time.

Since we don't actually reference `organisation_1` as a variable in the specs, you could move creation the organisations to the `before` block:

```ruby
before do
  create(:organisation, name: "ABC", created_at: Date.yesterday)
  create(:organisation, name: "DEF", created_at: 1.month.ago)
  create(:organisation, name: "XYZ", created_at: 2.years.from_now)

  sign_in_user user
  visit root_path
end
```

I think that's clearer than having two separate blocks as setup.

kjdchapman: I'd use real-looking data in acceptance specs, as they're often the first time a reader will encounter concepts, and it helps explain them better.

```suggestion
    let!(:organisation_1) { create(:organisation, name: "Department of Silly Hats", created_at: Date.yesterday) }
```

spec/features/admin/sort_view_in_organisation_list_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      visit root_path
+    end
+
+    it 'can sort the organisations by names in descending order' do
+      click_link 'Name'
+
+      expect(page.body).to match(/XYZ.*DEF.*ABC/m)
+    end
+
+    context 'can toggle the order of the dates' do
```
kjdchapman: I would describe behaviour by describing what the component *does*, rather than what it "should" or "can" do, as you end up with less noise around the key information.

I'm a little confused by which parts of this are `context` and which are `it` blocks, I don't think the doc output would be as clear as it could be.

Think the structure should be something like:
```ruby
context 'showing the list of organisations'
  context 'by default'
    it 'is sorted by name, ascending'
 
  context 'with name selected once'
    it 'is sorted by name, descending'

  context 'with date selected once'
    it 'is sorted by date, from oldest to newest'

  context 'with date selected twice'
    it 'is sorted by date, from newest to oldest'
```

spec/features/admin/sort_view_in_organisation_list_spec.rb
```diff
@@ -0,0 +1,40 @@
+describe 'sorting the columns in the organisation list' do
+  context 'when admin views the table with organisation data' do
+    let!(:user) { create(:user, :confirmed, admin: true) }
+
+    before do
+      create(:organisation, name: "Department of Apple Cakes", created_at: Date.yesterday)
+      create(:organisation, name: "Department of Silly Hats", created_at: 1.month.ago)
+      create(:organisation, name: "Department of Xylophones", created_at: 2.years.from_now)
+
+      sign_in_user user
```
kjdchapman: I think this file needs `require 'features/support/sign_up_helpers'` at the top, as that's where this `sign_in_user` method is defined. 

On my machine, this spec file fails if I run it in isolation.

spec/features/admin/sort_view_in_organisation_list_spec.rb
```diff
@@ -0,0 +1,40 @@
+describe 'sorting the columns in the organisation list' do
+  context 'when admin views the table with organisation data' do
+    let!(:user) { create(:user, :confirmed, admin: true) }
```
kjdchapman: nit: this doesn't need to be a `let!`

PR #296, opened 2018-11-12
Add basic permissions to users

kjdchapman:
This PR allows you to check whether a User `can_manage_locations` or `can_manage_team`, and makes it so all new Users have both permissions, but does not implement checking these permissions anywhere.

At one point this did have more methods for the permissions everyone has - like `view_locations` - but given that _everyone_ has these for the foreseeable future, leaving them out for now.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -104,21 +104,6 @@ describe User do
       it 'can manage locations' do
         expect(subject.can_manage_locations?).to be_truthy
       end
-
-      context 'after reload' do
-        let(:reloaded_user) do
-          subject.save!
-          User.find(subject.id)
-        end
-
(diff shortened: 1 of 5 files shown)


spec/models/user_spec.rb
```diff
@@ -92,4 +92,33 @@
...
+
+      it 'can manage team members' do
+        expect(subject.can_manage_team?).to be_truthy
+      end
+
+      it 'can manage locations' do
+        expect(subject.can_manage_locations?).to be_truthy
+      end
+
+      context 'after reload' do
```
kjdchapman: Not sure if this test is worth it or not - was useful as I was switching away from the in-memory implementation, but might be excessive / misleading now.

PR #298, opened 2018-11-12
Restrict managing team to users with permissions

emileswarts:
We don't want everyone to be able to add and remove team members.
Use the permission system to check that a user is allowed to manage the
team before displaying the button to the invite page.

Also redirect to the root page if the URL was visited directly.

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -1,4 +1,5 @@
 class Users::InvitationsController < Devise::InvitationsController
+  before_action :authorise_manage_team, only: %i(create new)
   before_action :add_organisation_to_params, :validate_invited_user, only: :create
 
 private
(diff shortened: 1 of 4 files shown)


spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -4,102 +4,134 @@
...
+            }.to change { User.count }.by(0)
+            expect(InviteUseCaseSpy.invite_count).to eq(0)
+            expect(page).to have_content("Email can't be blank")
+          end
         end
-      end
 
-      context "with an email that already exists" do
-        let(:invited_user_email) { user.email }
+        context 'tells user that email must be a valid gov.uk email' do
```
tommotaylor: This looks like it already exists on Line 47: `context 'with non gov.uk email address' do`

PR #299, opened 2018-11-12
Sort list of organisations by MoU upload date

Renny4Real:
screenshot 2018-11-12 at 17 13 18

screenshot 2018-11-12 at 17 13 24

### DIFF
--- a/spec/features/admin/sort_view_in_organisation_list_spec.rb
+++ b/spec/features/admin/sort_view_in_organisation_list_spec.rb
@@ -1,18 +1,13 @@
 require 'features/support/sign_up_helpers'
 
-describe 'sorting the columns in the organisation list' do
+describe 'sorting the columns in the organisation list', focus: true do
   context 'when admin views the table with organisation data' do
     let(:user) { create(:user, :confirmed, admin: true) }
 
     before do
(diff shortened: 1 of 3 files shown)


spec/features/admin/sort_view_in_organisation_list_spec.rb
```diff
@@ -38,5 +43,23 @@
...
+      it 'sorts the organistions from oldest to newest when MoU signed is clicked once' do
+        click_link 'MoU Signed'
+
+        expect(page.body).to match(/Apple Cakes.*Xylophones.*Silly Hats/m)
+      end
+
+      it 'sorts the organistions from newest to oldest when MoU signed is clicked twice' do
+        2.times { click_link 'MoU Signed' }
+
+        expect(page.body).to match(/Silly Hats.*Xylophones.*Apple Cakes/m)
```
tommotaylor: My only comment would be trying to make it more obvious what these organosations are. For instance you could call them: `No Mou`, `New MoU`, `Old Mou`. Or you could name them as variables and have `no_mou.name`, `new_mou.name` and `old_mou.name`. 

Basically just makes it a bit easier to read and understand the order to other developers.

PR #300, opened 2018-11-13
Limit width of org names

kjdchapman:
OLD:

image

NEW

image

(thanks @tommotaylor for the govuk-width tip - I've scourged the custom one as part of this PR)

### DIFF
--- a/app/assets/stylesheets/utilities/layout.scss
+++ b/app/assets/stylesheets/utilities/layout.scss
@@ -2,10 +2,6 @@
   text-align: right;
 }
 
-.column-third {
-  width: 33%;
-}
-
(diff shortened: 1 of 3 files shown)


tommotaylor:
Code looks good but I would suggest making it wider. All the other columns will never be very wide, they all have set data. Looking at the screenshot "MOU Signed" has as much real estate as name.

Nice refactor too. The less CSS the better!

kjdchapman:
Good shout - I've bumped it up to half instead of a third.

kjdchapman:
Whoops, did forget a second screenshot - this is what it looks like with 50%: 
image

PR #302, opened 2018-11-13
Fix display of Overview page on Internet Explorer

Renny4Real:
Before:
screenshot 2018-11-13 at 15 04 27

screenshot 2018-11-13 at 14 19 00


kjdchapman:
I think you're using 4-space indentation rather than 2-space indentation for this file - we haven't made any formal agreement to use 4-space, but pretty much the whole app is using it, I believe.

Having 4-space indentation is making the diff show more changes than intended, which is making it a little harder to understand exactly what's been changed - would you mind switching back to 2-space?

PR #303, opened 2018-11-13
Team permissions display

kjdchapman:
Overhauled the team members page to explain permissions. 

~This is all hard-coded at the moment, rather than working off our permissions model - wanted to split that out into another PR~ 
this turned out to be easy, so I just did it!

This PR uses only `govuk-design-system` classes (admittedly with a bunch of the `!-override` ones), was heavily influenced by Notify, but couldn't easily port their styling.

## Old

image

## New

image

## Mobile

image


kjdchapman:
Got a simple version of showing the actual saved permissions on the model working:

image

Struggled with writing a test around it, would love to hear opinions.

PR #312, opened 2018-11-19
Remove IPs

tommotaylor:
- All exists on the index page now

screenshot 2018-11-20 at 13 55 55


app/controllers/ips_controller.rb
```diff
@@ -41,7 +48,7 @@ def publish_for_performance_platform
         bucket: ENV.fetch('S3_PUBLISHED_LOCATIONS_IPS_BUCKET'),
         key: ENV.fetch('S3_PUBLISHED_LOCATIONS_IPS_OBJECT_KEY')
       ),
-      source_gateway: Gateways::Ips.new
+      source_gateway: Gateways::IpsGateway.new
```
tommotaylor: Had to rename this because of naming conflicts on `Ips` constant.

app/controllers/ips/remove_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Ips::RemoveController < ApplicationController
```
kjdchapman: I would put this on the `IpsController`.

The resource we're dealing with here is `ip`, and all this action does is `show` it - even though we know you can only delete it from this page. I don't think the extra controller makes it clearer.

app/controllers/ips_controller.rb
kjdchapman: (It's possible to fix this, but I think it's more faff than just doing this rename dodge)

spec/features/ips/remove_an_ip_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'features/support/sign_up_helpers'
+
+describe 'Remove an IP' do
+  let!(:user) { create(:user, :confirmed) }
```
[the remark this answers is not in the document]
kjdchapman: I share this concern, I think this spec will be clearer without `let!`.

kjdchapman:
Overall, this looks good to me - the separate step makes for a simple implementation and an easily explained UI.

I don't expect many users to delete many, many IPs at once (and research can correct us if so).

I'm happy to give it another look and approve when you've added in the extra spec you wanted.

app/controllers/ips/remove_controller.rb
tommotaylor: Actually this is intentional, and kind of the nexus of the whole thing. Basically we need a 2-step process to delete an IP. Step one is clicking remove, step two is confirming the remove. We can't have a pop-up or JS, so there needs to be a page refresh, which means there needs to be two controller endpoints handling this. (It could be done as two steps via the same endpoint but thats really ugly).

What this does is keep all our routes and controllers RESTful. It does this by creating the concept of an "IP removal". Step one takes you to an "Ip removal show page", or in other words "shows you the ip you want to remove", and then step two does the actual delete on the IP.

If this was all done on the `IpsController` it would not be RESTful as the actions would be:
- `ips#display_removal` (renders a page or partial that shows a delete button)(not RESTful)
- `ips#destroy` (deletes the ip)

Instead we have more, smaller, RESTful controllers:
- `ips/remove#show`  (renders a page or partial that shows a delete button)
- `ips#destroy`  (deletes the ip)
 
This pattern of many smaller RESTful controllers is a Rails pattern to prevent controllers from getting out of hand. http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

kjdchapman: I understand (and agree with) the desire to make this RESTful, and the linked blog post. I don't think the creation of an extra resource (the abstract "IP deletion request") is the right choice in this case, though.

If this was done on the IpsController, the actions could be
- `ips#show` (renders the page showing the IP details and a delete button)
- `ips#destroy` (deletes the ip)

Maybe you're not happy with this being a `show` page, because it mainly emphasises this one use case - deleting the IP?

I think having a separate controller is less clear than the `show` page being mainly about deletion.

tommotaylor: Yeah this page just doesn't feel like an IP show page. I agree having the `ips#show` display this page is simpler, but then the routes and url don't really make sense, eg:

`ips#show`:
- `ip_path(ip)`
- `wifi.service.gov.uk/ips/2345`

`ips/remove#show`:
- `ip_remove_path(ip)`
- `wifi.service.gov.uk/ips/2345/remove`

spec/features/ips/remove_an_ip_spec.rb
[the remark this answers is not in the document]
tommotaylor: The reason I've done it this way is because I need access to the `ip` variable in the expectations, and this ip needs to belong to the user's organisation. Doing this with a before block looks more confusing to me, eg:

```
before do
  user = create(:user)
  sign_in_user user
  create(:location, organisation: user.organisation)
  create(:ip, location: location)
end

  expect(page).to have_content(Ip.first.address)
```

app/controllers/ips/remove_controller.rb
kjdchapman: I'm happy with either of those routes, but I'm also happy if you want to separate this out.

app/controllers/ips_controller.rb
emileswarts: What is the Ips constant?  Aren't namespaces solving this already?

spec/requests/ips/delete_spec.rb
```diff
@@ -0,0 +1,29 @@
+describe "DELETE /ips/:id", type: :request do
+  let(:user) { create(:user, :confirmed) }
+  let(:location) { create(:location, organisation: user.organisation) }
+  let!(:ip) { create(:ip, location: location) }
+
+  before do
+    https!
```
kjdchapman: What does this do?

kjdchapman:
I like the change to get this all on the same page - still happy for this to go ahead!

spec/requests/ips/delete_spec.rb
tommotaylor: It makes the requests use HTTPS. Without this the requests are HTTP, which means they redirect to HTTPS and every response is a 307 redirect.

app/assets/stylesheets/components/red_button.scss
```diff
@@ -0,0 +1,7 @@
+.red-button {
+  background-color: $govuk-error-colour;
+
+  &:hover {
+    background-color: #990c1a;
```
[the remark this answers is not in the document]
tommotaylor: Couldn't see one. This is straight from Notify. We could use some kind of shading, or % darkening instead.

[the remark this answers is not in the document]
emileswarts: ah ok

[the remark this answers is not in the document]
tommotaylor: Eg: `darken( $base-color, 10% )`

app/controllers/ips_controller.rb
```diff
@@ -78,7 +89,16 @@ def params_with_existing_location
     ip_params.except(:location_attributes)
   end
 
-  def authorise_manage_locations
-    redirect_to(root_path) unless current_user.can_manage_locations?
+  def set_ip_to_delete
+    ip = Ip.find(params[:ip_id])
```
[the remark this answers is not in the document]
tommotaylor: Yes, good spotting

app/assets/stylesheets/components/red_button.scss
```diff
@@ -0,0 +1,7 @@
+.red-button {
+  background-color: $govuk-error-colour;
+
+  &:hover {
+    background-color: darken($govuk-error-colour, 10%);
```
emileswarts: nice nice

PR #314, opened 2018-11-20
Make ‘forgot password’ link the correct blue

quis:
There are probably other instances; this is just the first one I came across.

image


tommotaylor:
Closed in favour of https://github.com/alphagov/govwifi-admin/pull/320

PR #315, opened 2018-11-20
Allow setting permissions for a new invite

emileswarts:
When you invite another network administrator, you can set their
permissions to manage their team and ips/locations.

### DIFF
--- a/app/views/users/invitations/new.html.erb
+++ b/app/views/users/invitations/new.html.erb
@@ -28,6 +28,27 @@
         </div>
 
         <div class="actions">
+          <div class="govuk-form-group">
+            <fieldset class="govuk-fieldset" aria-describedby="waste-hint">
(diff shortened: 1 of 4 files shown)


app/views/users/invitations/new.html.erb
```diff
@@ -27,6 +27,27 @@
...
+                  <div class="govuk-checkboxes__item">
+                    <%= p.check_box :can_manage_locations, class: 'govuk-checkboxes__input' %>
+                    <label class="govuk-label govuk-checkboxes__label"><%= p.label :can_manage_locations, 'Manage locations' %></label>
+                  </div>
+                <% end %>
+              </div>
+            </fieldset>
+          </div>
+        </div>
+
```
emileswarts: The permission building needs some work to allow us to share this markup.
I will go for that in a separate PR.

PR #316, opened 2018-11-20
Remove team members

georgeschena:
Uses query string to display partial:

screenshot 2018-11-21 at 12 11 08


app/controllers/permission_controller.rb
```diff
@@ -3,6 +3,7 @@ class PermissionController < ApplicationController
 
   def edit
     @permission = Permission.find(params.fetch(:id))
+    @show_remove_partial = request.path.include?("remove")
```
emileswarts: I personally would favour a query param over string matching.

Means we can sidestep setting this instance variable in the controller and just have the view look for

<% if params[:remove].present? %>
...

Would also be able to remove our new route.

Know there has been a bit of back and forth on this, but it feels like we can get away with less code.

tommotaylor: Done

PR #335, opened 2018-11-28
Replace end user support links

Renny4Real:
Based on suggestions from the content designers to separate end-user support and admin-user support, end user support links redirect to the GOV.uk page and not to the GovWifi Product page.


tommotaylor:
@Renny4Real I approved but saw this link that may be better for "troubleshooting" https://www.gov.uk/guidance/solve-problems-with-connecting-to-govwifi

PR #339, opened 2018-11-29
Improve team members UX

georgeschena:
Change "resend invite" from button to link 
Pull "edit permissions" to be in right-hand column instead of below
Made changes to edit team member page with rewording and styling.

After:
screenshot 2018-11-29 at 14 20 22

screenshot 2018-11-29 at 14 20 39

### DIFF
--- a/app/assets/stylesheets/components/team_members.scss
+++ b/app/assets/stylesheets/components/team_members.scss
@@ -4,6 +4,7 @@
   cursor: pointer;
   text-decoration: underline;
   color: $govuk-link-colour;
+  background-color: transparent;
 
   &:hover {
     color: $govuk-link-hover-colour;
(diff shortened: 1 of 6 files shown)


kjdchapman:
Would it be better to have the 'Resend invite' link somewhere else? I think it gets lost where it is right now. 

What about:
> ### Invitation pending ([resend invite]())
> test@gov.uk


Interestingly, Notify doesn't even have this "retry" functionality, just a cancel:
image

kjdchapman:
Good to get another screenshot if the layout changes to address feedback.

(yay this right bit is now aligned!):
image

app/assets/stylesheets/components/team_members.scss
```diff
@@ -4,6 +4,7 @@
   cursor: pointer;
   text-decoration: underline;
   color: $govuk-link-colour;
+  background-color: transparent;
```
kjdchapman: Whoops, I skim-missed this. What does it do?

PR #340, opened 2018-11-29
Ensure error when password is blank when signing up

emileswarts:
This validation was removed when the password confirmation was removed.
Restore this and print the error in the form if the password is left
blank.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,6 +11,9 @@ class User < ApplicationRecord
     :rememberable, :trackable, :timeoutable, :validatable, :lockable
 
   validates :name, presence: true, on: :update
+  validates :password, presence: true,
+    length: { within: 6..80 },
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -11,6 +11,10 @@ class User < ApplicationRecord
     :rememberable, :trackable, :timeoutable, :validatable, :lockable
 
   validates :name, presence: true, on: :update
+  validates :password, presence: true,
+    confirmation: true,
```
kjdchapman: Should this line be here? Would have expected it to fail some tests if not, but surprises me.

tommotaylor: I would have thought removing the `confirmation: true` is what we want. I do recall this being a bit buggy when implementing it. But now we don't have confirmations we don't need it at all

PR #341, opened 2018-11-29
Add IP to location

tommotaylor:
This is the first cut using a `hidden_field` and utilising the existing `IpsController#create` action.

Next step will be using the `Locations::IpsController#create` action and refactoring the "after creating an IP" logic into a re-useable service. This won't change the user experience, it will just be a refactor.

screenshot 2018-11-29 at 15 34 14
screenshot 2018-11-29 at 16 06 50

### DIFF
--- a/app/views/ips/_table.html.erb
+++ b/app/views/ips/_table.html.erb
@@ -1,10 +1,13 @@
 <table class="govuk-table govuk-!-margin-bottom-8">
   <caption class="govuk-table__caption">
     <%= location.full_address %>
-    <p class="govuk-body govuk-!-margin-bottom-2">
-      <%= link_to "+ add IP", new_location_ip_path(location), class: "govuk-link" %>
-    </p>
+    <% if current_user.can_manage_locations? %>
+      <p class="govuk-body govuk-!-margin-bottom-2">
+        <%= link_to "+ add IP", new_location_ip_path(location), class: "govuk-link" %>
+      </p>
+    <% end %>
   </caption>
+  
   <tbody class="govuk-table__body" id="ips-table">
     <% location.ips.each do |ip| %>
       <tr class="govuk-table__row">
--- a/app/views/ips/_table.html.erb
+++ b/app/views/ips/_table.html.erb
@@ -1,6 +1,9 @@
-<table class="govuk-table">
-  <caption class="govuk-table__caption">
+<table class="govuk-table govuk-!-margin-bottom-8">
+  <caption class="govuk-table__caption"">
     <%= location.full_address %>
+    <p class="govuk-body govuk-!-margin-bottom-2">
+      <%= link_to "+ add IP", new_location_ip_path(location), class: "govuk-link" %>
+    </p>
   </caption>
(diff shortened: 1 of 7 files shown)


app/views/ips/_table.html.erb
```diff
@@ -1,6 +1,9 @@
-<table class="govuk-table">
-  <caption class="govuk-table__caption">
+<table class="govuk-table govuk-!-margin-bottom-8">
+  <caption class="govuk-table__caption"">
```
kjdchapman: nit: typo double-quote
```suggestion
  <caption class="govuk-table__caption">
```

app/views/locations/ips/new.html.erb
```diff
@@ -0,0 +1,23 @@
+<%= render "layouts/form_errors", resource: @ip %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= link_to "Back to list", ips_path, class: "govuk-back-link" %>
```
kjdchapman: Can we anchor back to the location? Not sure if it's available.

spec/features/ips/add_ip_to_location_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'features/support/sign_up_helpers'
+require 'features/support/errors_in_form'
+
+describe 'Add an IP to my account' do
+  context 'and the new location is invalid' do
+    let!(:user) { create(:user) }
```
kjdchapman: Pretty sure this works without the `!`.

```suggestion
    let(:user) { create(:user) }
```

kjdchapman:
I like this approach. It's clear what's going on in the code, and the user experience is what we want.

There's some duplication between the two ways of adding an IP, but I think the complication we'd introduce by de-duping it isn't worth it.

app/views/locations/ips/new.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <h2 class="govuk-heading-l">Add an IP address to <%= @location.address %></h2>
+
+    <%= render "ips/activation_notice" %>
+
+    <%= form_with model: @ip do |form| %>
+      <div class="govuk-form-group <%= field_error(@ip, :address) %>">
+        <%= form.label :ip, "Enter IP address (IPv4 only)", class: "govuk-label" %>
+        <%= form.text_field :address, id: :address, autocomplete: "off", class: "govuk-input" %>
+      </div>
+      <%= form.hidden_field :location_id, value: @location.id %>
```
tommotaylor: This is the contentious line.

kjdchapman: I like it, again, it's simple and clear.

Maybe I'm missing something though. Can you see problems with doing it this way?

tommotaylor: Hidden fields are generally a bad idea as you can manipulate the html to post a different ID. And then on top of this I'm using one controller action for these two forms, but their responsibility should probably be split between two different controllers.

But I think I'll stick with this solution for now, it gets the user flow right, then refactor to two controllers afterwards.

PR #345, opened 2018-12-02
Move RADIUS secret keys to IP's page

georgeschena:
screenshot 2018-12-03 at 11 41 30
screenshot 2018-12-03 at 11 41 37

### DIFF
--- a/app/views/ips/_table.html.erb
+++ b/app/views/ips/_table.html.erb
@@ -4,7 +4,7 @@
   <details class="govuk-details govuk-!-margin-bottom-0 govuk-!-margin-top-2">
     <summary class="govuk-details__summary">
     <span class="govuk-details__summary-text">
-      View radius secret key
+      View RADIUS secret key
     </span>
   </summary>
   <div class="govuk-details__text">
--- a/app/views/ips/_table.html.erb
+++ b/app/views/ips/_table.html.erb
@@ -7,11 +7,11 @@
       <%= location.radius_secret_key %>
       </div>
     </div>
-     <% if current_user.can_manage_locations? %>
+    <% if current_user.can_manage_locations? %>
       <p class="govuk-body govuk-!-margin-bottom-1 inline-block">
-        <%= link_to "Add IP to this location", new_location_ip_path(location), class: "govuk-link" %>
+      <%= link_to "Add IP to this location", new_location_ip_path(location), class: "govuk-link" %>
       </p>
-    <% end %>
+      <% end %>
   </caption>
 
   <tbody class="govuk-table__body" id="ips-table">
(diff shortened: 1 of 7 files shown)


tommotaylor:
@georgeschena We need to figure out how it will look with the "+ add IP" and this "View Radius secret key link". They are in the same place and thats causing the conflict.

Maybe start with merging in latest master and then have a play to see how they sit with eachother.

app/views/ips/_table.html.erb
```diff
@@ -1,6 +1,16 @@
 <table class="govuk-table">
   <caption class="govuk-table__caption">
     <%= location.full_address %>
+  <details class="govuk-details govuk-!-margin-bottom-0 govuk-!-margin-top-2">
```
kjdchapman: Feels like we're using these `!` classes a lot, worry it's going to make things more complicated.

Not sure if there's a way to avoid it, just wanted to share the worry out loud.

app/views/ips/_table.html.erb
```diff
@@ -1,6 +1,16 @@
...
+    <summary class="govuk-details__summary">
+    <span class="govuk-details__summary-text">
+      View RADIUS secret key
+    </span>
+  </summary>
+  <div class="govuk-details__text">
+  <%= location.radius_secret_key %>
```
kjdchapman: nit: indentation
```suggestion
    <%= location.radius_secret_key %>
```

kjdchapman: Slightly less of a nitpick: It looks like lines 4-13 all need to indent twice (four spaces) more, so that `details` is within ` caption`?

spec/features/ips/view_ip_addresses_spec.rb
```diff
@@ -37,6 +37,7 @@
       let(:address_two) { '123 Northern Street, Whitechapel' }
       let(:postcode_one) { 'SE4 4BK' }
       let(:postcode_two) { 'NE7 3UP' }
+      let!(:location) { create(:location, organisation: user.organisation) }
```
kjdchapman: You can do this without `let!`, which I think is better (feels like this spec needs some love in general, it's very long).

I pushed up a [possible solution on another branch](https://github.com/alphagov/govwifi-admin/blob/radius-secret-testing/spec/features/ips/view_ip_addresses_spec.rb).

app/views/ips/_table.html.erb
georgeschena: When messing around with padding and margins its hard not to use them to make the UI look more _professional_

spec/features/ips/view_ip_addresses_spec.rb
georgeschena: Makes more sense I like it

app/views/ips/_table.html.erb
kjdchapman: Yeah, it's no criticism of how this has ended up looking - think it does good work here - just want to make sure we're not face-kicking future-us.

tommotaylor:
Looks good to me

PR #346, opened 2018-12-03
Refactor IP creation controllers

tommotaylor:
Take 1: refactor using a service.

### DIFF
--- a/app/controllers/ips_controller.rb
+++ b/app/controllers/ips_controller.rb
@@ -6,29 +6,20 @@ class IpsController < ApplicationController
     @locations = available_locations
   end
 
-  # rubocop:disable Style/IfInsideElse
   def create
     @ip = Ip.new(create_params)
 
     if @ip.save
-      publish_for_performance_platform
-      publish_radius_whitelist
+      Services::HandleNewIp.new(ip: @ip).execute
       redirect_to(
         ips_path,
-        anchor: 'ips',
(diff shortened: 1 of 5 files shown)


app/controllers/ips_controller.rb
```diff
@@ -6,29 +6,20 @@ def new
...
   end
 
-  # rubocop:disable Style/IfInsideElse
   def create
     @ip = Ip.new(create_params)
 
     if @ip.save
-      publish_for_performance_platform
-      publish_radius_whitelist
+      Services::HandleNewIp.new(ip: @ip).execute
```
emileswarts: There is no way we can re-use this same controller and just check for the presence of params[:locations] ?
Would this complicate stuff? What is the Rails way of doing this?

kjdchapman:
I like parts of this approach - I think pulling the common, reasonably lengthy code out of the controllers and into its own component is good.

I don't like introducing the concept of a "Service", because I think a Use Case can do the same job whilst demanding less from a reader. I don't think I understand why this can't be a Use Case. 

It feels like the root problem we were trying to address here was the complex injection making the controller code less readable. We could have a Use Case Factory or similar to take care of this?

(If I understand right, whether this new class is a service or otherwise, it doesn't actually need to take any parameters?)

Happy to talk in-person if it's easier.

kjdchapman:
I'd still pick Use Case, but I like Facade better than Service. 

I still think it doesn't need to take ip as a parameter?

PR #347, opened 2018-12-03
Strip trailing and leading whitespace on username logs search

georgeschena:


spec/features/logging/view_auth_requests_for_a_username_spec.rb
```diff
@@ -50,6 +50,14 @@
       click_on "Submit"
     end
 
+    context "When username has a traling whitespace" do
+      let(:username) { "George " }
+
+      it "displays the username WITHOUT the trailing whitespace" do
+        expect(page).to have_content("Displaying logs for George")
```
tommotaylor: This will still pass if there is a trailing whitespace.

db/seeds.rb
```diff
@@ -45,6 +51,8 @@
   organisation_id: organisation.id
 )
 
+Ip.create(address: '31.242.140.103', location: location_2)
```
tommotaylor: With this new session and IP, I would do this more dynamically, so it runs no risk of being a duplicate, and makes a bit more sense to the next developer. You could give every IP in location_1 a session like this:

```
location_1.ips.each_with_index do |ip, index|
  Session.create(start: (Time.now - index.day).to_s,
    success: index.even? ? true : false,
    username: "garry",
    siteIP: ip.address
  )
end
```
That way one location has logs and another doesn't. Also all the logs will be different

tommotaylor: You will obviously have to do this after all the Ips are created for location_1

PR #348, opened 2018-12-04
WIP: Add many ips 

kjdchapman:
This has a lot of missing components, but shows which direction I'm going.

No javascript yet - I've just hard-coded a second IP field for now, to demonstrate how we might handle multiple IPs arriving at the controller.

Current status:
- saves two IPs if two are entered
- doesn't save one IP is one is blank
- feedback on both success and failure is broken
- no javascript for 3+ fields yet

### DIFF
--- a/app/controllers/locations/ips_controller.rb
+++ b/app/controllers/locations/ips_controller.rb
@@ -6,10 +6,11 @@ class Locations::IpsController < ApplicationController
   end
 
   def create
-    if ips.all?(&:valid?)
-      ips.each{|ip| ip.save}
+    ips = ips_from_params.each{|ip| ip.save}
 
+    if ips.all?(&:valid?)
       Facades::Ips::AfterCreate.new.execute
+
       redirect_to(
         ips_path,
         notice: "Ips added"
@@ -27,8 +28,9 @@ private
       .find_by(id: params[:location_id])
   end
 
-  def ips
-    params.dig(:ip, :address).map do |address|
+  def ips_from_params
+    addresses = params.dig(:ip, :address) || []
+    addresses.map do |address|
       Ip.new(address: address, location: @location)
     end
   end
--- a/app/controllers/locations/ips_controller.rb
+++ b/app/controllers/locations/ips_controller.rb
@@ -6,12 +6,13 @@ class Locations::IpsController < ApplicationController
   end
 
   def create
-    @ip = @location.ips.new(ip_params)
-    if @ip.save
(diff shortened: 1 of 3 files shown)


app/controllers/locations/ips_controller.rb
```diff
@@ -6,12 +6,13 @@ def new
   end
 
   def create
-    @ip = @location.ips.new(ip_params)
-    if @ip.save
+    if ips.all?(&:valid?)
```
kjdchapman: This doesn't feel like the "right" flow in a Rails context, but it's also not unclear (I hope).

spec/features/ips/add_ip_to_location_spec.rb
```diff
@@ -32,6 +32,22 @@
       end
     end
 
+    context 'with multiple IPs' do
+      before do
+        fill_in 'address', with: '10.0.0.1'
+        click_on 'Add another IP'
+        fill_in 'address2', with: '10.0.0.2'
```
kjdchapman: Had a bit of a play with capybara selectors, not sure how best to pick out the multiple select boxes when they come to be dynamically generated.

app/views/locations/ips/new.html.erb
```diff
@@ -10,13 +10,20 @@
 
     <%= form_with model: @ip, url: location_ips_path do |form| %>
       <div class="govuk-form-group <%= field_error(@ip, :address) %>">
-        <%= form.label :ip, "Enter IP address (IPv4 only)", class: "govuk-label" %>
-        <%= form.text_field :address, id: :address, autocomplete: "off", class: "govuk-input" %>
-      </div>
+        <%= form.label :address, "Enter IP address (IPv4 only)", class: "govuk-label" %>
+        <%= form.text_field :address, id: 'address', autocomplete: "off", class: "govuk-input", multiple: true %>
```
kjdchapman: `multiple: true` worked pretty nicely for the POST here - makes it so the addresses get picked up as an array in params, in the controller.

app/controllers/locations/ips_controller.rb
```diff
@@ -6,12 +6,13 @@ def new
...
+      ips.each{|ip| ip.save}
```
tommotaylor: You should be able to just save the params as an array of arrays, eg: the following works:
```
  Ip.create([[address: "1.1.1.1", location_id: 2], [address: "2.2.2.2", location_id: 2]])
```

Can you get it to pass in params like that?

PR #350, opened 2018-12-04
Rename Super Admin Spec file

Renny4Real:


spec/features/admin/sign_in_as_super_admin_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'features/support/sign_up_helpers'
 require 'features/support/not_signed_in'
 
-describe 'signing as an admin user' do
+describe 'signing as a super admin user' do
```
kjdchapman: Should it be this?
```suggestion
describe 'signing in as a super admin user' do
```

Renny4Real: 👍

PR #353, opened 2018-12-04
Add many IPs in nested form

kjdchapman:
This approach uses fields_for and updates the parent using attributes pretty much out of the box, which means things read pretty well, the code footprint is small. This solution will save successfully if you fill in all 5 boxes, but otherwise blows up.

It looks like getting the error messages in-line is going to be difficult, but possible. 'Wrong' IPs do stay in the inputs at least, rather than clearing everything out.

I also need to work out how this approach will work with the 'other' add IP page - I suspect a similar structure for newing up a location is going to make the existing stuff simpler

No testing yet, I'm basically spiking to discover how Rails wants this set up.

Suggestions/feedback very welcome!


tommotaylor:
I'm not sure I agree with the `LocationsController#edit` and `LocationsController#update` actions. Firstly we aren't editing a location here, we're creating Ips. Secondly we already have two controllers that have Ip creation, `IpsController#create` and `Locations::IpsController#create`.

Doesn't seem RESTful or Railsy with the `LocationsController#update` action.

kjdchapman:
Had a chat about the above with @tommotaylor & @emileswarts - agreed that we definitely don't want to end up with three controllers.

I'm happy to treat creating children as an update to their parent resource, but others didn't agree, so I'm similarly happy to use our existing structure. We'll pull this PR in that direction once we've got multiple IP creation working.

kjdchapman:
The footprint of this is pretty huge at the moment - I'm getting there with wrangling the moving parts, though. Trimming it down to the smallest part (splitting up the new/existing location add IP routes) so this can get merged soon.

Structure of controllers as-is still doesn't reflect our discussion from above, that is in the "wrangling TODO" pile.

kjdchapman:
By no means final, but helps illustrate:

image

app/models/location.rb
```diff
@@ -1,6 +1,9 @@
 class Location < ApplicationRecord
   belongs_to :organisation
+
   has_many :ips
+  validates_associated :ips
```
kjdchapman: Without this, it's possible to save locations with invalid IPs, though it seems to be inconsistent when.

With it, there's some kind of bug that seems to lead to an infinite validation loop sometimes - a bad enough bug that I don't want to merge this without fixing it.

kjdchapman:
So, two outstanding issues with this PR:
- Sometimes validation causes an infinite loop (this needs to be fixed before merging)
- I can add a location with no IPs at all (this seems less critical)

Some screenshots comparing OLD/NEW to follow shortly.

kjdchapman:
Had a chat with Laurence at GDS about the design. We talked a little about http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/list-entry.html and if there was a need for Javascript. 

He felt that starting with a similar style was good, but that adding in the JS later (as a "nice-to-have") was a fine approach.

To that end, I'm now happy with this PR - **once the major bug has been fixed**.

### Summary of this PR's changes
#### Separates out adding a new location
[Old version](https://user-images.githubusercontent.com/429326/49750662-6ccc5780-fca3-11e8-9a55-693e8c6cbd06.png)

[New version - step 1](https://user-images.githubusercontent.com/429326/49750818-c7fe4a00-fca3-11e8-829b-992b9993d9eb.png)

[New version - step 2](https://user-images.githubusercontent.com/429326/49750840-d2b8df00-fca3-11e8-96a9-b356f4d5fbcc.png)



#### Cleans up the controllers 
Now we have just `IpController` for adding IPs, and `LocationsController` for adding locations!

#### Introduces adding multiple IPs at Location creation
Allows blanks, and has proper error highlighting per-box!

image

app/controllers/ips_controller.rb
```diff
@@ -37,34 +38,21 @@ def destroy
...
     params.require(:ip).permit(:address, :location_id, location_attributes: %i[address postcode])
   end
 
   def create_params
-    return params_with_new_location if user_creates_new_location?
-    params_with_existing_location
+    ip_params.except(:location_attributes)
   end
 
   def user_creates_new_location?
```
Renny4Real: This method doesn't seem to be used elsewhere. It could be removed.

spec/models/ip_spec.rb
```diff
@@ -14,14 +14,14 @@
         it "does not save when address is an invalid IP" do
           expect(Ip.count).to eq(0)
           expect(ip.errors.full_messages).to eq([
-            "Address must be a valid IPv4 address (without subnet)"
+            "Address 'invalidIP' is not valid"
```
Renny4Real: What do the single quotes do in this context. Is this another form of string interpolation?

app/models/location.rb
kjdchapman: Had a chat with Tom and we're starting to track this down.

Looks like it's a combination of the custom validation on Ip (`validate :address_must_be_valid_ip`) and the validation declaration on its parent (`validates_associated :ips`). Digging into it.

We also discovered that the problem only occurs when at least two IPs - one invalid and one valid - are saved side-by-side. 

If `validates_associated` is removed, an invalid IP can be saved following the same pattern - saving a location with one valid IP and one invalid IP.

app/controllers/ips_controller.rb
kjdchapman: Good spot! I'll get rid of it.

PR #356, opened 2018-12-04
Allow non-gov.uk email addresses to be invited

tommotaylor:
A requirement as often GovWifi admins will be external suppliers eg: bt.com

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -14,15 +14,25 @@ private
 
   def validate_invited_user
     return if invited_user_already_exists?
-    render_invite_user_page if user_is_invalid?
+    return_user_to_invite_page unless email_is_valid?
   end
 
-  def user_is_invalid?
+  def email_is_valid?
+    return true if email_passes_validation?
+    set_user_object_with_errors
+    false
+  end
+
+  def email_passes_validation?
+    invite_params[:email].match(URI::MailTo::EMAIL_REGEXP).present?
+  end
+
+  def set_user_object_with_errors
     @user = User.new(invite_params)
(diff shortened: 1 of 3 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -14,15 +14,25 @@ def add_organisation_to_params
...
+  end
+
+  def email_passes_validation?
+    invite_params[:email].match(URI::MailTo::EMAIL_REGEXP).present?
+  end
+
+  def set_user_object_with_errors
     @user = User.new(invite_params)
-    !@user.validate
+    @user.errors.add(:email, " must be a valid email address")
```
emileswarts: Wonder if we can push this down into a model validation.
Validation can have a conditional to check whether the user is being created as an admin or invited.

tommotaylor: `!@user.validate` was using the model validation on the user. It checked that the email was present, valid email and that it was on the whitelist.

However in this instance we want to skip the whitelist check, and you can't do conditional validation like `!@user.validate(except: :whitelist)`. So in this instance I'm just checking that the email is an email manually (which covers presence and valid format).

I thought about separating validations from the model using something like [`dry-validation` gem](https://github.com/dry-rb/dry-validation) , but thought its overkill for one email field.

tommotaylor: Or were you thinking an instance method on `User` like `@user.email_valid`?

emileswarts: Was thinking of modifying the user model like this:

```
  validate :email_on_whitelist, on: :create, unless: :invited?
```

PR #360, opened 2018-12-07
Stop unconfirmed user invite causing a page error upon sign in

Renny4Real:
Problem: The `Users::InvitationsController#create` action was deleting permissions when an unconfirmed user was being invited again. This meant that users hit crazy errors if they were invited twice (or invited whilst being an unconfirmed user on a different org), as they didn't have permissions.

Solution: Stop existing users from being invited again.

Regression: This caused the resend invite process to stop working since it was using the `Users::InvitationsController#create` action to resend the invite to an existing user.

Solution: When resending an invite, delete the original invitation and then create a new one. This means the `Users::InvitationsController#create` only ever is called on new invites.


Renny4Real:
```
     ActionView::Template::Error:
       User#can_manage_locations? delegated to permission.can_manage_locations?, but permission is nil: #<User id: 12, email: "notconfirmedyet@gov.uk", created_at: "2018-12-07 11:12:44", updated_at: "2018-12-07 11:12:44", organisation_id: 10, name: "claire", mobile: nil, super_admin: false>
```
This is a new error that I'm getting but it does give me confidence that the confirmation is working on the right email.

kjdchapman:
I haven't had the chance to dig properly into this yet, but from my read-through it looks good! A few hanging comments/tests that might not need to get merged in, can go into more detail tomorrow.

Said it already in person, but really like this way of approaching the problem - testing around the broken case first is :+1:

PR #361, opened 2018-12-08
Super admins can remove an organisation 

georgeschena:


spec/features/admin/delete_organisation_as_super_admin_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'features/support/sign_up_helpers'
+
+describe 'deleting an organisation' do
+  let!(:admin_user) { create(:user, super_admin: true) }
```
kjdchapman: nit: Don't think you need the `!` here

```suggestion
  let(:admin_user) { create(:user, super_admin: true) }
```

spec/features/admin/delete_organisation_as_super_admin_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  let!(:organisation) { create(:organisation, name: "TestMe & Company") }
```
kjdchapman: same as above

```suggestion
  let(:organisation) { create(:organisation, name: "TestMe & Company") }
```

spec/models/organisation_spec.rb
```diff
@@ -11,5 +11,19 @@
         Organisation.create(name: "parks & rec dept")
       }.to change { Organisation.count }.by(0)
     end
+
+    context 'when deleting an organisation as an admin' do
+      let!(:org) { create(:organisation) }
+      let!(:user1) { create(:user, organisation: org) }
```
kjdchapman: I'd just call these `user`, `location` etc without the `1`.

PR #363, opened 2018-12-11
Admins can edit organisation details

georgeschena:


app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+  <h1 class="govuk-heading-l"> Organisation Settings</h1>
```
tommotaylor: white space before " Organistion Settings"

spec/features/admin/edit_organisation_details_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  context 'when visiting the edit organisations settings page' do
+    before do
+      sign_in_user admin_user
+      visit edit_organisation_path(organisation)
+    end
+
+    it 'allows the user to change their company name and service email' do
+      fill_in 'org-name', with: "New Company Name"
```
tommotaylor: You should just be able to use the label, eg: `fill_in "Organisation name"` rather than having to create an arbitrary ID.

spec/features/admin/edit_organisation_details_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'features/support/sign_up_helpers'
+
+describe 'editing an organisations details' do
+  let!(:admin_user) { create(:user, super_admin: true, organisation: organisation) }
```
tommotaylor: I don't think this user needs to be a super admin to see this.

app/controllers/organisations_controller.rb
```diff
@@ -0,0 +1,23 @@
+class OrganisationsController < ApplicationController
+  def index; end
+
+  def edit
+    @organisation = Organisation.find_by(id: params.fetch(:id))
+  end
+
+  def update
+    organisation = Organisation.find_by(id: params.fetch(:id))
+    if organisation.update_attributes(organisartion_params)
```
tommotaylor: better to user `.update` than `.update_attributes`. See the difference here: https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html

app/controllers/organisations_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      flash[:notice] = 'Organisation updated'
+      redirect_to organisations_path
+    else
+      flash[:alert] = 'Update failed'
```
tommotaylor: Doesn't handle the failure properly, when I save the form with an empty service email it gives me a green success banner and takes me back to the show page. It should re-render the edit page and show the errors in the form.

spec/features/admin/edit_organisation_details_spec.rb
georgeschena: This isn't working. when I run tests it can't find the element, why I changed it to an ID @tommotaylor

app/controllers/organisations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      flash[:notice] = 'Organisation updated'
+    else
+      render :edit
+    end
+  end
+
+private
+
+  def set_organisation
+    @organisation = Organisation.find(params.fetch(:id))
```
emileswarts: We do have a helper method called
```current_organisation```

So this could just be 
```
@organisation = current_organisation
```

Comment (georgeschena) [reply app/controllers/organisations_controller.rb]: I think its done in this this way to ensure not anyone can just hardcode the URL and change another organisations settings. @emileswarts

Comment (georgeschena) [reply app/controllers/organisations_controller.rb]: From my understanding now:

let's say I'm organisation ID 1
if I go to `organisations/1` then that's the same as `current_organisation`
but if I go to `organisations/2` then that's `Organisation.find(2)`, which isn't my `current_organisation`
so the two aren't equivalent

File: app/views/organisations/edit.html.erb [hunk 7]
```diff
@@ -0,0 +1,19 @@
+<%= render "layouts/form_errors", resource: @organisation %>
```
kjdchapman: I'd recommend the use of https://design-system.service.gov.uk/components/error-message/ to help out users more here. It seems like a minor improvement, but I think it's better to address this first time round rather than launch something that deviates from GDS standards.

Tom mentioned this, but it's kind of got lost in the thread.

kjdchapman: Specifically, the 
> Show an error message next to the field 
advice.

You can see how other parts of the app use the mentioned classes (like `govuk-input--error`) if it's not clear how to achieve this.

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <h1 class="govuk-heading-l">Edit Organisation Settings</h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-form-group govuk-body-m" >
+
+        <%= form_for @organisation do |form| %>
+          <%= form.label :name, "Organisation name", class: "govuk-label" %>
+          <%= form.text_field :name, class: "govuk-input" %><br />
+
+          <%= form.label :service_email, "Service email", class: "govuk-label govuk-!-padding-top-4" %>
```
kjdchapman: There's a bunch of classes to organise forms that work better than these `govuk-!` classes on this form, as this form doesn't do anything particularly special.

You can see them use in the examples in the design system - like https://design-system.service.gov.uk/patterns/addresses - or elsewhere in the app. 

Wrapping label/fields in `div`s with the `govuk-form-group`, is roughly what you want.

spec/features/admin/edit_organisation_details_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'features/support/sign_up_helpers'
+
+describe 'editing an organisations details' do
+  let!(:admin_user1) { create(:user, organisation: organisation) }
+  let!(:admin_user2) { create(:user) }
```
kjdchapman: This second admin user isn't relevant to most of the specs. I'd sink it into the spec on l35 (`"...tries to change another organisation..."`).

kjdchapman: Neither of these need to be `let!`, unless I'm misunderstanding something.

spec/features/admin/edit_organisation_details_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    before do
+      sign_in_user admin_user1
+      visit edit_organisation_path(organisation)
+    end
+
+    it 'allows the user to change their company name and service email' do
+      fill_in 'Organisation name', with: "New Company Name"
+      fill_in 'Service email', with: "NewServiceEmail@gov.uk"
+      click_on 'Save'
+      expect(page).to have_content("New Company Name")
```
kjdchapman: This test might be too permissive, but I'm not sure (depends on how `have_content` behaves)

Consider the scenario where someone accidentally breaks the validation on this model, and the save always fails. When I hit 'Save' on this page, the save will fail, and I'll be on the `new` view still, but with what I typed in still present. 

If `have_content` checks those input elements, this'll pass, even though the functionality is broken.

If `have_content` doesn't look in input elements, then this isn't a problem, and this test seems fine.

georgeschena: Bit confused by this. Speak to you face to face about it maybe

kjdchapman:
Mind including another screenshot of what it looks like after the styling changes? I'm pretty ready to tick this!

kjdchapman:
The alignment on the view page looks off to me:

image

Suspect this is because [the show page is using the form group classes](https://github.com/alphagov/govwifi-admin/pull/363/files#diff-5b0e3633dd4daec586b3b44602c456a2R4) - doesn't need these as it's not a form.

If that doesn't fix it, I'd happily use the `govuk-!` classes in this case to get the alignment right.

kjdchapman:
Taking a closer look at the markup on the edit view, it doesn't look right to me either. Putting together a commit that demonstrates.

kjdchapman:
Alignment and slight redesign on the `show` page:

image

Less markup, same effect on the `edit` page:

image

app/views/organisations/edit.html.erb
```diff
@@ -0,0 +1,39 @@
...
+
+<%= link_to "Back to settings", organisation_path, class: "govuk-back-link" %>
+<h1 class="govuk-heading-l">Edit Organisation Settings</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @organisation do |form| %>
+      <div class="govuk-form-group" >
+        <%= form.label :name, "Organisation name", class: "govuk-label" %>
+        <% if @organisation.errors.keys.include?(:name) %>
+          <%= form.text_field :name, class: "govuk-input govuk-input--error" %>
```
emileswarts: Small one here:
Could compress this into just:

```
<% extra_class = @organisation.errors.keys.include?(:name) ? "govuk-input--error" : "" %>
<%= form.text_field :name, class: "govuk-input #{extra_class}" %>
```

PR #364, opened 2018-12-11
Move email whitelist validation to controller

tommotaylor:
The validation that checks that a user's email is on the whitelist is only used during registration. During invitation the user can have any email address.

Having this validation on the model meant there was a lot of code in the invitations controller trying to avoid this check. 

As it only applies to 20% of the users, I have moved the whitelist validation off the model to the only controller that uses it. This simplifies the code changes coming in https://github.com/alphagov/govwifi-admin/pull/360


kjdchapman:
LGTM if it's making something else simpler!

PR #367, opened 2018-12-12
Block unconfirmed users from resetting passwords

Renny4Real:
**WHAT:** The aim was to stop a user who hadn't confirmed their account from trying to reset their password.

**WHY:** Users who went down this route encountered an error stating that `Name can't be blank`. The name referred to here is the name they would have set through _confirming and finishing the creation of their account._

screenshot_2018-12-04_at_14 44 21

**IN THIS PR:** An initial solution was to send a confirmation email to the unconfirmed user instead of the reset password email, whenever they clicked on the reset password link in the admin portal. 

However, in terms of user experience, the user still sees a flash notice along the lines of `An email with instructions on how to reset your password has been sent`. So, the current idea of a solution is to deal with this at the controller level so that the appropriate flash message from the `devise.en.yml` is seen by the user.

**UPDATE ON SOLUTION**
A method was added to the `PasswordsContoller` to check if a user has been confirmed before proceeding with reset password pathway. 

As a result of this, the user is now redirected to the `new_user_confirmation_path` and are informed that they need to confirm their account before attempting to reset their password. 
- This now also means that the user no longer receives an email with the confirmation link automatically but has to initiate the confirmation journey themselves.

screenshot 2018-12-14 at 13 29 00

Any feedback on this would be appreciated.

### DIFF
--- a/app/controllers/users/passwords_controller.rb
+++ b/app/controllers/users/passwords_controller.rb
@@ -1,12 +1,15 @@
 class Users::PasswordsController < Devise::PasswordsController
-  before_action :validate_user_is_confirmed, only: :edit
+  append_before_action :validate_user_is_confirmed, only: :create
+  append_before_action :assert_reset_token_passed, only: :edit
 
 protected
 
   def validate_user_is_confirmed
-    unless @user.confirmed?
-      redirect_to new_user_session_path
-      flash[:notice] = "Please confirm your account first"
+    current_user_details = params[:user]
+    user_email = current_user_details[:email]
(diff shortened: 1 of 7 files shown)


app/controllers/users/passwords_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Users::PasswordsController < Devise::PasswordsController
+  append_before_action :validate_user_is_confirmed, only: :create
+  append_before_action :assert_reset_token_passed, only: :edit
```
kjdchapman: Can this line go?
```suggestion

```

PR #369, opened 2018-12-13
Allow a user to download the GovWifi poster

georgeschena:


Renny4Real:
Wondering if this should have a test around it?

kjdchapman:
From a quick google, it looks like it's quite hard to test whether the download has happened - but maybe we should just test that the content ("Download a poster") is present on this page?

Renny4Real:
There's a spec around uploading and downloading the MOU. 
Might get some inspiration from there as well?

app/views/home/index.html.erb
```diff
@@ -44,7 +44,9 @@
     <div>
       <h3 class="govuk-heading-m"> 5. Check the connection works </h3>
       <p class="govuk-body"> Connect to your new GovWifi network, and sign in with an individual account. </p>
-      <details class="govuk-details">
+      <%= link_to "Download a poster to advertise GovWifi is available in your building(s)", asset_path('govwifi-poster.png'), { download: "" ,class: "govuk-link govuk-body-m"}  %>
```
kjdchapman: I think this doesn't need the govuk-body-m class?
```suggestion
       <%= link_to "Download a poster to advertise GovWifi is available in your building(s)", asset_path('govwifi-poster.png'), { download: "" ,class: "govuk-link"}  %>
```

PR #370, opened 2018-12-13
Add location with multiple IPs

kjdchapman:
This PR includes specs that isolate the seemingly-broken behaviour around adding a Location and IP at the same time.

With `validates_associated` included, we get a timeout, with it not included, you can save invalid IPs.

Opening a Rails issue, link to follow.


kjdchapman:
Here's the Rails issue: https://github.com/rails/rails/issues/34696

app/models/location.rb
```diff
@@ -1,6 +1,12 @@
 class Location < ApplicationRecord
   belongs_to :organisation
+
   has_many :ips, dependent: :destroy
+  accepts_nested_attributes_for :ips
+
+  # When present, this breaks ip_spec.rb:54
+  # When absent, this breaks location_spec.rb:42
+  # validates_associated :ips
```
emileswarts: This looks like what we want no?

kjdchapman:
This should now be working! Good spot @emileswarts on us having two `accepts_nested_attributes_for` pointing to each other. That was what was breaking it!

The Rails issue commenter pointed this out too, but I didn't parse it correctly. I'll go explain there what happened as well.

app/controllers/ips_controller.rb
```diff
@@ -2,12 +2,13 @@ class IpsController < ApplicationController
   before_action :authorise_manage_locations, only: %i(create new)
 
   def new
-    @ip = Ip.new
+    selected_location = Location.find_by(id: params[:location])
```
emileswarts: Would it be better to use:
```
current_organisation.locations.find(params.fetch(:location))
```

To ensure that the found location belongs to the organisation of the current user?

app/controllers/locations_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def create
+    @location = Location.new(location_params_without_blank_ips)
+
+    if @location.save
+      Facades::Ips::AfterCreate.new.execute
+      redirect_to ips_path, notice: "#{@location.full_address} added"
+    else
+      add_blank_ips!
```
emileswarts: Is this meant to be a bang method?
In Rails they are only used when they potentially raise an exception.

app/models/location.rb
```diff
@@ -1,6 +1,10 @@
 class Location < ApplicationRecord
   belongs_to :organisation
+
   has_many :ips, dependent: :destroy
+  accepts_nested_attributes_for :ips
+
+  validates_associated :ips
```
emileswarts: Nice nice

app/views/ips/index.html.erb
```diff
@@ -11,7 +11,7 @@
   </div>
 </div>
 
-<% if current_organisation.ips.count == 0 %>
+<% if current_organisation.locations.count == 0 %>
```
emileswarts: Perhaps 
```
current_organisation.locations.empty?
```

spec/models/ip_spec.rb
```diff
@@ -50,4 +50,32 @@
...
+      )
+    end
+
+    let(:ip) do
+      Ip.new(location_id: location_1.id, address: "127.0.2.1")
+    end
+
+    before { 20.times { create(:ip, location: location_1) } }
+
+    it 'saves without timing out' do
```
emileswarts: Do we still need this given we have fixed the root cause of the issue?

spec/models/ip_spec.rb
```diff
@@ -50,4 +50,32 @@
...
+        postcode: 'SE10SX',
+        organisation_id: create(:organisation).id
+      )
+    end
+
+    let(:ip) do
+      Ip.new(location_id: location_1.id, address: "127.0.2.1")
+    end
+
+    before { 20.times { create(:ip, location: location_1) } }
```
emileswarts: Struggling what this 20.times sets the spec up for.
Is it related to the timeout spec below?

app/controllers/locations_controller.rb
kjdchapman: This isn't my understanding of what `!` means on method names in Ruby, rather it's _"watch out, this could be dangerous!"_. 

In a ActiveRecord context it's often used for saying "this'll throw exceptions, watch out!". It often also points out it'll mutate something in-place (e.g. `sort` v `sort!`).

I've missed a key bit of idiomatic Ruby though - it's very rare to have `some_method!` defined without a safe version being available (`some_method`). I think I won't `!` here, even though modifying the params collection is a _little_ weird.

spec/models/ip_spec.rb
kjdchapman: Yeah, I wasn't sure either. Let's chop it out, less code is good?

spec/models/ip_spec.rb
kjdchapman: If we're gonna chop it, I can dodge this question :sweat_smile:

app/controllers/ips_controller.rb
kjdchapman: If the found location isn't in the organisation, the location selector will just default to the first entry. Not too bad. Would have to hack the URL to cause that behaviour, too.

Good spot though - I think people can use this controller to add IPs to locations that aren't in their organisation.

I think this is pre-existing though - I'd like to address it in a different PR so I can get this one merged, if that's acceptable to you?

app/controllers/locations_controller.rb
emileswarts: Okay yeah fair enough, let's come back to this.
Needs an automated test wrapped around it.

emileswarts: Woops this comment was meant for the location finding functionality.

PR #375, opened 2018-12-18
Autoload support files in spec_helper

kjdchapman:
A quick one before I dive into spies/mocks. Load time isn't significantly different in any situation, but reduces noise on each spec (tradeoff: some things are now more hidden).

For the whole suite:

image

For a single focused test that doesn't use these files:

image

For a single focused test that does use these files: 

image


emileswarts:
nice!

PR #379, opened 2018-12-19
Resize the input boxes on admin

Renny4Real:
**WHY:**
This was done based on a suggestion to edit the design of text boxes to suit the length of the input required in them.

**IN THIS PR:** 
Edit the input classnames for the appropriate text fields.

**SOLUTION EXAMPLE:**

**BEFORE:**

screenshot 2018-12-19 at 13 20 47

**AFTER:**
screenshot 2018-12-19 at 13 24 48

### DIFF
--- a/app/views/help/index.html.erb
+++ b/app/views/help/index.html.erb
@@ -47,7 +47,7 @@
         <label class="govuk-label" for="contact_number">
           Contact number (optional)
         </label>
-        <%= text_field_tag 'contact_number', nil, class: 'govuk-input', type: 'tel' %>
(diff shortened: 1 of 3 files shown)


kjdchapman:
Surprised that there's so few boxes that needed changing - is our app just that small? Or had we done this in some places already?

Renny4Real:
Some were already done in this PR #370 and some input fields seem to depend on the user  e.g address, password, etc. so I couldn't really estimate these. 
Should we agree on a standard throughout the app?

PR #380, opened 2018-12-19
Create generic notifications mock

kjdchapman:
This is the first step of refactoring the mocks/spies around notify. There's a 
few things I'm not 100% happy with, but I think it's worth doing a bit at a time.

If this is an acceptable approach, I should be able to extend this to all of the 
other feature files as well.


kjdchapman:
Some of the notify use cases implicitly relied on some of the things I changed. I've fixed them to not do that, let's see if that gets the suite passing.

PR #381, opened 2018-12-20
Extend notifications mock to include links

kjdchapman:
Doing these tests a few at the time to help inform the design of the notifications mock. This PR adds recording the URLs that are often found in the emails we sent out, should make it easier for these tests to.

Replacing five requires with a single one is pretty sweet.

### DIFF
--- a/spec/features/support/mock_notification_client.rb
+++ b/spec/features/support/mock_notification_client.rb
@@ -10,7 +10,19 @@ shared_context 'with a mocked notifications client' do
 
     # TODO: remove dependency on test entries in gov-notify.yml
     def send_email(args)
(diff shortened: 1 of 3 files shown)


spec/features/support/mock_notification_client.rb
```diff
@@ -10,7 +10,19 @@ def initialize(_unused)
 
     # TODO: remove dependency on test entries in gov-notify.yml
     def send_email(args)
-      self.class.notifications << args[:template_id]
+      self.class.notifications << {
```
Renny4Real: Just to check my understanding:  
`<<` appends to the end of an array right? 
So `self.class.notifications` is an array of hashes in this method?

PR #388, opened 2019-01-03
Add dashboard display to overview page

Renny4Real:
**WHY:**
It was suggested that if an organisation has setup and are running GovWifi then they don't need to see the get started checklist as the first page when they login. They should instead see something that gives them an overview of their current installation.

**IN THIS PR:**
The changes to get to the solution would be added in increments and not implemented as a one large feature. This may allow for user feedback changes/additions to be easily implemented. 
The `index.html` file for the `HomeController` was modified to include a separator. 
Above the separator, html elements were added to create the overview dashboard look.
Below the separator, the setup instructions remain unchanged in this PR.

**BEFORE:**
screenshot 2019-01-03 at 11 35 50

**AFTER:**
screenshot 2019-01-03 at 11 36 38

**NEXT STEPS:**
The next iteration would involve reorganisation. The setup instructions should have its own page and more stuff (as necessary) could be added to the overview page e.g. number of roaming endusers & guest users in the organisation.
The idea behind this is to change where the admin user lands after login based on whether they have setup GovWifi & running it in their organisation or not.

### DIFF
--- a/app/assets/stylesheets/utilities/layout.scss
+++ b/app/assets/stylesheets/utilities/layout.scss
@@ -13,3 +13,11 @@
 .inline-block {
   display: inline-block;
 }
+
+.dashboard-colour {
+  background:#2b8cc4;
+}
+
+.dashboard-font-size {
+  font-size: 35px;
+}
--- a/app/assets/stylesheets/utilities/layout.scss
+++ b/app/assets/stylesheets/utilities/layout.scss
@@ -15,9 +15,5 @@
 }
 
 .dashboard-colour {
-  background:#2b8cc4;
-}
-
-.dashboard-font-size {
-  font-size: 35px;
+  background: govuk-colour("light-blue");
 }
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -2,6 +2,8 @@ class HomeController < ApplicationController
   def index
     redirect_to admin_organisations_path if current_user.super_admin?
     @ips = current_organisation.ips
+    @locations = current_organisation.locations
+    @team_members = current_organisation.users
     @london_radius_ips = radius_ips[:london]
     @dublin_radius_ips = radius_ips[:dublin]
   end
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/utilities/layout.scss
```diff
@@ -13,3 +13,11 @@
 .inline-block {
   display: inline-block;
 }
+
+.dashboard-colour {
+  background:#2b8cc4;
```
kjdchapman: It'd be good to use the standard `govuk-colour` here, I assume they've got a fitting blue?

[Here's an example](https://github.com/alphagov/govwifi-admin/blob/master/app/assets/stylesheets/components/label_inside_input.scss#L3) from another stylesheet in the application, and the [list of colours can be found on the design system site](https://design-system.service.gov.uk/styles/colour/).

app/assets/stylesheets/utilities/layout.scss
```diff
@@ -13,3 +13,11 @@
...
+}
+
+.dashboard-font-size {
+  font-size: 35px;
```
kjdchapman: Similarly to colour, it'd be good to use the govuk classes here. There's [override classes](https://design-system.service.gov.uk/styles/typography/#font-override-classes) if you can't find the right size from the standard list.

kjdchapman:
Nice, think this is a good first step as you've described.

Thinking out loud, I wonder if it's OK to release this without splitting it up. It's not much to scroll past to get to the setup instructions, so I think it's alright, but be interested to hear opinions.

The three boxes look a little off in the screenshot - like they're 'hanging over' on the left. Suspect that's something to do with them not having the right grid classes?

I left some small comments on the styling as well, I'd be happy to approve it with that (and the alignment issue) sorted.

Renny4Real:
**UPDATE:**
Previously, the two info blocks on the edge appeared as though a bit of them was hanging just over the imaginary boundary that contained the rest of the page. The separator was also slightly longer than this boundary.

**SOLUTION:**
Make all the elements fit within the defined grid layout for the page by modifying the html elements involved.

screenshot 2019-01-03 at 13 03 50

PR #389, opened 2019-01-03
Improve searching logs UX

kjdchapman:
# Overview

There's more to do on error messaging but I'd like to do that in a separate PR, as this is getting chonky.

This PR:
- Introduces a "LogsSearch" resource, model and controller, which is responsible for search processing, validity
- Pulls apart the flow into one-question-per-page with clear navigation between
- Improves (and shortens!) a lot of the messaging around this feature
- Improves the language around the 'zero results' case, adding guidance on where to go next if it's surprising
- Tweaks the acceptance specs around this feature (pleasingly, wasn't much work! Nice work team on making these!)

# Navigating to 'Logs':
### OLD
image

### NEW:
(first screen)
image

(second screen)
image

# Seeing results
### OLD
image

### NEW
image

# Empty results
### OLD
image

### NEW
image

Thanks to Ashley for help with the design on this!


Renny4Real:
Looks really good. Didn't realise until now that there were multiple ways to search the logs and that these were in different parts of the platform.

app/controllers/logs_searches_controller.rb
```diff
@@ -0,0 +1,25 @@
+class LogsSearchesController < ApplicationController
+  def create
+    @search = LogsSearch.new(search_params)
+
+    if !@search.first_step && @search.valid?
```
Renny4Real: Don't really understand what this conditional does. 
What is the `first_step` method doing on `@search` and why do we want the `not` value of this and it's validity to be true? 
What does the `by` method do on `@search`?

kjdchapman: Searching logs is split into two steps, picking what you want to filter by, and entering a search term. This `create` method covers both steps, so there's some slight awkwardness to cover it.  

The `first_step` method is just a flag on the `LogsSearch` model. It's set to `true` if we're on that first, "choose what you want to filter by" step of searching. I would welcome naming suggestions!

The `#by` method is what we're searching by - IP or username. I think I should name it `#filter` instead.

This conditional is a ruby idiom (I believe), sometimes known as "[short circuit evaluation](https://blog.revathskumar.com/2013/05/short-circuit-evaluation-in-ruby.html)". The `@search.valid` part won't be evaluated if `!@search.first_step` is false, so we only check validity on the second step. 

This is to stop error messages showing on the first step (as there's never a `term` filled in on the model at that point.)

I wasn't super happy with the naming/structure here, I'll happily take suggestions! Going to look at it again as part of searching by location, I imagine.

spec/features/logging/view_auth_requests_for_a_username_spec.rb
```diff
@@ -10,31 +10,31 @@
       Session.create!(
         start: 3.days.ago,
         username: username,
-        mac: '',
-        ap: '',
-        siteIP: '1.1.1.1',
+        mac: "",
```
emileswarts: We really need to establish a convention around single vs double quotes.

kjdchapman: Agreed. I almost didn't include this in the pull request, it annoyed me both ways!

I don't care apart from consistency. Do you have a preference? iirc @tommotaylor slightly preferred double-quotes, but don't quote me on that.

Flip a coin then add a custom rubocop?

PR #391, opened 2019-01-04
Useful feedback to user if no bullet selected

kjdchapman:
This fixes the immediate issue of selecting no filter causing unhelpful errors for a user.

The error message is not ideal, but in the short-term it's better than this page crashing:

image

I plan to think about the structure of the validation/error messages as I work on searching by location.

### DIFF
--- a/app/controllers/logs_searches_controller.rb
+++ b/app/controllers/logs_searches_controller.rb
@@ -2,24 +2,37 @@ class LogsSearchesController < ApplicationController
   def create
     @search = LogsSearch.new(search_params)
 
-    if !@search.first_step && @search.valid?
-      redirect_to logs_path(@search.by.to_sym => @search.term)
+    @search.first_step ? filter_choice : term_choice
+  end
+
+  def filter_choice
+    if @search.filter.present?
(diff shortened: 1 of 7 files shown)


emileswarts:
Would it be simpler to select one radio button by default so the user can never get into this error state?

kjdchapman:
Considered it @emileswarts , but [design system on radio buttons explicitly recommends against](https://design-system.service.gov.uk/components/radios/):
> Do not pre-select radio options as this makes it more likely that users will:
> - not realise they’ve missed a question
> - submit the wrong answer

app/controllers/logs_searches_controller.rb
```diff
@@ -2,24 +2,37 @@ class LogsSearchesController < ApplicationController
   def create
     @search = LogsSearch.new(search_params)
 
-    if !@search.first_step && @search.valid?
-      redirect_to logs_path(@search.by.to_sym => @search.term)
+    @search.first_step ? filter_choice : term_choice
```
Renny4Real: Nice! 👍

PR #392, opened 2019-01-04
Separate overview and setup instructions pages

Renny4Real:
**WHY:**
Next iteration from PR #388 
An organisation that has setup GovWifi and is running it should see something that gives them an overview of their current installation. They do not need to see the setup instructions on the same page. Similarly, an organisation that has NOT set up GovWifi does not need to see an overview of their current installation but should definitely see how to setup.

**IN THIS PR:**
- `SetupController` and its`index.html` were added. The instructions for setting up were moved to this `index.html`. Organisations that have not yet setup GovWifi (determined by number of IPs for now) will be redirected to this page from `root`. Organisations that have setup GovWifi will not be redirected from the root page i.e. Overview.

- Tests were added and/or edited & made to pass before the transfer of contents from Overview to Setting Up as a precaution. **Suggestions and feedback on this aspect would be appreciated.**

**BEFORE:**

screenshot 2019-01-03 at 11 36 38

**AFTER:**

screenshot 2019-01-04 at 12 34 39

- Unless an Organisation has setup GovWifi, they will not see `Overview` as a link in the navigation.

screenshot 2019-01-04 at 12 35 29

- All users (regardless of setup status) will see `Setting Up` as a link in the navigation.
screenshot 2019-01-04 at 12 32 18

- `Active IPs` was renamed `Added IPs` to avoid confusion between Number of IPs with and without sessions running on them.  `Active IPs` may be implemented as a later feature. **Suggestions and feedback on this change would be appreciated.**

- `Added IPs` and `Admins` link to the IPs page and Team Members Page respectively.


Renny4Real:
Based on commit fa1c5ea.
Did not make `Locations` a link in the overview as there was no unique page to link to this. The closest was the IPs page which is already being linked to `Settings` and `Added IPs`. 
**Any suggestions & feedback on this would be appreciated.**

app/assets/stylesheets/utilities/layout.scss
```diff
@@ -14,6 +14,10 @@
   display: inline-block;
 }
 
-.dashboard-colour {
+.dashboard-background {
```
kjdchapman: I would pull these into a separate file in `app/assets/stylesheets/components`, maybe `dashboard.scss`. That then needs an extra import statement in [`application.scss`](https://github.com/alphagov/govwifi-admin/blob/master/app/assets/stylesheets/application.scss) and [`application_ie8.scss`](https://github.com/alphagov/govwifi-admin/blob/master/app/assets/stylesheets/application_ie8.scss).

app/controllers/home_controller.rb
```diff
@@ -1,6 +1,12 @@
 class HomeController < ApplicationController
   def index
-    redirect_to admin_organisations_path if current_user.super_admin?
+    if current_user.super_admin?
```
kjdchapman: This doesn't look idomatic to me, but I'm not sure what the right "Rails way" to do it is.

(From talking IRL, there were some linting problems around doing `redirect and return`, but worth having a quick read around early returns from a controller method).

app/controllers/setup_controller.rb
```diff
@@ -0,0 +1,16 @@
+class SetupController < ApplicationController
```
kjdchapman: I would call this `SetupInstructionController` or similar - worried that a reader might think this is _us_ setting up things at first glance.

app/views/application/_subnavigation.html.erb
```diff
@@ -13,7 +13,10 @@
       <%= link_to "Organisations", admin_organisations_path, class: active_tab(admin_organisations_path) %>
       <%= link_to "MOU", admin_mou_index_path, class: active_tab(admin_mou_index_path) %>
     <% else %>
-      <%= link_to "Overview", root_path, class: active_tab(root_path) %>
+      <% if current_organisation.ips.count != 0 %>
```
kjdchapman:
```suggestion
      <% if current_organisation.ips.present? %>
```

app/views/home/index.html.erb
```diff
@@ -2,17 +2,17 @@
...
   <div class="govuk-grid-column-full">
     <div class="govuk-grid-column-one-third">
-      <div class="govuk-panel govuk-panel--confirmation dashboard-colour">
+      <div class="govuk-panel govuk-panel--confirmation dashboard-background">
         <h1 class="govuk-panel__title text-left">
           <%= @team_members.count %>
         </h1>
         <p class="govuk-panel__body text-left govuk-!-font-size-27">
-          Team Members
+          <%= link_to "Admins", team_members_path, class: "govuk-link govuk-link--no-visited-state dashboard-link-colour" %>
```
kjdchapman: I would keep this as Team Members to match the nav, can use Capybara's `within` method if tests were tricky with the duplication.

app/views/home/index.html.erb
```diff
@@ -22,88 +22,15 @@
...
     </div>
     <div class="govuk-grid-column-one-third">
-      <div class="govuk-panel govuk-panel--confirmation dashboard-colour">
+      <div class="govuk-panel govuk-panel--confirmation dashboard-background">
         <h1 class="govuk-panel__title text-left">
           <%= @ips.count %>
         </h1>
         <p class="govuk-panel__body text-left govuk-!-font-size-27">
-          Active IPs
+          <%= link_to "Added IPs", ips_path, class: "govuk-link govuk-link--no-visited-state dashboard-link-colour" %>
```
kjdchapman:
```suggestion
          <%= link_to "IPs", ips_path, class: "govuk-link govuk-link--no-visited-state dashboard-link-colour" %>
```

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  get 'setup/index'
```
kjdchapman: I would stick with [resourceful routing](https://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default) for this, mainly because that's how we've managed routes elsewhere.

(assuming we renamed from `setup` to `setup_instructions`)
```suggestion
  resources :setup_instructions, only %i[Index]
```

spec/features/view_overview_dashboard_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        location_one.update(radius_secret_key: radius_secret_key)
+
+        create(:ip, address: ip_one, location: location_one)
+        create(:ip, address: ip_two, location: location_one)
+
+        sign_in_user user
+        visit root_path
+      end
+
+      it 'shows the dashboard' do
```
kjdchapman: Careful about using two words for the same thing - overview/dashboard.

I'd assert on (whichever term you go with): `expect(page).to have_content("Overview")`

app/controllers/home_controller.rb
Renny4Real: Initially, this was written using `redirect and return` and all the current tests passed. 
However, the linter prefers `redirect && return` and this caused some tests (and some unrelated tests to fail).

I have time-boxed, researched and had a play around with this but changing from `redirect and return` to `redirect && return` still caused tests to fail. 

- Currently (at time of writing this) I have refactored the conditional to look like:
`  return redirect_to admin_organisations_path if current_user.super_admin?
    redirect_to setup_index_path unless current_organisation.ips.present?` which passes the tests. 
**Still I'm open to insight/thoughts into what may be causing `redirect && return` to fail some tests.**

Renny4Real:
**UPDATE:**
Added functionality (whilst renaming the links) to pluralise the links depending on their count.

screenshot 2019-01-07 at 15 04 37

app/controllers/home_controller.rb
kjdchapman: Yeah, I'm not sure how to resolve this - think you're right to timebox and move on.

app/views/home/index.html.erb
```diff
@@ -1,109 +1,36 @@
...
+  <div class="govuk-grid-column-full" id="overview">
+    <h2 class="govuk-heading-l govuk-grid-column-full">Overview</h2>
+    <div class="govuk-grid-column-one-third" id="team-members">
+      <div class="govuk-panel govuk-panel--confirmation summary-background">
         <h1 class="govuk-panel__title text-left">
           <%= @team_members.count %>
         </h1>
         <p class="govuk-panel__body text-left govuk-!-font-size-27">
-          Team Members
+          <%= link_to "#{"Team Member".pluralize(@team_members.count)}", team_members_path, class: "govuk-link govuk-link--no-visited-state summary-link-colour" %>
```
kjdchapman: I'd be happy pluralizing these in all cases - when it's a status screen like this I think the plural headings ("Team Members") make sense even if it's 0 or 1.

c.f. [Locations](https://github.com/alphagov/govwifi-admin/pull/392/files#diff-d9db7f7cb7dcd9bbe759b990b685bddaR20) and [IPs](https://github.com/alphagov/govwifi-admin/pull/392/files#diff-d9db7f7cb7dcd9bbe759b990b685bddaR30).

(Sorry, missed this first time round! Not a big deal, could be a "merge and deal with later" situation)

Renny4Real: _Status Screen_. Never thought about it that way. 🤔 
Now that I think about it, leaving them as plurals does look better and consistent. Thanks!

PR #394, opened 2019-01-07
Giving admins more information

georgeschena:
Add Beta phase banner 

screenshot 2019-01-07 at 12 26 27

Add some form of time frame when we will reply to support enquires

screenshot 2019-01-07 at 12 27 37


georgeschena:
Yeah I realised this after. My bad @kjdchapman

PR #402, opened 2019-01-10
Prevent loopback IP addresses from being used

cybojenix:
This is useful for 2 things:
- It prevents an admin entering an IP address that will never work for them
- It prevents an admin accidentally/maliciously trying to conflict with any internal clients we use for healtchecks

### DIFF
--- a/lib/use_cases/administrator/check_if_valid_ip.rb
+++ b/lib/use_cases/administrator/check_if_valid_ip.rb
@@ -13,7 +13,9 @@ module UseCases
       def valid_address?
         address.present? &&
           address_is_ipv4? &&
-          !(address_is_subnet? || address_allows_all?)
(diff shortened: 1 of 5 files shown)


lib/use_cases/administrator/check_if_valid_ip.rb
```diff
@@ -13,7 +13,9 @@ def execute(address)
       def valid_address?
         address.present? &&
           address_is_ipv4? &&
-          !(address_is_subnet? || address_allows_all?)
+          !address_is_subnet? &&
```
kjdchapman: Think we should probably invert & rename these methods, at a scan the `!` can get lost.

Suggest `address_is_not_subnet?`, `address_is_not_wildcard?`, `address_is_not_loopback?`?

PR #403, opened 2019-01-10
Fix the colour of the overview links in unique browser scenarios

Renny4Real:
**WHY:**
In certain instances, the link styles for the links in the overview would be a different colour and/or disappear when the mouse hovers over it. One such instance is the incognito mode in Google Chrome.

**IN THIS PR:**
More CSS pseudo-classes were added to the `summary-link-colour` CSS class to work around the limitations some browsers may put on changing the colour of links, especially visited links.

**BEFORE:**

screenshot 2019-01-10 at 11 29 06

**Google Chrome's Incognito Mode**

screenshot 2019-01-10 at 11 29 09

**Hovering over a link in this state (note: IPs has disappeared**

**AFTER:**

screenshot 2019-01-10 at 11 41 18

### DIFF
--- a/app/assets/stylesheets/components/overview.scss
+++ b/app/assets/stylesheets/components/overview.scss
@@ -13,7 +13,3 @@
 .summary-link-colour:hover {
   color: govuk-colour("white")
 }
-
-.summary-link-colour:active {
-  color: govuk-colour("white")
-}
--- a/app/assets/stylesheets/components/overview.scss
+++ b/app/assets/stylesheets/components/overview.scss
@@ -2,6 +2,18 @@
   background: govuk-colour("light-blue");
 }
 
+.summary-link-colour:link {
+  color: govuk-colour("white")
+}
+
 .summary-link-colour:visited {
+  color: govuk-colour("white")
+}
+
+.summary-link-colour:hover {
+  color: govuk-colour("white")
+}
+
+.summary-link-colour:active {
   color: govuk-colour("white")
 }


kjdchapman:
If I understand pseudo-classes right, they're just further levels of specificity on top of existing selectors... and it looks like we're picking... all the options. 

Maybe you already tried it, but does this work?:

```scss
.summary-link-colour {
  color: govuk-colour("white")
}
```

Or maybe these rules are to override some browser behaviour - if they're effectively injecting their own declarations with pseudo-classes, the above might not be specific enough?

Renny4Real:
Good call...I think it has to do with how the gov-link class is working. The links here are using the `govuk-link--no-visited-state` modifier class. 
I've tried setting the `summary-link-colour` class as white (with and without this gov-uk modifier class i.e `govuk-link--no-visited-state`) and the styles of the links don't work as desired across all cases.

**1.**
screenshot 2019-01-10 at 12 11 55

**With the gov-uk modifier class and `summary-link-colour` set to white**

**2.**
screenshot 2019-01-10 at 12 14 10

**Without the gov-uk modifier class i.e using only `govuk-link summary-link-colour` and `summary-link-colour` set to white**
_The blank link is the result of hovering over it._

**3.**
The choice of using the gov-uk modifier class was based on keeping the colour of the link consistent no matter the state. Changing the CSS style for only the _visited_ state of the link when using this modifier class seemed to do the job except in those unique browser cases.

screenshot 2019-01-10 at 12 19 16

**With the gov-uk modifier class & only `summary-link-colour:visited` set to white**
_The normal cases_

Researching on how to fix the bug led me to some info on how some browsers have restrictions on what you can do with the `:visited` CSS Style.
A couple of the links I came across: 

- Section on [Styling restrictions](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited )
- [CSS :visited Selector](https://www.w3schools.com/cssref/sel_visited.asp)

So I played around with each CSS pseudo class style to see which ones simulated the desired behaviour of the links in the unique cases (without breaking the style in normal cases). The pseudo-classes were added iteratively i.e trying a class style and seeing its effects on the whole.

For example, in the case of the changes made in this PR, if I remove only `summary-link-colour:link` and leave all the other changes. The normal browser cases still worked but the incognito browser case gave this:

screenshot 2019-01-10 at 13 06 26

**UPDATE:** But I have found that removing `summary-link-colour:active` works for all the cases. I tried your suggestion and found this out in process, Thanks! 👍

PR #404, opened 2019-01-10
Search logs by location

kjdchapman:
Wanted to give some visibility on this as it's been off `master` a while.

This PR adds searching by location. In its current state there's:
- some of duplication across searching by username/IP/location
- how the functionality is spread across the controller, use case and gateway doesn't look right

But it is currently in a working state, so just a little cleanup to go:

image

image


spec/features/logging/choosing_how_to_filter_spec.rb
```diff
@@ -5,8 +5,39 @@
...
 
   context 'with no filter chosen' do
-    before { click_on "Go to search" }
+    before { click_on 'Go to search' }
 
     it_behaves_like 'errors in form'
   end
+
+  # this'll add setup to the other tests too, if implemented
+  xcontext 'with no IPs' do
```
kjdchapman: Presented for discussion. 

Not sure if we need to worry about this case right now, got a few ways we could handle it, too (my other major idea was "hide logs in the subnavigation menu 'til you've got an IP")

app/controllers/logs_controller.rb
```diff
@@ -1,14 +1,29 @@
 class LogsController < ApplicationController
   def index
-    logs = UseCases::Administrator::GetAuthRequests.new(
+    use_case = UseCases::Administrator::GetAuthRequests.new(
```
kjdchapman: This controller is the main bit I'm not happy about. Feels messy.

spec/features/logging/choosing_how_to_filter_spec.rb
Renny4Real: Hiding it seems like a better method. The need to view logs doesn't seem realistic without an IP to search against and username logs shouldn't exist if there are no IPs to receive the traffic.

kjdchapman: With some time to think about it, I agree. It's a pretty simple solution, too.

app/controllers/logs_controller.rb
```diff
@@ -1,14 +1,46 @@
...
+      redirect_to(location_new_logs_search_path) && return
+    end
+
+    if valid_search_params.blank?
+      redirect_to(new_logs_search_path) && return
+    end
+
+    @location_name = Location.find_by(id: params[:location])&.address
+
+    @logs = get_auth_requests
```
emileswarts: No biggie, but given this is only called in one place I would generally to not pull this out into a method.  I always try only do restful methods in controllers

app/controllers/logs_controller.rb
```diff
@@ -1,14 +1,46 @@
...
-    logs = UseCases::Administrator::GetAuthRequests.new(
+    if location_outside_organisation?
+      redirect_to(location_new_logs_search_path) && return
+    end
+
+    if valid_search_params.blank?
+      redirect_to(new_logs_search_path) && return
+    end
+
+    @location_name = Location.find_by(id: params[:location])&.address
```
emileswarts: .find is designed to look up by id
```
Location.find(params[:location])
```

And would probably call it @location_address ?

Also wonder if we could say:
```
current_organisation.locations.find(params[:location])
```

which would negate the need to validate that the location belongs to the organisation?

app/controllers/logs_controller.rb
```diff
@@ -1,14 +1,46 @@
...
     )
+  end
 
-    @logs = logs.fetch(:results)
+  def ip_params
+    if params[:location].present?
+      Location.find(params[:location]).ips.map(&:address)
+    else
+      params[:ip] ? [params[:ip]] : nil
+    end
```
emileswarts: Still wonder if this above is the responsibility of the use case.
Happy to discuss but feel like some of its responsibilities have leaked out into this controller.

app/views/logs/index.html.erb
```diff
@@ -51,9 +53,11 @@
   </table>
 <% else %>
   <div class="govuk-grid-row">
-    <h1 class="govuk-heading-l govuk-grid-column-full">
+    <h1 class="govuk-heading-l govuk-grid-column-three-quarters">
       <% if params[:username].present? %>
         "<%= params[:username] %>" <strong>isn't reaching the GovWifi service</strong>
+      <% elsif params[:location] %>
+        Traffic from <%= @location_name %> <strong>isn't reaching the GovWifi service</strong>
```
emileswarts: Something about this repetition makes me think that the view knows to much about the internals of the search.

I know we don't use presenters but an object that has properties like:
{
search_term: 'foo',
result_count: 39,
results: [...]
}

Would be easy to scale without introducing more branching logic.
Thoughts?

app/views/logs_searches/ip.html.erb
```diff
@@ -5,7 +5,7 @@
     <%= link_to "Back to search options", new_logs_search_path, class: "govuk-back-link" %>
   </div>
 
-  <h1 class='govuk-heading-l govuk-grid-column-full''>Search logs by IP address </h1>
+  <h1 class='govuk-heading-l govuk-grid-column-full'>Search logs by IP address </h1>
```
emileswarts: 😲

lib/gateways/sessions.rb
```diff
@@ -3,15 +3,18 @@ class Sessions
...
     end
 
-    def search(username: nil, ip: nil)
-      query = username.present? ? { username: username } : { siteIp: ip }
+    def search(username: nil, ips: nil)
+      results = Session
+        .where(siteIP: organisation_ips)
+        .where('start >= ?', 2.weeks.ago)
+        .order(start: :desc)
+        .limit(MAXIMUM_RESULTS_COUNT)
```
emileswarts: We're running 2 queries?
Does this above load the results into memory?

lib/gateways/sessions.rb
```diff
@@ -3,15 +3,18 @@ class Sessions
     MAXIMUM_RESULTS_COUNT = 500
 
     def initialize(ips:)
-      @ips = ips
+      @organisation_ips = ips
```
emileswarts: Just curious what the concern was with this variable renaming?

app/controllers/logs_controller.rb
kjdchapman: `.find` throws a RecordNotFound exception if it fails - using `find_by` instead allows us to retain the control flow.

I like your solution better though - just calling `find` and 404ing if it's not a location.

app/controllers/logs_controller.rb
kjdchapman: Yeah, I went back and forth on this, it looked messy in-place. Gonna look at it again

RESTful-only in controller is interesting, I'm definitely happy to use arbitrary private methods. I can think of some reasons to avoid extra methods, but interested to hear your reasons?

app/controllers/logs_controller.rb
kjdchapman: Mm, it's a lot easier to sink some of this into the use case now it's cleaner, but there were some challenges around it that stopped me. 

I'll address the other things, look at this again, then let's chat.

app/views/logs/index.html.erb
kjdchapman: Yeah, I like this. Considered doing it, but was worried about how much time I'm spending on this.

If we're happy for me to look at the use case some more, I could do it as part of that.

app/views/logs_searches/ip.html.erb
kjdchapman: Sneaky apostrophes.

lib/gateways/sessions.rb
kjdchapman: If this is using arel under the hood, it's lazy and won't execute the query 'til the `.map` call. I can double-check that, though.

lib/gateways/sessions.rb
kjdchapman: I think it's not clear to the reader / consumer of the use case what these `ips` are, because there are two concepts named `ips`.

Happily taking suggestions on names to differentiate _ips that I'm allowed to show because they belong to the organisation_ and _ips that I've requested to filter the search by_.

(I actually wanted to rename the named parameter too, but that broke the interface so I left it 'til later.)

spec/features/logging/view_auth_requests_for_a_location_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        expect(page).to have_content("Traffic from #{location.address}"\
+          " isn't reaching the GovWifi service")
+      end
+    end
+
+    context 'choosing a location with logs' do
+      let(:ip) { create(:ip, location: location) }
+      let(:other_ip) { create(:ip) }
+
+      # TODO: session factory?
```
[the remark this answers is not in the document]
kjdchapman: Yeah, I shouldn't push PRs that have `# TODO` in though, my bad :blush:

app/controllers/logs_controller.rb
kjdchapman: Yeah, with a bit of a play I like this structure better. Pushed [some changes to the controller](https://github.com/alphagov/govwifi-admin/pull/404/commits/c83117a178aa650da4afb7cd71a14752f90a259a).

app/controllers/logs_controller.rb
kjdchapman: Still interested to hear your opinions, but think this is addressed in https://github.com/alphagov/govwifi-admin/pull/404/commits/c83117a178aa650da4afb7cd71a14752f90a259a

lib/gateways/sessions.rb
emileswarts: Ah right yes I see what you mean

app/views/logs/index.html.erb
kjdchapman: Talked offline, we're happy to move past this for now. Would like to come back to it at some point.

app/controllers/logs_controller.rb
kjdchapman: This was a good spot, I've rearranged this into something less painful looking now. Think the code that deals with "make a location into an ip list" now looks a lot cleaner.

app/controllers/logs_controller.rb
kjdchapman: And as I played with it more, I actually ended up putting this back in (https://github.com/alphagov/govwifi-admin/pull/404/commits/f37358b5b0caa9e5a2dfc435f6f6c9d05af04596). I think the use case's composition isn't the most important thing to a reader, so sinking it down one level of indirection feels right.

PR #409, opened 2019-01-14
Add view of number of unique user connections in last twenty four hours

Renny4Real:
**WHY:**
 It was suggested to add a view of the number of unique user connections within an organisation as a way to further describe the current installation that an organisation has.

**IN THIS PR:**
The activity section is another section on the overview page (below the overview summary).
This section displays a sentence informing the admin user of the number of **UNIQUE SUCCESSFUL** connections within their organisation that were made within a period of 24 hours.

----------------------------------------------------------------------------------------------------

**BEFORE:** 

screenshot 2019-01-14 at 15 38 22

**AFTER:**

screenshot 2019-01-15 at 17 36 03

----------------------------------------------------------------------------------------------------

- `unique_user_count` method returns the number of unique & successful connections made in an organisation without/within a given time frame.
- A new use case called `GetUniqueUserRequests` and new gateway called `UniqueConnections` were added based on feedback.
 - `UniqueConnections` contains the `unique_user_count` method.
- `GetUniqueUserRequests` uses the gateway to obtain the number of unique & successful connections and returns a hash of `connection_count`.
- The value of `connection_count` is then assigned to an instance variable (`@connections`) in the `HomeController` and added to the appropriate section in the `index.html` file for the Overview page.

**Any suggestions and/or feedback would be appreciated.**


Renny4Real:
**UPDATE:**
Add conditional statement to determine the form of message based on number of connections.

screenshot 2019-01-14 at 16 40 48

app/controllers/home_controller.rb
```diff
@@ -4,11 +4,22 @@ def index
 
     redirect_to setup_instructions_path unless current_organisation.ips.present?
 
+    connections = UseCases::Administrator::GetAuthRequests.new(
+      authentication_logs_gateway: Gateways::Sessions.new(
+        ips: current_organisation.ips.map(&:address)
+      )
+    ).execute(
+      username: nil,
+      ips: @ips
```
kjdchapman: This variable (`@ips`) isn't defined until [line 7](https://github.com/alphagov/govwifi-admin/pull/409/files#diff-14dfef9f01460f69f175a8e6bd7d143dR16), so it'll be `nil` at this point. 

You'll need to define `@ips` before you call  `execute` on this use case.

spec/gateways/sessions_spec.rb
```diff
@@ -123,4 +122,44 @@
...
+        Session.create(start: today_date, success: false, username: username_1, siteIP: '127.0.0.2')
+        Session.create(start: today_date, success: true, username: username_2, siteIP: '127.0.0.3')
+        Session.create(start: today_date, success: false, username: username_2, siteIP: '127.0.0.4')
+        Session.create(start: today_date, success: false, username: username_2, siteIP: '127.0.0.5')
+
+        result = subject.count_distinct_users
+        expect(result).to eq(2)
+      end
+
+      it 'counts the number of unique successful connections' do
```
tommotaylor: It looks like this spec is testing the exact same thing as the spec above. Or am I missing something? If its not the same, it needs a better description so the next developer that comes along can see what specific scenario is being tested here.

lib/gateways/sessions.rb
```diff
@@ -28,6 +28,10 @@ def search(username: nil, ips: nil)
       end
     end
 
+    def count_distinct_users
```
tommotaylor: I would name this `unique_user_count`, it fits more with what the user would see. "distinct" is really just a ruby method. Its also common practice to have the `_count` but last in ruby.

lib/use_cases/administrator/get_auth_requests.rb
```diff
@@ -7,10 +7,7 @@ def initialize(authentication_logs_gateway:)
 
       def execute(username: nil, ips: nil)
```
tommotaylor: I'm wondering if we should rename these keyword args as `username_filter:` and `ip_filter:` to make them more descriptive. It took me a little while to figure out what these were doing as we already have an `ips:` argument for the  gateway. The fact that we now passing in `nil` also confuses it a bit.

lib/gateways/sessions.rb
```diff
@@ -28,6 +28,10 @@ def search(username: nil, ips: nil)
...
+      Session.where(siteIP: organisation_ips).where('start >= ?', 1.day.ago).where(success: true).distinct.count(:username)
```
tommotaylor: I find it a bit odd that this `UseCase` is used for searching through the past two weeks of logs, and at the same time it returns the number of unique connections in the past day. 

For instance it could return 500 results, as well as a `count_distinct_users` of `0`. This is pretty confusing. I was pondering wether this `count_distinct_users` should be in a different use case and couldn't really make up my mind, but this peculiarity makes me think it should be in a different use case. That way you can also get rid of the `ips:` and `username:` arguments which are superfluous in:
```
@connections = get_auth_requests.execute(
      ips: nil,
      username: nil
    ).fetch(:connection_count)
```

So you could have something like:
```
@connections = get_unique_user_requests_count.execute[:result]
```

Eventually we might end up extending it to take a date range, something like:
```
@connections = get_unique_user_requests_count.execute(date_range: 15.days)[:result]
```

kjdchapman: I would strongly vote for this UseCase being split up, if time allows.

Renny4Real:
@tommotaylor Thanks for the feedback! 
I did realise after pushing this up that the tests might confuse another developer reading it. I planned to refactor and clean up after getting some feedback so as to have a bit of a focus. 
Your argument for a separate Use case does sound good as I did get a bit confused with what `GetAuthRequests` was doing before trying to use it. 
I assumed at the time that it was better to keep everything concise which is why I tried to use the Use Case but I can understand how that might confuse another developer looking at the code.

lib/gateways/unique_connections.rb
```diff
@@ -0,0 +1,17 @@
+module Gateways
+  class UniqueConnections
+    def initialize(ips:)
+      @organisation_ips = ips
+    end
+
+    def unique_user_count(date_range: nil)
+      return Session.where(siteIP: organisation_ips).where('start >= ?', date_range).where(success: true).distinct.count(:username) if date_range.present?
```
tommotaylor: This line is really long. Short lines are easier to read.

tommotaylor:
Looks good to me Renny, nice work!

PR #410, opened 2019-01-17
Add better errors to logging pages

tommotaylor:
- Added the error highlighting class to each form field
- Added an error hint to each form field
- Improved the error messages
- Removed the duplicate error summaries when a search term was blank (it used to say "blank" and "invalid", now it just says "blank".

screenshot 2019-01-17 at 17 03 40
screenshot 2019-01-17 at 17 03 04
screenshot 2019-01-17 at 17 03 14
screenshot 2019-01-17 at 17 03 23
screenshot 2019-01-17 at 17 03 32

### DIFF
--- a/app/models/logs_search.rb
+++ b/app/models/logs_search.rb
@@ -4,8 +4,7 @@ class LogsSearch
   attr_accessor :filter, :first_step
   attr_writer :term
 
-  validates :term, presence: true
-  validate :validate_term
+  validate :validate_term, :validate_presence
 
   def term
     @term&.strip
@@ -22,16 +21,24 @@ private
     end
   end
 
+  def validate_presence
+    if term.length == 0
+      self.errors.add(:search_term, 'cannot be empty')
+    end
+  end
+
   def validate_username
-    unless term.length == 5 || term.length == 6
-      self.errors.add(:term, 'must be 5 or 6 characters')
+    unless term.length == 5 || term.length == 6 || term.length == 0
+      self.errors.add(:search_term, 'must be 5 or 6 characters')
     end
   end
 
   def validate_ip
+    validate_presence if term.length == 0 && return
+
     checker = UseCases::Administrator::CheckIfValidIp.new
     unless checker.execute(self.term)[:success]
-      errors.add(:term, "must be a valid IP")
+      errors.add(:search_term, 'must be a valid IP address')
     end
   end
 end
--- a/app/models/logs_search.rb
+++ b/app/models/logs_search.rb
@@ -22,19 +22,21 @@ private
   end
 
   def validate_presence
-    if term.length == 0
-      self.errors.add(:search_term, 'cannot be empty')
-    end
+    self.errors.add(:search_term, 'cannot be empty') if term.empty?
   end
 
   def validate_username
-    unless term.length == 5 || term.length == 6 || term.length == 0
+    if term_is_present_but_incorrect_length
       self.errors.add(:search_term, 'must be 5 or 6 characters')
     end
   end
 
+  def term_is_present_but_incorrect_length
+    term.present? && term.length != 5 && term.length != 6
+  end
+
   def validate_ip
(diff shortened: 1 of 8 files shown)


app/models/logs_search.rb
```diff
@@ -22,16 +21,24 @@ def validate_term
     end
   end
 
+  def validate_presence
+    if term.length == 0
+      self.errors.add(:search_term, 'cannot be empty')
+    end
+  end
```
tommotaylor: I tried to use the `ActiveRecord` type I18n validations in `en.yml` here but I couldn't get anything to work. 
- I tried including `ActiveMode::Translations`, 
- I also tried using the `activemodel:` key in the `en.yml` file, 
- I also tried adding a message directly to `validate :term, presence: { message: "Hello"}`

I couldn't get any of that to work correctly so I went for a custom `validate_presence` method here in the model.

tommotaylor: It actually works better this way as it means we avoid the duplicate error summary eg:
screenshot 2019-01-17 at 17 41 03

app/models/logs_search.rb
```diff
@@ -22,16 +21,24 @@ def validate_term
     end
   end
 
+  def validate_presence
+    if term.length == 0
```
Renny4Real: I think `term.length == 0` can be written as `term.length.zero?`. 
I remember getting linting errors for this.

kjdchapman: Or `term.empty?` (or `term.blank?` if you want to use ActiveSupport)

app/models/logs_search.rb
kjdchapman: I'm fine with this - it might surprise a reader a little ("why didn't they use the standard validations here?"), but it's not unclear what's going on.

The bonus single error message in the summary is nice, too.

app/models/logs_search.rb
```diff
@@ -22,16 +21,24 @@ def validate_term
...
+      self.errors.add(:search_term, 'cannot be empty')
+    end
+  end
+
   def validate_username
-    unless term.length == 5 || term.length == 6
-      self.errors.add(:term, 'must be 5 or 6 characters')
+    unless term.length == 5 || term.length == 6 || term.length == 0
```
kjdchapman: I don't understand why this test is here and in `validate_presence`.

app/views/logs_searches/_error_message.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if field_error(@search, key).present? %>
+  <p class="govuk-body govuk-error-message">Please <%= message %></p>
```
kjdchapman: [GOV Design System](https://design-system.service.gov.uk/components/error-message/#be-clear-and-concise) says "Do not use ‘please’ because it implies a choice".

```suggestion
  <p class="govuk-body govuk-error-message"><%= message %></p>
```

(not sure if you'll need to mess with capitalisation because of this)

kjdchapman:
@tommotaylor Just in case there's any trouble with VPN/Jenkins access from NZ, the reason this is failing on jenkins is just because of linting. Local `make lint` should sort it out.

(I would have just sorted it, but some of the code it's complaining about relates to some of the comments Renny and I have made, and I don't want to confuse things).

app/views/logs_searches/_error_message.html.erb
tommotaylor: 👍

app/models/logs_search.rb
tommotaylor: 👍

app/models/logs_search.rb
tommotaylor: This is to skip adding the error `'must be 5 or 6 characters'` if the search term is empty. This avoids the duplicate error message.

Achieves the same as line 37:
```
  def validate_ip
    validate_presence if term.empty? && return
```

PR #412, opened 2019-01-17
Fix confusion when an organisation is adding IPs for the first time with no existing locations

Renny4Real:
**WHY:**
When a user's organisation has no locations and the user wants to add an IP, they get an empty drop-down selector in the "Add IP address" flow. A link to `Add a new location` is present but may cause confusion for a user whose desire is to **add a new IP address** since it is not stated that a location is required first before adding an IP address.

**Current Display for organisations without any locations:**
screenshot 2019-01-17 at 12 17 25

**Current Display for organisations with at least one location:**
screenshot 2019-01-17 at 12 21 48

**IN THIS PR:**
- A working suggestion is to redirect the user to the `new_location_path` when they click on the `add an IP address` button, if their organisation has not added any locations. 
- When they land on the page, a message is displayed to explain to the user that their organisation has no locations to add an IP address to. This message will only been shown if the user's organisation has no locations.

**Step 1:**
screenshot 2019-01-17 at 11 58 22

**Step 2:**
screenshot 2019-01-17 at 11 57 20

**Suggestions/feedback would be appreciated.**

### DIFF
--- a/app/views/locations/new.erb
+++ b/app/views/locations/new.erb
@@ -19,7 +19,7 @@
     <%= link_to 'Back to list', ips_path, class: 'govuk-back-link' %>
 
     <% if current_organisation.locations.empty? %>
-      <h2 class='govuk-heading-l'>Add your first location and IP address(es)</h2>
+      <h2 class='govuk-heading-l'>Add your first location and IP addresses</h2>
     <% else %>
       <h2 class='govuk-heading-l'>Add a new location</h2>
     <% end %>
--- a/app/views/locations/new.erb
+++ b/app/views/locations/new.erb
@@ -15,18 +15,16 @@
 <% end %>
 
 <div class='govuk-grid-row'>
-  <div class='govuk-grid-column-two-thirds'>
+  <div class='govuk-grid-column-full'>
     <%= link_to 'Back to list', ips_path, class: 'govuk-back-link' %>
 
     <% if current_organisation.locations.empty? %>
-      <h2 class='govuk-heading-l'>Add a new location</h2>
-      <div class="govuk-inset-text">
-        <strong>You have no existing locations to add a new IP address to.</strong>
-      </div>
+      <h2 class='govuk-heading-l'>Add your first location and IP address(es)</h2>
     <% else %>
       <h2 class='govuk-heading-l'>Add a new location</h2>
     <% end %>
-
+  </div>
+  <div class='govuk-grid-column-two-thirds'>
     <%= form_for @location, html: { autocomplete: "off" } do |f| %>
       <div class="govuk-form-group">
         <%= f.label :address, class: 'govuk-label' %>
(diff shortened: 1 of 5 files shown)


app/views/locations/new.erb
```diff
@@ -18,7 +18,14 @@
   <div class='govuk-grid-column-two-thirds'>
     <%= link_to 'Back to list', ips_path, class: 'govuk-back-link' %>
 
-    <h2 class='govuk-heading-l'>Add a new location</h2>
+    <% if current_organisation.locations.empty? %>
```
kjdchapman: I like the idea, but I worry that it'll be brittle. If we changed the way the user flows through the app, the message might end up confusing.

Still, it's better than what we have now - if no-one else is weighing in with objections, I'd be happy for us to start with this.

app/views/locations/new.erb
```diff
@@ -18,7 +18,14 @@
...
+      <h2 class='govuk-heading-l'>Add a new location</h2>
+      <div class="govuk-inset-text">
```
kjdchapman: I'm not sure if the `inset-text` styling is useful to highlight this message or not. I'd try it as just a paragraph and see what it looks like.

app/views/locations/new.erb
```diff
@@ -18,7 +18,14 @@
...
+        <strong>You have no existing locations to add a new IP address to.</strong>
```
kjdchapman: I'm trying to work out what confusion this message deals with, or if it's just as clear without it.

I considered:
```suggestion
        <strong>You must add a location with your first IPs.</strong>
```
but this might suggest to some users that they can't use this form yet, and need to find some "new location" in the app before they're allowed to use it.

Then there's just the straight re-word of what you have:
```suggestion
        <strong>You have no locations to which you can add a new IP.</strong>
```
but I think this has the same problem as suggestion no.1, users might think they're not allowed to use this form yet.

I would vote for removing this text, and adding something back in if user research shows users are confused when they land here from "Add IP address".

app/views/locations/new.erb
kjdchapman: In case confusing in lieu of later comments - this is only a relevant comment if you don't agree with my recommendation to remove this text!

app/views/locations/new.erb
kjdchapman: In case it's confusing in lieu of later comments - this is only a relevant comment if you don't agree with my recommendation to remove this text!

Renny4Real:
**UPDATE:**
- Change the path for the `add the IPs` link on the setting up page based on whether the user's organisation has no locations (redirects to `new_location_path`) or just has no IPs (redirects to `new_ips_path`). 

- Remove the inset text and attempt to make the form heading more explanatory.

screenshot 2019-01-17 at 16 58 49

PR #413, opened 2019-01-17
Improve signed-in support page

kjdchapman:
### This PR:
- Uses account email instead of asking a user to enter it
- Gives clearer direction to other ways of resolving problems
- Cuts out superfluous text
- Removes back link (there are multiple ways to get here and plenty of escape routes, "back" is more likely to confuse than help)

### OLD

image

### NEW

image

This page is serving multiple needs, I would expect to split it out in the near future, but think this is an improvement already.

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -35,6 +35,10 @@ class HelpController < ApplicationController
   def create
     @support_form = SupportForm.new(support_form_params)
 
+    if @support_form.email.blank?
+      @support_form.email = current_user&.email
+    end
+
     if @support_form.valid?
       template_id = GOV_NOTIFY_CONFIG['help_email']['template_id']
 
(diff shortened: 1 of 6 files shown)


app/controllers/help_controller.rb
```diff
@@ -35,6 +35,10 @@ def signed_in
   def create
     @support_form = SupportForm.new(support_form_params)
 
+    if @support_form.email.blank?
```
Renny4Real: Would this affect any other place we've used a support form? e.g. offline? or is this the only place?

kjdchapman: This does evaluate to `true` if you post empty emails on any of the other forms, but the next line does nothing if `current_user` isn't present. 

The `&.` on that [next line](https://github.com/alphagov/govwifi-admin/pull/413/files/2ce3d029250e93de7d24cd535e2ef2504a411614#diff-32e4cccbb561b8d9e869a635fbcba1eeR39) is the "[safe navigator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/)" operator. It's a terser way to write `unless thing_you_are_working_with.nil?`, but using it might sometimes make things less readable.

I also tried:

```
if @support_form.email.blank? & current_user.present?
  @support_form.email = current_user.email
end
```

but felt that long test was less readable.

PR #414, opened 2019-01-17
Stop IPs in seeds.rb clashing

kjdchapman:
I thought this'd be very unlikely, but it's happened to me twice now, so I'm fixing it.


tommotaylor:
Nice, this happened to me yesterday as well.

PR #416, opened 2019-01-18
Allow users to delete empty locations

tommotaylor:
Allows a user to delete a location if:
- they have "manage locations" permissions
- they own the location
- the location has no IPs

Not 100% sure on location of "Remove this location" link.

Step 1:
screenshot 2019-01-18 at 15 05 02

Step 2:
screenshot 2019-01-18 at 15 06 48

Step 3:
screenshot 2019-01-18 at 15 06 54


spec/requests/locations/delete_spec.rb
```diff
@@ -0,0 +1,41 @@
+describe "DELETE /locations/:id", type: :request do
+  let(:user) { create(:user) }
+  let!(:location) { create(:location, organisation: user.organisation) }
+
+  before do
+    https!
```
Renny4Real: What does this do?

kjdchapman:
I think the "Remove location" link location is good, because it's only for orgs with no IPs, it doesn't get lost visually.

PR #419, opened 2019-01-21
Organisation Register Sign Up

georgeschena:
We saw issues where admins signed up the same organisation multiple times causing 'ghost' organisations to persist in the database. We changed the organisation name input from a free text field to a select dropdown populated by the organisation names found within the Govuk register API.

We had to refactor the tests due to the fact we had to only test from the organisations in the register, this reducing the amount we could use to test with.


emileswarts:
Realise we never checked what this looks like on mobile.

tommotaylor:
@georgeschena @emileswarts approved, but I'll let you guys remove that duplication of `FactoryBot.reload!`

PR #420, opened 2019-01-21
Fix visibility of the top navigation links on mobile view

Renny4Real:
**WHY:**
When viewed on a mobile device, the top navigation links: `Documentation`, `Support` & `Sign Out`(or `Sign in` depending on the user's state) disappeared from the user's view.

**PREVIOUS DISPLAY ON MOBILE:**
screenshot 2019-01-21 at 16 09 41

**IN THIS PR:**
- Include the gov-uk styled `button` html tag within the `application.html.erb` file.
- Add gov-uk classes to the `html` and `body` tags based on the html layout of the [Customised page template](https://design-system.service.gov.uk/styles/page-template/) example.
- Add span tags to align the menu to the right of the screen.
- Add media query for desktop and iPad Pro screens. Reasoning explained below:

The custom css class involved was done in a previous PR to keep a consistent layout when comparing the main navigations of the admin portal, product page & tech docs (explained in this commit: 5ceed80).

This custom css class however, overlaid the menu (as seen below) unto the rest of the page since its max height was hard set:

screenshot 2019-01-21 at 11 54 07

The media query was added to preserve the main navigation consistency (referred to in this commit: 5ceed80) and to prevent the styling from affecting devices with a browser window less than 1024 px (i.e. mobile devices). 

1024 px was chosen as a result of checking the view through various screen sizes.

**iPad view- (768 x 1024)**
screenshot 2019-01-21 at 15 38 26

**iPad Pro view- (1024 x 1366)**
screenshot 2019-01-21 at 15 38 37

At 1024 px, it was safe to assume that the styling that creates the menu button would no be longer active. 

**CURRENT DISPLAY ON MOBILE:**

**Example 1:**
screenshot 2019-01-21 at 15 11 45

**Example 2:**
screenshot 2019-01-21 at 15 12 00

**Any suggestions and/or feedback would be appreciated**

### DIFF
--- a/app/assets/stylesheets/components/main_nav.scss
+++ b/app/assets/stylesheets/components/main_nav.scss
@@ -1,6 +1,8 @@
-.main-nav {
-  display: flex;
-  justify-content: space-between;
+@media only screen and (min-width: 1024px) {
+  .main-nav {
+    display: flex;
+    justify-content: space-between;
+  }
 }
 
 .product-name{
--- a/app/assets/stylesheets/components/main_nav.scss
+++ b/app/assets/stylesheets/components/main_nav.scss
@@ -18,9 +18,11 @@
   padding: 9px 0;
 }
 
-.govuk-header__container {
-  max-height: 40px;
-  padding-top: 11px;
+@media only screen and (min-width: 1024px) {
+  .govuk-header__container {
+    max-height: 40px;
+    padding-top: 11px;
+  }
 }
 
 .govuk-header__logotype-crown {
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" class="govuk-template ">
+<html lang="en" class="govuk-template app-html-class">
   <head>
     <!-- test post please ignore -->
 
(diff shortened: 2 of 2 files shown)


Renny4Real:
@emileswarts This is styled slightly differently due to our custom CSS classes. The nav links are on the left for the tech docs and the product page. The `span` tags were there before and I think they help with some form of styling on the navigation.

Renny4Real:
**UPDATE:**
Get the navigation links to show on the left (instead of right) whilst maintaining the styling of the header on the desktop view. 
`Tech Docs` and `Product page` have their menu links display on the left side. Another GDS site has the menu styled like this also. 
See below.
 
**SCREENSHOT FROM TECH DOCS:**
screenshot 2019-01-21 at 17 05 59

Renny4Real:
Removing the `main-nav` CSS class moved the navigation links to the left but compromised the position of the links in larger browser windows:

**VIEW IN LARGE BROWSER WINDOWS (WITHOUT `main-nav` CSS):**

screenshot 2019-01-22 at 10 41 29

**VIEW IN LARGE BROWSER WINDOWS (WITH `main-nav` CSS):**
screenshot 2019-01-22 at 10 40 45

To keep the styling the same for large browser windows, a media query was added to the `main-nav` CSS class. This was done so that the CSS class did not affect small browser windows i.e mobile view and the navigation links were placed on the left.

**CURRENT DISPLAY ON MOBILE:**
screenshot 2019-01-22 at 10 25 02

**Any suggestions/feedback would be appreciated**

PR #423, opened 2019-01-23
Combine organisation/team members pages

Renny4Real:
**WHY:**
It was decided that the sub-navigation was bloated. 
A design suggestion was that the "team members" and "organisation settings" pages could be combined and viewed as one page.

**IN THIS PR:**
- The contents of the Team Members' `index.html.erb` was moved to the Organisations' `show.html.erb`.
- The `index` action for the `TeamMembersController` was changed to redirect to the `OrganisationsContoller`'s index action.
- The `OrganisationsContoller`'s index action renders the show page for the `OrganisationsContoller`.
- Small changes e.g. changing the name of back links, were also done.

**BEFORE:**

screenshot 2019-01-23 at 16 43 31

**AFTER:**

screenshot 2019-01-23 at 16 33 41

**Any suggestions/feedback on this approach would be appreciated**


app/controllers/organisations_controller.rb
```diff
@@ -2,7 +2,9 @@ class OrganisationsController < ApplicationController
   before_action :set_organisation, only: %i[edit update]
   before_action :validate_user_is_part_of_organisation, only: %i[edit update]
 
-  def index; end
+  def index
+    render :show
```
emileswarts: Why do we need to do this?

app/controllers/team_members_controller.rb
```diff
@@ -3,7 +3,7 @@ class TeamMembersController < ApplicationController
   before_action :validate_can_manage_team, only: %i[edit update]
 
   def index
-    @team_members = current_user&.organisation&.users || []
+    redirect_to organisation_path(current_organisation)
```
emileswarts: Can we remove this controller?

app/controllers/organisations_controller.rb
Renny4Real: This was done because of the redirect from the `TeamMembersController`. Wasn't sure if removing the `TeamMembersContoller` completely was the way to go. It seemed that the two controllers handled different things. e.g edit permissions vs edit organisation name.

app/controllers/team_members_controller.rb
Renny4Real: I don't mind to try if there's no objection to its removal.

kjdchapman:
I'm not sure if this is the right approach. 

With the current wording, the subnav hasn't actually gotten much smaller in terms of character count. With different wording, I think it'll be difficult to communicate what's under these headings.

I'd be happy to see this merged in with slightly terser headings (Setup Instructions -> Setup, Organisation Settings -> Settings), but I think we need to look at a third-level nav under a "Settings" tab, like Notify.

Might be worth Emile mentioning this a problem we're having in the design crit session this afternoon.

Renny4Real:
**UPDATE:**
Rearrange the view of the condensed page and rename the link to a shorter term.

screenshot 2019-01-25 at 10 44 56

**Any suggestions/feedback on this design approach would be appreciated.** 
Not sure if the `OrganisationsController` &`TeamMembersController` should be condensed into one controller since there is a lot of functionality in both but open to opinions.

kjdchapman:
I don't like the "invite team member" being outside of the context of the "team members" list, but I understand the desire for it to be prominent in the top-right. 

if it were me, I'd take out the `section-break` in between the main heading (**Settings**) and the first sub-heading (**Organisation**), and move the "Invite team member button down".

ashley-fg:
I think it looks better. I would agree with Kyle’s comment about removing the section break above ‘Organisation’. Also make ‘Change my org name & email’ the same size as ‘Edit permissions’. I would also change ‘Change my org name & email’ to just read: 'Change'. 

I agree that it feels weird that the button is not with the team members. Can it be put in line with the team members heading under the section break?

Renny4Real:
**UPDATE 2:**

screenshot 2019-01-25 at 13 14 32

@kjdchapman @ashley-fg Thanks, I've implemented the suggested changes.

ashley-fg:
I think it’s working better, but would be good to get a GDS designer’s perspective. Only thing I would question is whether the section breaks between the team members now make it look like they are new sections in the same way 'Organisations' and 'Team members' are separate sections.

kjdchapman:
Had a pass at this to try and make it clearer:
- "Change" -> "Edit organisation" for consistency with "Edit team members"
- Remove one subheader
- Correct the spacing in the second section

I agree with @ashley-fg 's point that the section breaks aren't good, but I can't see a better way to do it - suggestions very welcome.

image

I'd be happy for us to merge this as-is, as it's an improvement, then visit the nav as a wider piece later.

@emileswarts - did we get any comments on the nav as part of the design review?

emileswarts:
No comments about the nav no

app/controllers/organisations_controller.rb
```diff
@@ -15,6 +17,10 @@ def update
     end
   end
 
+  def show
+    @team_members = current_user&.organisation&.users || []
```
emileswarts: Would the first part of this conditional be an empty array anyway?

app/views/application/_subnavigation.html.erb
```diff
@@ -16,10 +16,9 @@
         <% if current_organisation.ips.present? %>
           <%= link_to "Overview", root_path, class: active_tab(root_path) %>
         <% end %>
-        <%= link_to "Setting Up", setup_instructions_path, class: active_tab(setup_instructions_path) %>
-        <%= link_to 'Organisation', organisation_path(current_organisation), class: active_tab("organisation") %>
-        <%= link_to "Team members", team_members_path, class: "#{active_tab("team")}" + "#{active_tab("user")}"%>
+        <%= link_to "Setup Instructions", setup_instructions_path, class: active_tab(setup_instructions_path) %>
```
emileswarts: Given we've had "the discussion", can we stick to single quotes when not interpolating?

app/views/organisations/show.html.erb
```diff
@@ -1,8 +1,30 @@
...
+    <h2 class="govuk-heading-l">Settings</h2>
+  </div>
+</div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">
+      Organisation Name: <b><%= current_organisation.name %></b>
+    </p>
+    <p class="govuk-body">
+      Service Email: <b><%= current_organisation.service_email%></b>
```
emileswarts: no space before closing %>

app/controllers/organisations_controller.rb
kjdchapman: I think safe navigation operator returns `nil` if the object you're calling on is `nil`?

```
irb(main):001:0> something = nil
=> nil
irb(main):002:0> something&.method
=> nil
```

app/views/application/_subnavigation.html.erb
kjdchapman: Took the decision to just fix this whole file, despite the noise it adds to the PR. 

I'd still be interested to hear opinions on this for future changes.

app/controllers/organisations_controller.rb
tommotaylor: Current user must have an organisation to sign in, and it will always have at least one user (the current user). So I wonder if we need any of the `&`s or the `|| []`.

tommotaylor:
Closed in favour of https://github.com/alphagov/govwifi-admin/pull/432

PR #432, opened 2019-01-31
Condense subnav

tommotaylor:
- Subnav condensed and rearranged
- Service email displayed on team page

screenshot 2019-01-31 at 13 22 07

### DIFF
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -2,8 +2,6 @@ class OrganisationsController < ApplicationController
   before_action :set_organisation, only: %i[edit update]
   before_action :validate_user_is_part_of_organisation, only: %i[edit update]
 
-  def index; end
-
   def edit; end
 
   def update
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -7,7 +7,7 @@ class OrganisationsController < ApplicationController
   def update
     if @organisation.update(organisation_params)
       redirect_to team_members_path
-      flash[:notice] = 'Organisation updated'
+      flash[:notice] = 'Service email updated'
     else
       render :edit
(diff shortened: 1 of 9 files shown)


georgeschena:
Should the title of the page be something like 'Team' or 'Team Settings' instead of 'Team members'?

tommotaylor:
@georgeschena The way I see it is this is the team page, and the service email is part of the team as its how the team receives their communications. Its like the "team email". As we can no longer change an organisation's name, this PR gets rid of the idea of "Organisation settings"

tommotaylor:
Having said that "Team settings" would probably look pretty good

config/routes.rb
```diff
@@ -39,7 +39,7 @@
     get 'location', on: :new
   end
 
-  resources :organisations, only: %i[show edit update]
+  resources :organisations, only: %i[edit update]
```
emileswarts: Do we need to keep the show action for the super admins?

app/controllers/organisations_controller.rb
```diff
@@ -2,14 +2,12 @@ class OrganisationsController < ApplicationController
...
 
-  def index; end
-
   def edit; end
 
   def update
     if @organisation.update(organisation_params)
-      redirect_to organisation_path(@organisation)
-      flash[:notice] = 'Organisation updated'
+      redirect_to team_members_path
```
emileswarts: I was tempted the first time I saw this to create a new controller which isn't tied to team_members or organisations.

It could just be called team and host the team_members and organisation functionality.

Not vital, happy to discuss or not do this at all.  Just thought I'd mention it.

config/routes.rb
tommotaylor: All super admin routes are nested under `namespace: :admin`

PR #436, opened 2019-02-01
Fix updating service email bug

georgeschena:


lib/gateways/organisation_register_gateway.rb
```diff
@@ -3,7 +3,7 @@
 
 module Gateways
   class OrganisationRegisterGateway
-    REGISTER_URL = "#{SITE_CONFIG['registers_url']}/records.json".freeze
+    REGISTER_URL = "#{SITE_CONFIG['registers_url']}/records.json?page-size=5000".freeze
```
kjdchapman: I'll allow it :man_judge:

PR #439, opened 2019-02-04
WIP: Implement local auths

georgeschena:


app/models/organisation.rb
```diff
@@ -16,7 +16,7 @@ def validate_in_register?
 
   def self.fetch_organisations_from_register
     UseCases::Organisation::FetchOrganisationRegister.new(
-      organisations_gateway: Gateways::OrganisationRegisterGateway.new
+      organisations_gateway: Gateways::OrganisationsGateway.new
```
emileswarts: I know there have been some discussions around this.

I thought Gateways did know about where they get data from.
Like S3 gateway or a database gateway.

For me GovUkRegisterGateway would be more descriptive here.
@kjdchapman You have other ideas about this?

kjdchapman: I liked @emileswarts 's idea of getting `GovUkRegister` in the name, feel like it should mention `Organisations` too though. 

Without namespacing it looks pretty beefy - `GovUkRegisterOrganisationsGateway` :grimacing:

spec/gateways/govuk_organisations_register_gateway_spec.rb
```diff
@@ -0,0 +1,25 @@
+describe Gateways::GovukOrganisationsRegisterGateway do
```
kjdchapman: I know this is pre-existing behaviour, but I don't like these specs relying on the behaviour of a docker container - feels like the set-up for this test is very distant to the assertion, which could be confusing for a reader.

I like the idea of the docker container for supporting local dev, but I'm not sure about it for automated testing. I know other people at Made Tech have used a similar strategy, be interested to hear how it panned out for them long-term.

This PR isn't a regression - this paradigm already exists in the codebase - so I don't want to block us moving forward. 
 Just wanted to share the thought out loud.

kjdchapman: And of course, the _moment_ I hit submit, I work out that they don't rely on the docker container at all, they rely on something in `spec_helper`.

That's better, but still suffering from the same problem - the setup is distant from the assertion. I think this could be pretty cleanly addressed with the use of `shared_context` - if that context is even for anything but this single test?

kjdchapman:
This looks good to me, but I'd prefer to leave it to @emileswarts to tick as he left comments originally.

app/models/organisation.rb
```diff
@@ -16,7 +16,7 @@ def validate_in_register?
 
   def self.fetch_organisations_from_register
     UseCases::Organisation::FetchOrganisationRegister.new(
-      organisations_gateway: Gateways::OrganisationRegisterGateway.new
+      govuk_organisations_register_gateway: Gateways::GovukOrganisationsRegisterGateway.new
```
emileswarts: I'd leave the param "govuk_organisations_register_gateway" as just "organisations_gateway".
This leaves it open for us to inject a different gateway in the future

spec/gateways/govuk_organisations_register_gateway_spec.rb
emileswarts: Yeah i hear you @kjdchapman.
Let's try move this setup closer to the assertions after we get this in.

PR #440, opened 2019-02-04
Correct reply address on support emails

kjdchapman:
Current implementation can post tickets into a sandbox ZD account. You'll need to set up env-vars in the `Dockerfile` to see this behaviour:

```Dockerfile
ENV ZENDESK_API_ENDPOINT 'https://your-sandbox-domain.zendesk.com/api/v2/'
ENV ZENDESK_API_USER 'your-sandbox-user@your-sandbox-domain.co.uk'

# you can generate this token from your account in Zendesk Admin
ENV ZENDESK_API_TOKEN 'abcdefghihgfedcba' 
```

Things still to do:
- [x] Identify exactly what metadata (tags? groups?) the ZD team needs for us
- [x] Change the specs for the rest of the help pages to test against ZD tickets instead of notify
- [x] Remove the 'notify' code path entirely for the help pages
- [x] Some alternate code path for local dev - maybe just spitting the ticket details at a logger?
- [ ] Switch from testing through overriding a constant (Zendesk::Client) to testing at the boundary (using webmock)


app/controllers/help_controller.rb
```diff
@@ -40,19 +40,33 @@ def create
...
+          tickets_gateway: Gateways::ZendeskSupportTickets.new
+        ).execute(
+          requester: {
+            email: sender_email,
+            name: params[:support_form][:name] || current_user&.name,
+            organisation: sender_organisation_name
+          },
+          details: params[:support_form][:details]
+        )
+      else
```
kjdchapman: This else branch will lose all the notify code once the specs have been sorted out.

kjdchapman:
Alright, from my perspective, this is good to go. This version happily posts into the sandbox environment from local, with the right tags and contact info on:

image

I've tested all the individual forms (with the different combinations of optional inputs).

Things not included: 
- Reworking the testing (this'll take some time)
- Sinking the 'organisation' field into the body of the support ticket (this isn't a big job but time is tight)

lib/gateways/zendesk_support_tickets.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize
+    @client = ZendeskAPI::Client.new do |config|
+      config.url = ENV.fetch('ZENDESK_API_ENDPOINT')
+      config.username = ENV.fetch('ZENDESK_API_USER')
+      config.token = ENV.fetch('ZENDESK_API_TOKEN')
+    end
+  end
+
+  def create(subject:, email:, name:, body:)
+    name = 'Unknown' if name.blank?
```
emileswarts: We allow this in the form?

app/controllers/help_controller.rb
```diff
@@ -41,19 +41,23 @@ def create
     end
 
     if @support_form.valid?
-      template_id = GOV_NOTIFY_CONFIG['help_email']['template_id']
+      if ENV['ZENDESK_API_ENDPOINT'].present?
```
emileswarts: Bit worried that this could mean that a miss-configuration go undetected, not sure we will look in the logs for this message.

If we remove this conditional will it blow up?

spec/use_cases/administrator/create_support_ticket_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  subject { described_class.new(tickets_gateway: tickets_gateway_spy) }
+
+  let(:tickets_gateway_spy) { spy(:create) }
+
+  before do
+    subject.execute(
+      requester: {
+        email: 'help@support.com',
+        name: 'Helpy McHelpface',
+        organisation: 'Parks & Rec'
```
emileswarts: We don't use organisation in the request?

spec/features/support/mock_zendesk_client.rb
```diff
@@ -0,0 +1,47 @@
+shared_context 'with a mocked support tickets client' do
+  class ZendeskClientMock
```
emileswarts: More like a spy this?  Interesting though, I like it

lib/gateways/zendesk_support_tickets.rb
kjdchapman: We do - some of the forms don't even have it.

It's mandatory for the Zendesk API though, if we want to fill in an e-mail, so have to fill it in with something.

app/controllers/help_controller.rb
kjdchapman: Chatted about this briefly in real life - this conditional's basically for local dev. 

Misconfiguration will make it blow up later down the stack, where we're using `fetch` etc.

spec/features/support/mock_zendesk_client.rb
kjdchapman: After talking with @craigjbass & @danielburnley yesterday I actually don't like it! Can explain more in-person, think I had a level-up moment on how to do this stuff yesterday.

Does the job for now though, was really helpful when growing the thing.

spec/use_cases/administrator/create_support_ticket_spec.rb
kjdchapman: We don't. My intent is to add to the body of the support request, but I skipped it for now as I'm worried about time. 

Happy to work on adding it in now if you think it'd be a big miss - haven't seen a support ticket yet where it's been essential to providing a good response.

PR #445, opened 2019-02-06
Add custom organisations 

georgeschena:
Allows custom organisations to be added, as before you could only choose organisations from the register provided. However now we have the ability to append other organisations to these registers to allow us to add other organisations that may not be on the official register yet.

### DIFF
--- a/spec/use_cases/organisation/fetch_organisations_register_spec.rb
+++ b/spec/use_cases/organisation/fetch_organisations_register_spec.rb
@@ -1,5 +1,5 @@
 describe UseCases::Organisation::FetchOrganisationRegister do
-  let(:gateway_spy) { spy(government_orgs: [], local_authorities: []) }
+  let(:gateway_spy) { spy(government_orgs: [], local_authorities: [], custom_orgs: []) }
 
   it 'calls government_orgs on the gateway' do
     described_class.new(organisations_gateway: gateway_spy).execute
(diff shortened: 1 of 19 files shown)


spec/use_cases/organisation/fetch_organisations_register_spec.rb
```diff
@@ -1,13 +1,45 @@
...
+      expect(gateway_spy).to have_received(:local_authorities)
+    end
+
+    it 'calls custom_orgs' do
+      described_class.new(organisations_gateway: gateway_spy).execute
+      expect(gateway_spy).to have_received(:custom_orgs)
+    end
+  end
+
+  context 'get the custom orgs' do
```
tommotaylor: "with custom organisations" is better england

spec/use_cases/organisation/fetch_organisations_register_spec.rb
```diff
@@ -1,13 +1,45 @@
...
+    it 'returns the custom orgs' do
+      response = described_class.new(organisations_gateway: custom_orgs_list_gateway).execute
+      expect(response).to eq(['Custom Org 1', 'Custom Org 2', 'Custom Org 3'])
+    end
   end
 
-  it 'calls local_authorities on the gateway' do
-    described_class.new(organisations_gateway: gateway_spy).execute
-    expect(gateway_spy).to have_received(:local_authorities)
+  context 'with gov, local and custom orgs' do
```
tommotaylor: same again, may as well use full words if they fit

PR #447, opened 2019-02-07
UI for adding custom orgs

georgeschena:
**This PR provides the user with a UI to add custom organisations to the combined registers list, so that the organisation added can then sign up to GovWifi through the normal signup journey.**

screenshot 2019-02-08 at 10 00 03

### DIFF
--- a/app/controllers/admin/custom_organisations_controller.rb
+++ b/app/controllers/admin/custom_organisations_controller.rb
@@ -1,15 +1,26 @@
 class Admin::CustomOrganisationsController < AdminController
-
-  def index; end
+  def index
+    all_custom_organisation_names
+  end
 
   def create
-    CustomOrganisationName.create(custom_org_params)
+    all_custom_organisation_names
+    created_org = CustomOrganisationName.create(custom_org_params)
 
-    flash[:notice] = 'Successfully added a custom organisation'
-    redirect_to root_path
+    if created_org.valid?
+      flash[:notice] = 'Successfully added a custom organisation'
+      redirect_to root_path
+    else
+      flash[:alert] = "Organisation name can't be blank"
+      render :index
+    end
   end
 
 private
+
+  def all_custom_organisation_names
+    @all_orgs = CustomOrganisationName.all
+  end
 
   def custom_org_params
     params.require(:custom_organisations).permit(:name)
--- a/app/controllers/admin/custom_organisations_controller.rb
+++ b/app/controllers/admin/custom_organisations_controller.rb
@@ -1,14 +1,17 @@
 class Admin::CustomOrganisationsController < AdminController
 
-  def index
-    @number_of_custom_orgs = CustomOrganisationName.count
-  end
+  def index; end
 
   def create
-    # Add whatever is in that text field into the database
-    add_custom_organiastion = CustomOrganisationName.create(params[:name])
+    CustomOrganisationName.create(custom_org_params)
 
     flash[:notice] = 'Successfully added a custom organisation'
     redirect_to root_path
+  end
+
+private
+
+  def custom_org_params
+    params.require(:custom_organisations).permit(:name)
   end
 end
--- a/app/controllers/admin/custom_organisations_controller.rb
+++ b/app/controllers/admin/custom_organisations_controller.rb
@@ -1,26 +1,20 @@
 class Admin::CustomOrganisationsController < AdminController
   def index
-    all_custom_organisation_names
+    @all_orgs = CustomOrganisationName.all
+    @custom_organisation = CustomOrganisationName.new
   end
 
   def create
-    all_custom_organisation_names
-    created_org = CustomOrganisationName.create(custom_org_params)
+    @all_orgs = CustomOrganisationName.all
+    @custom_organisation = CustomOrganisationName.new(custom_org_params)
 
-    if created_org.valid?
-      flash[:notice] = 'Successfully added a custom organisation'
-      redirect_to root_path
-    else
-      flash[:alert] = "Organisation name can't be blank"
-      render :index
+    if @custom_organisation.save
+      flash[:notice] = "Custom organisation has been successfully added"
     end
+    render :index
   end
 
 private
-
-  def all_custom_organisation_names
-    @all_orgs = CustomOrganisationName.all
-  end
 
   def custom_org_params
     params.require(:custom_organisations).permit(:name)
--- a/app/controllers/admin/custom_organisations_controller.rb
+++ b/app/controllers/admin/custom_organisations_controller.rb
@@ -1,0 +1,14 @@
+class Admin::CustomOrganisationsController < AdminController
+
+  def index
+    @number_of_custom_orgs = CustomOrganisationName.count
+  end
+
+  def create
+    # Add whatever is in that text field into the database
+    add_custom_organiastion = CustomOrganisationName.create(params[:name])
+
+    flash[:notice] = 'Successfully added a custom organisation'
+    redirect_to root_path
+  end
+end
--- a/app/controllers/admin/custom_organisations_controller.rb
+++ b/app/controllers/admin/custom_organisations_controller.rb
@@ -1,11 +1,11 @@
 class Admin::CustomOrganisationsController < AdminController
   def index
-    @all_orgs = CustomOrganisationName.all
+    @custom_organisation_names = CustomOrganisationName.all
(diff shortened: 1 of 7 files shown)


app/views/admin/custom_organisations/index.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-l">Add Organistions to the Register</h2>
```
kjdchapman: I think this is misleading - I'd read "Register" as what we're getting from Registers, and we're not adding it back to them.

As I tried to write a simpler sentence, it made me think - this is a pretty convoluted process:
- someone tries to sign up with `Custom Organisation Inc.`
- we get an e-mail
- we add `Custom Organisation Inc.` (and have to spell it exactly right!)
- we go back to the user and tell them to sign up again

I understand why we've ended up with this to start but it makes explaining it in simple terms a pain.

Maybe something like
> ### Allow an organisation access to the admin platform
> Enter the organisation's full name:
> [Text box]
> [Button] Allow organisation

app/views/application/_subnavigation.html.erb
```diff
@@ -11,6 +11,7 @@
     <nav class='govuk-body govuk-!-margin-bottom-0 '>
       <% if current_user.super_admin? %>
         <%= link_to 'Organisations', admin_organisations_path, class: active_tab(admin_organisations_path) %>
+        <%= link_to 'Custom Organisations', admin_custom_organisations_path, class: active_tab(admin_custom_organisations_path) %>
```
kjdchapman: `Custom Organisations` feels OK here, definitely think we're implying that these screens will _create_ the organisation rather than just _allow_ people to create them, though.

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      sign_in_user admin_user
+      visit admin_custom_organisations_path
+    end
+
+    it 'will show the add custom organisations page' do
+      expect(page.body).to have_content('Allow an organisation access to the admin platform')
+    end
+
+    it 'will redirect you to organiastions page once a custom org is added' do
+      fill_in 'custom_organisations[name]', with: 'Custom Org name'
```
emileswarts: You can target the label to make specs more readable

"Enter the organisation's full name"

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+    end
+
+    context 'if the organisation name is blank' do
+      it 'will error with the reason' do
+        fill_in 'custom_organisations[name]', with: ' '
+        click_on 'Allow organisation'
+        expect(page).to have_content("Name can't be blank")
+      end
+    end
```
emileswarts: would ensure blank lines around these

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    context 'after addding a custom organisation' do
+      let!(:org1) { CustomOrganisationName.create(name: 'Custom Org 1') }
+      let!(:org2) { CustomOrganisationName.create(name: 'Custom Org 2') }
```
emileswarts: Can we turn this into a factory with FactoryBot?

app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::CustomOrganisationsController < AdminController
+  def index
+    @all_orgs = CustomOrganisationName.all
```
emileswarts: I know this is within the context of the custom organisation controller but think improving this variable name to something like "custom_organisation_names" could be helpful?

spec/features/admin/add_custom_organisation_spec.rb
georgeschena: I feel like as this feature is pretty small I don't really see the huge use of spending the time to make a Factory bot for it imo, but I see ur point

georgeschena: With he other Factory bots we use it across multiple test files while here its only being used here, id say defer for now

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      visit admin_custom_organisations_path
+    end
+
+    it 'will show the add custom organisations page' do
+      expect(page.body).to have_content('Allow an organisation access to the admin platform')
+    end
+
+    it 'will redirect you to organiastions page once a custom org is added' do
+      fill_in "Enter the organisation's full name", with: 'Custom Org name'
+      click_on 'Allow organisation'
```
tommotaylor: I would check that the `CustomOrganisationName` has actually been created here. You would do that like so:
```
expect do
  click_on 'Allow organisation'
end.to change {CustomOrganisationName.count}.by(1)
```
There is some debate about checking the database like this from feature specs, but as this is the only spec checking the create action (theres no controller or request specs, which makes sense), then I think we should cover it here.

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(page).to have_content('Custom organisation has been successfully added')
+    end
+
+    context 'sign out and find custom organisation' do
+      before do
+        fill_in "Enter the organisation's full name", with: 'Custom Org name'
+        click_on 'Allow organisation'
+        sign_out
```
tommotaylor: These lines 26, 27 and 28 are already covered above, so feel like duplication. You could just setup the test here with:
```
before do
        CustomOrganisationName.create(name: 'Custom Org name')
        sign_up_for_account(email: 'default@gov.uk')
        visit confirmation_email_link
end
```

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        sign_up_for_account(email: 'default@gov.uk')
+        visit confirmation_email_link
+      end
+
+      it 'the custom organisation will appear when creating an account' do
```
tommotaylor: `it 'the` doesn't make sense. Specs need to be readable. Something like `it 'displays the custom organisation name in the list` would be better for instance.

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    before do
+      sign_in_user admin_user
+      visit admin_custom_organisations_path
+    end
+
+    it 'will show the add custom organisations page' do
+      expect(page.body).to have_content('Allow an organisation access to the admin platform')
+    end
+
+    it 'will redirect you to organiastions page once a custom org is added' do
```
tommotaylor: spelling "organiastions". Probably worth doing a search on codebase for this spelling mistake.

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    context 'sign out and find custom organisations' do
+      before do
+        sign_out
+        CustomOrganisationName.create(name: 'Custom Org name')
+        sign_up_for_account(email: 'default@gov.uk')
+        visit confirmation_email_link
+      end
+
+      it 'displays the custom organisation name in the list` would be better for instance' do
```
tommotaylor: Haha George. You blindly copied and pasted the entire line

spec/features/admin/add_custom_organisation_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      fill_in "Enter the organisation's full name", with: 'Custom Org name'
+      expect do
+        click_on 'Allow organisation'
+        .to change { CustomOrganisationName.count }.by(1)
+      end
+    end
+
+    context 'sign out and find custom organisations' do
+      before do
+        sign_out
```
tommotaylor: The reason you have to sign out here is because this context is nested in the one above. I don't think you need it nested. You are currently running lines 9 - 12 before this context which doesn't need it.

spec/features/admin/add_custom_organisation_spec.rb
georgeschena: 👁

georgeschena: LOOOOL

PR #448, opened 2019-02-07
Activity stat for super-admin view

kjdchapman:
Not sure if the stat is better here in the panel, or as a sentence underneath. Could argue for different time periods to 24 hours, too.

image

I had a quick play with getting this on the list page too - definitely doable, but more work, so not pursuing it for now.


lib/gateways/unique_connections.rb
```diff
@@ -5,9 +5,14 @@ def initialize(ips:)
...
 
     def unique_user_count(date_range: nil)
-      return Session.where(siteIP: organisation_ips).where('start >= ?', date_range).where(success: true).distinct.count(:username) if date_range.present?
+      query = Session
+        .where(success: true)
+        .distinct
+        .where(siteIP: organisation_ips)
 
-      Session.where(siteIP: organisation_ips).where(success: true).distinct.count(:username)
+      query = query.where('start >= ?', date_range) if date_range.present?
```
emileswarts: When do we use a different date range here?
Realise you didn't introduce this functionality but wondering where it is used.

kjdchapman: We don't, as far as I know - I guess it was built prioritising flexibility over YAGNI?

I would have removed it as part of this PR, but I thought part of the PR feedback might have been "we should really do this for a week instead of 24 hours"!

I'd be happy to remove it, as I haven't gotten that feedback.

emileswarts: Interesting, I'd be tempted to remove it.  Less code to think about

emileswarts:
Loving the "super-admin" name refactors

spec/features/super-admin/view_organisation_page_spec.rb
```diff
@@ -75,6 +60,27 @@
...
+            start: Time.now.to_s
+          )
+        end
+      end
+
+      it 'shows five unique connections' do
+        visit admin_organisation_path(organisation)
+
+        within('#unique-connections') do
+          expect(page).to have_content('5')
```
emileswarts: Think this could easily be a false positive if there was a 5 anywhere on the page.
Would consider targeting a specific piece of markup with "within", or refining the match string.

kjdchapman:
@emileswarts I removed the custom date range stuff, think this was a good idea. Mind scanning this again quick before I merge?

PR #457, opened 2019-02-11
Change text size of subnav elements 

georgeschena:
screenshot 2019-02-11 at 16 26 05

### DIFF
--- a/app/assets/stylesheets/components/sub_nav.scss
+++ b/app/assets/stylesheets/components/sub_nav.scss
@@ -20,6 +20,7 @@
       display: inline-block;
       text-decoration: none;
       border-bottom: 5px solid white;
+      font-size: 16px;
     }
 
     .active {


kjdchapman:
Could you include a screenshot of the change? 

It's a good habit even when we've discussed the change in-person - helps out people in the future doing code archaeology.

app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -20,6 +20,7 @@
       display: inline-block;
       text-decoration: none;
       border-bottom: 5px solid white;
+      font-size: 16px;
```
kjdchapman: I'd use a font size override class (`govuk-!-font-size-16`) rather than this custom CSS. Helps call out that we're doing something weird, easier to change in the future, etc.

https://design-system.service.gov.uk/styles/typography

georgeschena: I feel this just clutters up the HTML, as well as I thought we try and avoid using this type of styling, as we use it as a last resort when styling? @kjdchapman

kjdchapman: Had a chat in-person about this. Think you're right that since there's already a custom class (with a bunch of padding/margins) here it's better to be consistent.

PR #462, opened 2019-02-12
Make postcodes mandatory and valid

georgeschena:


app/views/locations/new.erb
```diff
@@ -37,7 +42,11 @@
 
       <div class="govuk-form-group">
         <%= f.label :postcode, class: 'govuk-label' %>
-        <%= f.text_field :postcode, class: 'govuk-input govuk-input--width-10' %>
+        <% if @location.errors.keys.include?(:postcode) %>
+          <%= f.text_field :postcode, class: 'govuk-input govuk-input--error' %>
+        <% else %>
+          <%= f.text_field :postcode, class: 'govuk-input govuk-input--width-10' %>
```
kjdchapman: Does this width class need to be on line 46 as well?

app/models/location.rb
```diff
@@ -7,15 +9,23 @@ class Location < ApplicationRecord
   validates_associated :ips
 
   validates :address, presence: true
+  validates :postcode, presence: true
```
emileswarts: Think you can squash the above two into a single line

validates :postcode, :address, presence: true

app/models/location.rb
```diff
@@ -7,15 +9,23 @@ class Location < ApplicationRecord
...
 
   def full_address
-    postcode.blank? ? address : "#{address}, #{postcode}"
+    "#{address}, #{postcode}"
   end
 
 private
 
+  def validate_postcode
+    if postcode.nil? || !UKPostcode.parse(postcode).valid?
```
emileswarts: Do we need a nil check?
Won't nil be invalid according to this gem?

spec/models/location_spec.rb
```diff
@@ -34,13 +35,19 @@
...
       end
+    end
+  end
 
-      context 'but a blank postcode' do
-        let(:postcode) { '' }
+  describe 'invalid postcode entry' do
+    context 'with an invalid postcode' do
+      it "errors if the postcode is not valid" do
+        subject.postcode = "WHATEVER POSTCODE"
```
emileswarts: I'd stick to single quotes for non interpolated strings for consistency

app/models/location.rb
georgeschena: In the Location model tests it moans about postcode being nil as on create I assume it doesn't  create a postcode from what I'm aware of? @emileswarts

emileswarts: So can the first part of this conditional be removed? Not 100% clear on this.

georgeschena: It can't

emileswarts:
```ruby
unless UKPostcode.parse(postcode.to_s).valid?
     errors.add(:postcode, "must be valid")
end
```

app/models/location.rb
```diff
@@ -6,16 +8,23 @@ class Location < ApplicationRecord
...
+      errors.add(:postcode, "must be valid")
```
kjdchapman: nit: single quotes
```suggestion
      errors.add(:postcode, 'must be valid')
```

app/models/location.rb
georgeschena: I LIKE

spec/features/locations/add_new_location_spec.rb
```diff
@@ -81,13 +81,37 @@
...
+
+      context 'the postcode is blank' do
+        before do
+          fill_in 'Address', with: '30 Square'
+          fill_in 'Postcode', with: ''
+          fill_in ip_input, with: '10.0.0.3'
+          click_on 'Add new location'
+        end
+
+        it_behaves_like "errors in form"
```
kjdchapman: nit: quote marks
```suggestion
        it_behaves_like 'errors in form'
```

spec/models/location_spec.rb
```diff
@@ -34,13 +35,24 @@
           expect(subject.full_address).to eq('121 Fictional Street, FI5 S67')
         end
       end
+    end
+  end
+
+  describe 'Entering a postcode' do
+    context 'which is in the incorrect format' do
+      it 'errors as the postcode dosent match the correct format' do
```
kjdchapman: nit: spelling
```suggestion
      it 'errors as the postcode does not match the correct format' do
```

PR #463, opened 2019-02-12
[WIP]: Fix secondary navigation view on mobile

Renny4Real:
**WHY:**
The view of the secondary navigation on mobile needed to be improved upon. Its contents were not arranged properly.
This PR is tagged [WIP] for now as it's more of a design approach rather than an immediate solution.

**IN THIS PR:**
- The links in the sub navigation were changed to become list items. (The idea came from inspecting some websites that feature a drop down menu and finding out that these menu items are list items in HTML).
- The list items were made to be arranged in-line only for large screens (which replicates how it looks now.)

screenshot 2019-02-12 at 16 45 51

**_How it appears when viewed on a large screen e.g. desktop_**

- In smaller screens, the list items were arranged as a block (which is the default display for a list).

**BEFORE:**

screenshot 2019-02-12 at 16 45 03

**_How it appears when viewed on a mobile screen._**

**AFTER:**

screenshot 2019-02-12 at 16 24 54

**_How it could appear when viewed on a mobile screen._**

**Suggestions/Feedback on this approach would be appreciated**

### DIFF
--- a/app/assets/stylesheets/components/sub_nav.scss
+++ b/app/assets/stylesheets/components/sub_nav.scss
@@ -19,8 +19,9 @@
       padding:0;
 
       li {
-        display:inline-block;
-        // float: left;
+        @media only screen and (min-width: 600px) {
+          display: inline-block;
+        }
         margin-right: 1px;
       }
     }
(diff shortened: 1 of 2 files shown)


tommotaylor:
Closed in favour of https://github.com/alphagov/govwifi-admin/pull/477

PR #465, opened 2019-02-13
add the database as a dependency for the app

cybojenix:
this will ensure the database is launched when running the app directly

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
   app:
     build: .
     links:
+      - db
       - rr_db
     ports:
     - "8080:3000"


kjdchapman:
What does this fix?

cybojenix:
This directly fixes `make shell`, which relies on automatically loading services, as apposed to `make serve`, which manually launches all the services

kjdchapman:
As far as I can see, it doesn't fix it for me. I can provide more information if that helps diagnose? Or we could chat?

kjdchapman:
(It looks like it brought up the other containers correctly, but didn't write to their dbs:
```
# Mysql2::Error:
#   Unknown database 'govwifi_admin_test'
```

cybojenix:
ah, ok I'll prepare a change specific for that.
The `serve` and `test` commands both populate the database at launch time. Will prepare a fix.

While this PR doesn't fix that issue, it's still best practice to associate an app with its dependent services

PR #469, opened 2019-02-14
Various quick wins for docker

cybojenix:
### Ignore generated files for the docker context
This avoids stale cache making it in, as well as `node_modules`, which can cause complications if compiled packages are used.
We also avoid putting implicit hidden files going in, such as the git directory.

### Spin up services at the same time, instead of one by one
Docker can do this in parallel, so let's do it in parallel.

### Use volumes instead of bind mounts
Bind mounts are expensive, and we don't need them. Use volumes for cache, storage, and node modules
**note**, storage will now be discarded. Given the database is discarded, I assume this doesn't matter for our usecase.

### Avoid precompiling the assets at build time
This takes time, so defer it for when they're actually required.
**note** they'll still precompile on Jenkins

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,3 +16,4 @@ node_modules
 public/assets
 storage
 tmp
+log
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,18 @@
 coverage
+
+# Ignore all hidden files
+.*
+**/.*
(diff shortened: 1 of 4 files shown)


tommotaylor:
Looks really nice @cybojenix, much faster for me on the second run.

PR #471, opened 2019-02-14
Recommend adding to many RADIUS servers

kjdchapman:
Couldn't find a way to make this clear without changing the structure a bit, so that's what I did.

image

Thanks to Pippa for helping with a first round of feedback - happy to hear more though!

### DIFF
--- a/app/views/application/_radius_details.html.erb
+++ b/app/views/application/_radius_details.html.erb
@@ -5,10 +5,14 @@
     </span>
   </summary>
   <div class="govuk-details__text">
+    <p class="govuk-body">
+      We recommend using all of our RADIUS servers, so that you'll be protected from a single server being unavailable. If you can't use them all, try to use at least one from each region.
+    </p>
+
     <% if current_organisation.locations.empty? %>
       <p>Your RADIUS secret keys will be generated when you <%= link_to 'add your first IP address', new_ip_path, class: "govuk-link" %></p>
     <% else %>
-      <p> Your RADIUS secret keys are shown under <%= link_to "your organisation's locations", ips_path, class: "govuk-link" %></p>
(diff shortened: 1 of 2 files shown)


app/views/application/_radius_details.html.erb
```diff
@@ -8,9 +8,9 @@
     <% if current_organisation.locations.empty? %>
       <p>Your RADIUS secret keys will be generated when you <%= link_to 'add your first IP address', new_ip_path, class: "govuk-link" %></p>
     <% else %>
-    <p> Your RADIUS secret keys are shown under <%= link_to "your organisation's locations", ips_path, class: "govuk-link" %></p>
+      <p> Your RADIUS secret keys are shown under <%= link_to "your organisation's locations", ips_path, class: "govuk-link" %></p>
     <% end %>
-    <p> We recommend connecting to one RADIUS server from each region: </p>
```
emileswarts: Should we be explicit about the fact that you have to use ALL of the ips for failover?

kjdchapman: Er, I think we are, in this PR:
> We recommend using all of our RADIUS servers, so that you'll be protected from a single server being unavailable.

I wonder if the OLD/NEW screenshots weren't particularly clear!

emileswarts: oh right I see it's on the other page.
We don't want to change this message that you removed to encourage people as they look at the list?

kjdchapman: It's within this box - Pippa recommended having it as the first thing people see within that section:

image

kjdchapman:
I missed pushing one of the changes to this, my bad. The screenshot didn't line up with the actual content, which was confusing!

PR #472, opened 2019-02-14
Delete custom organisations

georgeschena:


app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -9,11 +9,17 @@ def create
...
 
     if @custom_organisation.save
-      flash[:notice] = "Custom organisation has been successfully added"
+      flash.now[:notice] = "Custom organisation has been successfully added"
     end
     render :index
   end
 
+  def destroy
+    custom_org = CustomOrganisationName.find_by(params.fetch(:id))
```
tommotaylor: This won't work. There is `.find` and `.find_by`. `.find_by` requires a key, `.find` only takes an ID: https://api.rubyonrails.org/v5.2.2/classes/ActiveRecord/FinderMethods.html#method-i-find_by

app/views/admin/custom_organisations/index.html.erb
```diff
@@ -16,19 +17,20 @@
       <% end %>
 
     <table class="govuk-table">
-    <h2 class="govuk-heading-m">Custom Organistions</h2>
+    <h2 class="govuk-heading-m">Custom Organisations that are already in our register</h2>
```
tommotaylor: Nice

app/views/admin/custom_organisations/index.html.erb
```diff
@@ -16,19 +17,20 @@
...
           <th class="govuk-table__header" scope="col">Name</th>
           <th class="govuk-table__header" scope="col">Allowed on</th>
+          <th class="govuk-table__header" scope="col"> </th>
         </tr>
       </thead>
       <tbody class="govuk-table__body">
         <% @custom_organisation_names.each do |organisation| %>
-        <tbody class="govuk-table__body">
-          <tr class="govuk-table__row">
+          <tr class="govuk-table__row" id="data-id-<%= organisation.id %>">
```
tommotaylor: Sorry I think I led you astray here. The standard approach is `data-id="7"` not `id="data-id-7"`

You may as well just use `id=7` to make it easy.

spec/features/admin/delete_custom_organosation_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  context 'when visiting the custom organisations page' do
+    before do
+      sign_in_user admin_user
+      visit admin_custom_organisations_path
+      CustomOrganisationName.create(name: 'DummyOrg1')
+      CustomOrganisationName.create(name: 'DummyOrg2')
+    end
+
+    it 'allows the user to delete a custom organisation' do
+      within("tr#data-id-1") do
```
tommotaylor: You need to do two things here. You need to match the selector with the format of your `id` / `data-id` and you need to pass in the `id` of the `CustomOrganisationName` you are targeting.

I would do something like if you were using `data-id=7`:
```
let!(:custom_org_name) { CustomOrganisationName.create(name: 'DummyOrg1') }

within('tr[data-id="#{custom_org_name.id}"]') do
end
```
or if you are just doing `id=7` the following may work
```
html_selector = "tr#" + custom_org_name.id
within(html_selector) do
end
```

spec/features/admin/delete_custom_organosation_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  context 'changes the number of custom orgs in the register' do
+    before do
+      CustomOrganisationName.create(name: 'DummyOrg2')
+      sign_in_user admin_user
+      visit admin_custom_organisations_path
+    end
+
+    it 'will remove the custom org from the register' do
+      click_link 'Remove'
```
tommotaylor: Again you need to make sure this is the right remove button as there are many on the page.

app/models/custom_organisation_name.rb
```diff
@@ -1,3 +1,16 @@
...
+  def validate_in_register?
+    unless Organisation.fetch_organisations_from_register.exclude?(name.strip)
+      errors.add(:name, "is already in our register")
+    end
+  end
+
+  def self.fetch_organisations_from_register
+    UseCases::Organisation::FetchOrganisationRegister.new(
+      organisations_gateway: Gateways::GovukOrganisationsRegisterGateway.new
+    ).execute.sort
```
emileswarts: Where is this being used?

spec/features/super_admin/add_custom_organisation_spec.rb
```diff
@@ -23,11 +23,18 @@
...
         fill_in 'custom_organisations[name]', with: ' '
         click_on 'Allow organisation'
         expect(page).to have_content("Name can't be blank")
       end
+
+      it 'will error if its already in our register' do
+        CustomOrganisationName.create(name: 'Custom Org name')
+        fill_in 'custom_organisations[name]', with: 'Custom Org name'
+        click_on 'Allow organisation'
+        expect(page).to have_content("Name is already in our register")
```
emileswarts: Are we referring to this as our register?

Maybe it can just say "This organisation already exists" ?

spec/features/admin/delete_custom_organisation_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'support/confirmation_use_case'
+
+describe 'deleting a custom organisation name' do
+  include_examples 'confirmation use case spy'
+  let!(:custom_org_name1) { CustomOrganisationName.create(name: 'DummyOrg1') }
+  let!(:custom_org_name2) { CustomOrganisationName.create(name: 'DummyOrg2') }
+  let!(:custom_org_name3) { CustomOrganisationName.create(name: 'DummyOrg3') }
```
emileswarts: This no factory model management is spreading

spec/features/admin/delete_custom_organisation_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'support/confirmation_use_case'
+
+describe 'deleting a custom organisation name' do
+  include_examples 'confirmation use case spy'
```
emileswarts: We need this included example?

app/models/custom_organisation_name.rb
```diff
@@ -1,3 +1,16 @@
 class CustomOrganisationName < ApplicationRecord
-  validates :name, presence: true, uniqueness: { case_sensitive: false }
+  validate :validate_in_register?
```
emileswarts: Wonder if it's redundant to have the validate method have "validate" in its name.

spec/features/super_admin/add_custom_organisation_spec.rb
georgeschena: Yeah to our register, I feel it gives Steve move context by saying "Name is already in our register"

app/models/custom_organisation_name.rb
```diff
@@ -1,3 +1,16 @@
...
+  validates :name, presence: true
+
+  def validate_in_register?
+    unless Organisation.fetch_organisations_from_register.exclude?(name.strip)
```
emileswarts: unless -> exclude could probably be converted to if -> include and be more readable?

spec/features/admin/delete_custom_organisation_spec.rb
georgeschena: Last time it will be used 👁

app/models/custom_organisation_name.rb
georgeschena: Change the method name to 'custom_org_in_register?'

app/models/custom_organisation_name.rb
georgeschena: Im down with this

app/views/admin/custom_organisations/index.html.erb
```diff
@@ -16,19 +17,20 @@
...
         </tr>
       </thead>
       <tbody class="govuk-table__body">
         <% @custom_organisation_names.each do |organisation| %>
-        <tbody class="govuk-table__body">
-          <tr class="govuk-table__row">
+          <tr class="govuk-table__row" id="<%= organisation.id %>">
             <td class="govuk-table__cell"><%= organisation.name %></td>
             <td class="govuk-table__cell"><%= organisation.created_at.strftime('%e %b %Y') %></td>
+            <td class="govuk-table__cell"><%= link_to 'Remove', admin_custom_organisations_path(id: organisation.id, remove_custom_org: true), class: 'govuk-link red-link' %></td>
```
emileswarts: Looks like we are setting an html id twice with the value of the organisation id.
Ids need to be unique

spec/features/admin/delete_custom_organisation_spec.rb
georgeschena: Nope, works without

spec/features/admin/delete_custom_organisation_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    before do
+      sign_in_user admin_user
+      visit admin_custom_organisations_path
+    end
+
+    it 'allows the user to delete a custom organisation' do
+      html_selector = "tr#" + custom_org_name1.id.to_s
+      within(html_selector) do
+        click_link 'Remove'
+      end
```
emileswarts: Haven't tested this, but would this work?

page.find('#organisation-#{custom_org_name1.id}'').click

Not sure we need to add ids to so many elements on that page

georgeschena: Its so we can click the right remove button when deleting an organisation, so we need to make sure in the test that we delete the correct custom org

app/models/custom_organisation_name.rb
georgeschena: Removed it, forgot to get rid of it

app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -9,11 +9,17 @@ def create
...
     if @custom_organisation.save
-      flash[:notice] = "Custom organisation has been successfully added"
+      flash.now[:notice] = "Custom organisation has been successfully added"
     end
     render :index
   end
 
+  def destroy
+    custom_org = CustomOrganisationName.find(params.fetch(:id))
+    custom_org.destroy
```
emileswarts: Not vital because it's just the super admins who see this, but the convention is to only redirect if it was successful:

```
if custom_org.destroy
    redirect_to admin_custom_organisations_path, notice: "Successfully removed #{custom_org.name}"
end
```

Will prevent confusing where it says it deleted the org but that didn't really happen.

app/models/custom_organisation_name.rb
```diff
@@ -1,3 +1,10 @@
 class CustomOrganisationName < ApplicationRecord
-  validates :name, presence: true, uniqueness: { case_sensitive: false }
+  validate :custom_org_in_register?
+  validates :name, presence: true
```
emileswarts: Intentional to remove the uniqueness validation?

georgeschena: Yeah as the 'custom_org_in_register?' does this for us

spec/features/super_admin/add_custom_organisation_spec.rb
emileswarts: I mean just the naming here in general.
If the error message refers to it as our register, why doesn't the code do the same?

PR #477, opened 2019-02-18
Add responsive breakpoints to subnavigation

tommotaylor:
- Adds the gov.uk design system breakpoints to the subnav.
- On mobile the subnav and org name go to a new line and are centred.
- On desktop the subnav is inline and right aligned.

Mobile:
screenshot 2019-02-18 at 12 09 55

Tablet & Desktop:
screenshot 2019-02-18 at 11 33 07

screenshot 2019-02-18 at 12 10 25

### DIFF
--- a/app/assets/stylesheets/components/sub_nav.scss
+++ b/app/assets/stylesheets/components/sub_nav.scss
@@ -43,15 +43,11 @@
       display: inline-block;
       text-decoration: none;
       font-size: 16px;
-    }
 
-    .active {
-      border-bottom: 5px solid black;
-      margin-bottom: 0px !important;
+      &.active {
+        border-bottom: 5px solid black;
+        margin-bottom: 0px;
+      }
     }
   }
-}
-
-.text-decoration-none {
-  text-decoration: none;
 }
--- a/app/assets/stylesheets/components/sub_nav.scss
+++ b/app/assets/stylesheets/components/sub_nav.scss
@@ -1,5 +1,4 @@
 .subnav {
-  display: flex;
   margin-left: 0px;
   margin-right: 0px;
   border-bottom: 1px solid black;
@@ -10,21 +9,36 @@
     padding-bottom: 10px;
   }
 
+  .nav-links {
+    display: flex;
+    justify-content: space-around;
+
+    @media (min-width: 640px) {
+      justify-content: flex-end;
+    }
+  }
+
   nav {
     text-align: right;
+    margin-top: 4px;
+
+    > :first-child {
+      margin-left: 0px;
+    }
 
     a {
       margin-left: 15px;
+      margin-bottom: 5px;
       padding-top: 10px;
       padding-bottom: 5px;
       display: inline-block;
       text-decoration: none;
-      border-bottom: 5px solid white;
       font-size: 16px;
     }
 
     .active {
       border-bottom: 5px solid black;
+      margin-bottom: 0px !important;
     }
   }
 }
--- a/app/assets/stylesheets/components/sub_nav.scss
+++ b/app/assets/stylesheets/components/sub_nav.scss
@@ -33,6 +33,8 @@
 
   nav {
     text-align: right;
+    // This margin is required to keep the left and right columns the same height.
+    // Flexbox cannot be used on mobile as it breaks the design system styles.
     margin-top: 4px;
 
     // Margin that separates subnav links is not required on first link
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -10,21 +9,36 @@
...
+    }
+  }
+
   nav {
     text-align: right;
+    margin-top: 4px;
+
+    > :first-child {
+      margin-left: 0px;
+    }
```
tommotaylor: This ensures that the padding-left that spaces out the subnav links only applies between the links, not on the left of the first link.

app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -10,21 +9,36 @@
...
 
     a {
       margin-left: 15px;
+      margin-bottom: 5px;
       padding-top: 10px;
       padding-bottom: 5px;
       display: inline-block;
       text-decoration: none;
-      border-bottom: 5px solid white;
```
tommotaylor: Replaced this with `margin-bottom: 5px;` above which means we don't get that funky little underline in safari.

app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -10,21 +9,36 @@
...
       font-size: 16px;
     }
 
     .active {
       border-bottom: 5px solid black;
+      margin-bottom: 0px !important;
```
tommotaylor: This ensures if the link is active, it has 5px underline rather than the 5px margin which the rest have from L31. Ensures that all the subnav links line up.

app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -10,21 +9,36 @@
...
+    justify-content: space-around;
+
+    @media (min-width: 640px) {
+      justify-content: flex-end;
+    }
+  }
+
   nav {
     text-align: right;
+    margin-top: 4px;
```
tommotaylor: I am very dissappointed with this margin top. We previously used flexbox to ensure that the left column and right column were both 100% height. Unfortunately we can't use `display: flex` anymore as it breaks the responsive `-from-desktop` classes that Gov.uk design system provides.

This 4px makes sure the left column and right column are the same height.

app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -10,21 +9,36 @@
     padding-bottom: 10px;
   }
 
+  .nav-links {
+    display: flex;
+    justify-content: space-around;
+
+    @media (min-width: 640px) {
+      justify-content: flex-end;
```
tommotaylor: Right align the subnav links on desktop. Centred otherwise.

app/views/application/_subnavigation.html.erb
```diff
@@ -1,13 +1,13 @@
...
+  <div class='govuk-grid-column-one-quarter govuk-grid-column-one-third-from-desktop govuk-!-padding-left-0'>
     <p class='govuk-body organisation-name'>
       <% unless current_user.super_admin? %>
         <strong><%= current_organisation.name %></strong>
       <% end %>
     </p>
   </div>
 
-  <div class='govuk-grid-column-two-thirds govuk-!-margin-bottom-0 govuk-!-padding-right-0 flex-end'>
+  <div class='govuk-grid-column-three-quarters govuk-grid-column-two-thirds-from-desktop govuk-!-margin-bottom-0 govuk-!-padding-0 nav-links'>
```
tommotaylor: Combination of the two grid classes allows for mobile, tablet and desktop.

app/assets/stylesheets/components/sub_nav.scss
```diff
@@ -1,34 +1,49 @@
 .subnav {
-  display: flex;
-  margin-left: 0px;
-  margin-right: 0px;
   border-bottom: 1px solid black;
 
+  @media (min-width: 640px) {
```
cybojenix: does 640px mean anything? It seems to be used a lot

app/assets/stylesheets/components/sub_nav.scss
cybojenix: Would it be worth putting this comment as documentation instead?

cybojenix:
Looks fine, but won't approve as I don't have the domain knowledge for the admin frontend yet.

Would be nice to have your comments converted to documentation, as I feel there's some valuable information in there that will get lost otherwise

app/assets/stylesheets/components/sub_nav.scss
tommotaylor: Its the Gov.uk design system breakpoint. Which interestingly is not the same as bootstraps which  I consider the industry norm

PR #478, opened 2019-02-18
Basic scaffold for managing authorised email domains

emileswarts:


app/controllers/admin/authorised_email_domains_controller.rb
```diff
@@ -0,0 +1,25 @@
+class Admin::AuthorisedEmailDomainsController < AdminController
+  def index
+    @authorised_email_domains = AuthorisedEmailDomain.all
```
georgeschena: Is this used to list all the allowed domains?

emileswarts: Yup

app/views/admin/authorised_email_domains/index.html.erb
```diff
@@ -0,0 +1,3 @@
+ <%= @authorised_email_domains.each do |custom_email_domain| %>
```
tommotaylor: don't think you need this `=`

emileswarts: Ah yes good catch

PR #480, opened 2019-02-18
Use case to generate a signup whitelist

emileswarts:
Given a list of domains, generate the PERL regular expression that will
be used to make sure someone is allowed to sign up to the admin or user
signup API.

A list of:
foo.com, bar.gov.uk

Will result in a regex like:
^.*@(foo\.com|bar\.gov\.uk)$


tommotaylor:
Looks good to me

PR #485, opened 2019-02-19
Style the whitelist

emileswarts:
Also rename the feature spec file to not be about adding exclusively


app/views/admin/authorised_email_domains/index.html.erb
```diff
@@ -1,3 +1,11 @@
-<% @authorised_email_domains.each do |custom_email_domain| %>
-  <%= custom_email_domain %>
-<% end %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Whitelisted Signup Domains</h1>
```
georgeschena: We agreed on calling it whitelisted domains? Appaz some people don't know what Whitelisted means

emileswarts: Yeah this isn't exposed to end users, only super admins

PR #488, opened 2019-02-21
Upload whitelist after deleting an entry

Renny4Real:
**WHY:**
As part of the upcoming functionality to delete an email domain. This makes sure that a new whitelist is uploaded after a deletion occurs.

**IN THIS PR:**
- Add a test to describe the function in the feature.
- Add a test to describe the DELETE request.
- Add remove button in the whitelisted email domains table.

**STEP ONE:**

screenshot 2019-02-22 at 12 43 01

**STEP TWO:**

screenshot 2019-02-22 at 12 44 12

**STEP THREE:**

screenshot 2019-02-22 at 12 57 08

### DIFF
--- a/app/controllers/admin/authorised_email_domains_controller.rb
+++ b/app/controllers/admin/authorised_email_domains_controller.rb
@@ -26,6 +26,13 @@ class Admin::AuthorisedEmailDomainsController < AdminController
     end
   end
 
+  def destroy
+    authorised_email_domain = AuthorisedEmailDomain.find_by(id: params.fetch(:id))
+    authorised_email_domain.destroy
+
+    redirect_to admin_authorised_email_domains_path, notice: "#{authorised_email_domain.name} has been removed"
+  end
+
 private
 
   def authorised_email_params
--- a/app/controllers/admin/authorised_email_domains_controller.rb
+++ b/app/controllers/admin/authorised_email_domains_controller.rb
@@ -27,8 +27,17 @@ class Admin::AuthorisedEmailDomainsController < AdminController
   end
 
   def destroy
-    authorised_email_domain = AuthorisedEmailDomain.find_by(id: params.fetch(:id))
+    authorised_email_domain = AuthorisedEmailDomain.find_by(id: authorised_email_params.fetch(:id))
+
     authorised_email_domain.destroy
+    # UseCases::Administrator::PublishSignupWhitelist.new(
+    #     destination_gateway: Gateways::S3.new(
+    #       bucket: ENV.fetch('S3_SIGNUP_WHITELIST_BUCKET'),
+    #       key: ENV.fetch('S3_SIGNUP_WHITELIST_OBJECT_KEY')
+    #     ),
+    #     source_gateway: Gateways::AuthorisedEmailDomains.new,
+    #     presenter: UseCases::Administrator::CreateSignupWhitelist.new
+    #   ).execute
 
     redirect_to admin_authorised_email_domains_path, notice: "#{authorised_email_domain.name} has been removed"
   end
--- a/app/controllers/admin/authorised_email_domains_controller.rb
+++ b/app/controllers/admin/authorised_email_domains_controller.rb
@@ -27,7 +27,7 @@ class Admin::AuthorisedEmailDomainsController < AdminController
   end
 
   def destroy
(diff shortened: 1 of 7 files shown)


app/controllers/admin/authorised_email_domains_controller.rb
```diff
@@ -26,6 +26,13 @@ def create
     end
   end
 
+  def destroy
+    authorised_email_domain = AuthorisedEmailDomain.find_by(id: params.fetch(:id))
```
tommotaylor: use the sanitised params: `authorised_email_params` from L38 rather than the raw `params`

spec/features/super_admin/authorised_email_domain_spec.rb
```diff
@@ -37,6 +37,21 @@
         end
       end
 
+      # context 'given I want to delete a whitelisted domain' do
+      #   it 'removes a domain' do
+      #     create(:authorised_email_domain, name: 'police.uk')
+      #     visit admin_authorised_email_domains_path
+
+      #     expect { click_on 'Delete' }.to change { AuthorisedEmailDomain.count }.by(-1)
+      #     expect(page).to have_content("#{authorised_email_domain.name} has been deleted")
```
tommotaylor: The reason you are gettin...
```
undefined local variable or method 'authorised_email_domain' for #<RSpec::ExampleGroups::DELETEAuthorisedEmailDomainsId::
```

Is because `authorised_email_domain` is not defined. 

At L42 you create an authorised email domain using a factory `create(:authorised_email_domain, name: 'police.uk')` but you do not save this object to a variable. If you want to save it to a variable so you can refer to it later you should do something like:
`let(:authorised_email_domain,) { create(:authorised_email_domain, name: 'police.uk') }`

spec/requests/email_domain/delete_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  before do
+    https!
+    login_as(admin_user, scope: :user)
+    create(:authorised_email_domain, name: 'some.domain.org.uk')
+  end
+
+  context "when the user is a super admin" do
+    it "deletes the email domain" do
+      expect {
+        delete admin_authorised_email_domains_path(authorised_email_domain)
```
tommotaylor: Same as above, `authorised_email_domain` doesn't reference anything, so is `nil`.

Renny4Real:
Not too sure when I'm getting this error: `No route matches [DELETE] "/admin/authorised_email_domains.7"`. 
Is this the right approach?

spec/requests/email_domain/delete_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'publishes the authorised domains to S3' do
+      expect_any_instance_of(Gateways::S3).to receive(:upload)
+      delete admin_authorised_email_domain_path(email_domain_1)
+    end
+  end
+
+  context "when the user is not super admin" do
+    before do
+      https!
+      login_as(normal_user, scope: :user)
```
tommotaylor: Rather than having `let(:normal_user) { create(:user, super_admin: false) }` right up the top of the file and then referencing it here, you could make this a bit more readable by creating the normal user right here where you need it.

Eg:
```
before do
   https!
   login_as( create(:user), scope: :user)
end
```

Also you should use the method `sign_in_user user` not `login_as(user)`. Its defined in `features/support.sign_up_helpers.rb`.

spec/requests/email_domain/delete_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    it "does not delete the email domain" do
+      expect {
+        delete admin_authorised_email_domain_path(email_domain_2)
```
tommotaylor: You shouldn't need an additional `email_domain_2` here. The `email_domain_1` declared at the start should be recreated and available for each `context`

app/views/admin/authorised_email_domains/_table.html.erb
```diff
@@ -1,9 +1,10 @@
 <table class="govuk-table govuk-!-margin-bottom-8">
-  <tbody class="govuk-table__body" id="ips-table">
+  <tbody class="govuk-table__body" id="domains-table">
```
tommotaylor: If you aren't using this id anywhere you probably don't need to declare it.

Renny4Real:
@tommotaylor Thanks.
 Yeah, more to come in upcoming PR's. 
Don't want to merge too much all at once.

PR #489, opened 2019-02-25
Add read method to s3 gateway

Renny4Real:
NB: In a later change, will rename the upload method to write to maintain consistency.

**IN THIS PR:**
- Add read method to s3 gateway.
- This will be used to fetch the user sign up whitelist regex from the s3 bucket. 
- This read method is currently unused at this point.

### DIFF
--- a/spec/gateways/s3_spec.rb
+++ b/spec/gateways/s3_spec.rb
@@ -22,10 +22,6 @@ describe Gateways::S3 do
       }
     end
 
-    after do
-      Rails.application.config.s3_aws_config = { stub_responses: true }
-    end
-
     it 'returns the contents' do
       expect(subject.read).to eq('some data')
     end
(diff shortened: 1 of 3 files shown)


spec/gateways/s3_spec.rb
```diff
@@ -8,4 +8,26 @@
...
+            if context.params.fetch(:bucket) == bucket && context.params.fetch(:key) == key
+              { body: 'some data' }
+            end
+          }
+        }
+      }
+    end
+
+    after do
+      Rails.application.config.s3_aws_config = { stub_responses: true }
```
cybojenix: it might be better to restore this to the actual original value, to reduce duplication and potential hidden breakages.

```ruby
around do |example|
  original_s3_aws_config = Rails.application.config.s3_aws_config
  begin
    example.run
  ensure
    Rails.application.config.s3_aws_config = original_s3_aws_config
    # set config
  end
end
```

PR #490, opened 2019-02-25
Allow any subdomain for signup whitelist

Renny4Real:
We need this regex to be more lenient and to allow any subdomains of trusted TLDs.
This won't cause an issue and should remain largely the same.

### DIFF
--- a/lib/use_cases/administrator/create_signup_whitelist.rb
+++ b/lib/use_cases/administrator/create_signup_whitelist.rb
@@ -6,7 +6,7 @@ module UseCases
       def execute(domains)
         return NOOP_REGEX if domains.empty?
 
-        '^.*@([a-zA-Z0-9.-]+)?' + literal_dot(domains_list(domains) + '$')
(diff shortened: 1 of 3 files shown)


lib/use_cases/administrator/create_signup_whitelist.rb
```diff
@@ -6,7 +6,7 @@ class CreateSignupWhitelist
       def execute(domains)
         return NOOP_REGEX if domains.empty?
 
-        '^.*@' + literal_dot(domains_list(domains) + '$')
+        '^.*@([a-zA-Z0-9.-]+)?' + literal_dot(domains_list(domains) + '$')
```
cybojenix: this allows more than just subdomains, as the `.` is now optional.
`someone@maliciousgov.uk` will match

cybojenix: suggest changing to `([a-zA-Z0-9-]+\.)?`

PR #499, opened 2019-02-26
Add stats for super admin

georgeschena:
screenshot 2019-02-26 at 15 29 00

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -29,3 +29,12 @@ $govuk-font-url-function: "font-url";
   margin-left: 8px;
   vertical-align: middle;
 }
+
+.total-org-stat-results-title {
+  font-size: 63px;
+  padding-right: 54px;
+}
+
+.admin-panel-stats-container {
+  padding-left: 125px;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -12,6 +12,7 @@ $govuk-font-url-function: "font-url";
 @import "./components/sub_nav";
 @import "./components/team_members";
 @import "./utilities/typography";
+@import "./components/admin_stats_panel.scss";
 
 .govuk-main-wrapper {
   margin-top: -25px;
@@ -28,13 +29,4 @@ $govuk-font-url-function: "font-url";
 .beta-logo {
   margin-left: 8px;
   vertical-align: middle;
-}
-
-.total-org-stat-results-title {
-  font-size: 63px;
-  padding-right: 54px;
-}
-
-.admin-panel-stats-container {
-  padding-left: 125px;
 }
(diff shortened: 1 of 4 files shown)


app/controllers/admin/organisations_controller.rb
```diff
@@ -3,6 +3,10 @@ class Admin::OrganisationsController < AdminController
 
   def index
     @organisations = Organisation.includes(:signed_mou_attachment).order("#{sort_column} #{sort_direction}").all
+
+    @total_number_of_organisations = Organisation.all.count
+    @total_number_of_locations = Location.all.count
+    @total_number_of_ips = Ip.all.count
```
tommotaylor: `Location.count` and `Ip.count` is more succinct.

app/assets/stylesheets/application.scss
```diff
@@ -29,3 +29,12 @@ $govuk-font-url-function: "font-url";
...
   vertical-align: middle;
 }
+
+.total-org-stat-results-title {
+  font-size: 63px;
+  padding-right: 54px;
+}
+
+.admin-panel-stats-container {
+  padding-left: 125px;
```
tommotaylor: This seems a bit weird. Are you using padding to centre the container? If so, you should use something like flexbox instead.

app/assets/stylesheets/application.scss
```diff
@@ -12,6 +12,7 @@ $govuk-font-url-function: "font-url";
 @import "./components/sub_nav";
 @import "./components/team_members";
 @import "./utilities/typography";
+@import "./components/admin_stats_panel.scss";
```
tommotaylor: This list used to be alphabetical. 😞

app/views/admin/organisations/index.html.erb
```diff
@@ -1,7 +1,31 @@
-<h1 class="govuk-heading-l">All organisations</h1>
-<h3 class="govuk-heading-m">
-  <%= pluralize(@organisations.count, 'organisation') %> are signed up to GovWifi
-</h3>
+<h1 class="govuk-heading-l">Summary of all organisations</h1>
+<div class="govuk-grid-column admin-panel-stats-container">
+  <div class="govuk-grid-column-one-third stats-align-center">
+    <div class="govuk-heading-l">
+    <strong class="total-org-stat-results-title"><%= @total_number_of_locations%></strong>
```
emileswarts: Closing tag whitespace is off, not sure why rubocop isn't picking up on this.
Have added to tech debt board to check

kjdchapman:
Had a conversation with Steve, and think we can do something much simpler here. 

He's certainly interested in how many locations we have total (he gets asked that fairly regularly!), but not interested in how many IP addresses we have total. 

Going to put up another PR to demonstrate.

kjdchapman:
Suggested an alternate approach in https://github.com/alphagov/govwifi-admin/pull/503

PR #502, opened 2019-02-27
Disable sentry notifications

emileswarts:
Print to Cloudwatch instead when someone couldn't sign up due to
whitelisting


kjdchapman:
I suddenly realised this was about whitelisting, I hadn't realised what those errors were earlier. Just skipped straight to annoyed by alerts instead of thinking, my bad.

Feels like it's too loud in Sentry, but too quiet in Cloudwatch. Can we do anything else useful with this metric?

emileswarts:
@kjdchapman Yeah we need something in between.
This was added so I could see how the new feature was doing.

PR #503, opened 2019-02-27
Show total orgs/locations to super-admins

kjdchapman:
Steve wasn't interested in IP totals (especially now you can't add a location without an IP) - from talking/demoing to him, I believe this is what is needed.

It's possible I missed some context whilst out of the office, so feel free to challenge this!

image

### DIFF
--- a/app/views/admin/organisations/index.html.erb
+++ b/app/views/admin/organisations/index.html.erb
@@ -6,7 +6,7 @@
   </span>
    across
   <span class="govuk-!-font-weight-bold">
-    <%= pluralize(@organisations.count, 'organisation') %>.
+    <%= pluralize(@organisations.length, 'organisation') %>.
   </span>
 </h3>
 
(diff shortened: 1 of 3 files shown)


kjdchapman:
This is an alternate approach to #499.

app/views/admin/organisations/index.html.erb
```diff
@@ -1,6 +1,13 @@
...
-<h3 class="govuk-heading-m">
-  <%= pluralize(@organisations.count, 'organisation') %> are signed up to GovWifi
+<h3 class="govuk-body">
+  GovWifi is in
+  <span class="govuk-!-font-weight-bold">
+    <%= pluralize(@location_count, 'location') %>
+  </span>
+   across
+  <span class="govuk-!-font-weight-bold">
+    <%= pluralize(@organisations.count, 'organisation') %>.
```
tommotaylor: Would `@organisations.length` be more performant here, as the `@organisations` is already loaded, calling `.length` on it as if its an array means it doesn't run the SQL again.

kjdchapman: 'scuse me missing this same comment on the other PR! I'll try it out - hadn't realised `.count` would trigger another query.

kjdchapman: Yep, that works, cuts out a query - nice one.

kjdchapman: Addressed in https://github.com/alphagov/govwifi-admin/pull/503/commits/b7905fbb5a31e77e6902d6a2eed561b6c532d116

PR #506, opened 2019-02-28
WIP: Replace cloudwatch status with external service page

kjdchapman:
**This is a draft** because:
- we need to pay for statuspage.io first. We should be able to do that in short order with Steve.
- Steve was keen on having the service status be easily available rather than buried in support - perhaps in the top-level nav?

image

We could look at piping our status into this external page somehow. I'm not sure how needed it is given the other teams' behaviour.

I did a little styling work on the external service page so it looks like Notify's:

image


cybojenix:
Can reopen this once we're ready

PR #508, opened 2019-03-04
Stop passing the user email as a query string to the pending URL

Renny4Real:
**WHY:**
PII is any data that can potentially identify a particular individual. 
We currently pass the email address of a new user signing up as a query string in the pending page URL. This allowed for the confirmation message to be personalised to the specific user.
 
**INITIAL SCREENSHOT:**

screenshot 2019-03-04 at 16 07 24
**Please note:** 
**- The URL in the screenshot.**
**- The personalised message.**

**IN THIS PR:**
However, in order be more compliant with GDPR standards and be prepared to use Google Analytics more effectively _(We were advised that G.A have some policies around user-specific details being passed as query strings in URLs)_:
1. We have stopped passing params to this URL.

**RESULT 1:**

screenshot 2019-03-04 at 16 09 02
**Please note:** 
**- The URL in the screenshot.**
**- The lack of a personalised message.**

2. The personalised message was changed to be more generic.

**RESULT 2:**

screenshot 2019-03-04 at 16 09 41
**Please note:** 
**- The URL in the screenshot.**
**- The generic message.**

**Is there a better solution?**
So currently there is a failing test as a result of the lack of a personalised message.
- Is there a way to make the user details (ideally the user's email) available as a variable in the `pending method` in the `ConfirmationsController`?
- Is there another way to hide what we are passing to the URL? 
- Have read some stuff about changing the GET request to a POST, any insight on this?

**The failing test could be edited/deleted but wanted some feedback on this first, would be nice to keep the personalised message, if possible.**

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -6,8 +6,8 @@ class Users::RegistrationsController < Devise::RegistrationsController
 protected
 
   # The path used after sign up for inactive accounts.
-  def after_inactive_sign_up_path_for(resource)
-    users_confirmations_pending_path(email: resource.email)
+  def after_inactive_sign_up_path_for(_resource)
+    users_confirmations_pending_path
   end
 
   def validate_email_on_whitelist
--- a/app/views/users/confirmations/pending.html.erb
+++ b/app/views/users/confirmations/pending.html.erb
@@ -1,7 +1,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-three-quarters">
     <h1 class="govuk-heading-l">Check your email</h1>
-    <p class="govuk-body-l">A confirmation email has been sent to <strong><%= params[:email] %>.</strong></p>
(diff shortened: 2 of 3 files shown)


emileswarts:
Wonder if "your email address" should be in bold.
Almost like someone would get confused about which inbox they should check.

Renny4Real:
@emileswarts Yeah, could remove the emphasis, if it could potentially confuse the user.

Renny4Real:
**UPDATE:**

**STYLE 1:**

screenshot 2019-03-05 at 10 07 39

**STYLE 2:**

screenshot 2019-03-05 at 10 08 24

**Suggestions and/or feedback would be appreciated**

georgeschena:
Only comment I have is maybe specify for them to check their government email

Renny4Real:
@georgeschena They might not be government department. Some orgs are whitelisted.

Renny4Real:
**CURRENT SOLUTION:**

screenshot 2019-03-05 at 10 07 39

PR #511, opened 2019-03-05
Better MoU upload feedback

georgeschena:
screenshot 2019-03-05 at 13 05 16

### DIFF
--- a/app/views/mou/index.html.erb
+++ b/app/views/mou/index.html.erb
@@ -1,5 +1,3 @@
-<% mou_attached = @current_org_signed_mou %>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h2 class="govuk-heading-l">Sign your agreement</h2>
@@ -17,7 +15,7 @@
       </p>
     <% end %>
 
-    <% if mou_attached %>
+    <% if @current_org_signed_mou %>
       <p class="govuk-body">
         A signed MoU was uploaded on <%= @current_org_signed_mou.created_at.strftime('%e %b %Y') %>,
         <%= link_to "download and view the document.", rails_blob_path(@current_org_signed_mou, disposition: "attachment"), class: "govuk-link" %>
@@ -33,7 +31,7 @@
         <div class="govuk-form-group">
           <%= form.file_field(:signed_mou, class: 'govuk-file-upload') %>
         </div>
-        <% submit_text = mou_attached ? 'Replace MOU' : 'Upload MOU' %>
+        <% submit_text = @current_org_signed_mou ? 'Replace MOU' : 'Upload MOU' %>
         <%= form.submit(submit_text, class: 'govuk-button') %>
       <% end %>
     </div>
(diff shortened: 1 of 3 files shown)


app/views/mou/index.html.erb
```diff
@@ -14,12 +16,25 @@
         <%= link_to "support form", help_index_path, class: "govuk-link" %> if you need a copy.
       </p>
     <% end %>
+
+    <% if current_organisation.signed_mou.attached? %>
```
emileswarts: A variable for this has been created above

spec/features/upload_download_mous_spec.rb
```diff
@@ -23,14 +23,24 @@
...
         expect(page).to have_content("MOU uploaded successfully.")
       end
+
+      it 'allows the user to see when the mou was uploaded' do
+        expect(page).to have_content("A signed MoU was uploaded on")
+      end
+
+      it 'allows the user to click the link and download their MoU' do
+        expect(page).to have_content("download and view the document.")
+        click_on 'download and view the document.'
```
emileswarts: Is this click_on asserting that the button is on the page?
I would usually expect an action like this with an assertion at the end.

georgeschena: Yeah I don't need that click really the line above ensures its on the page anyway

app/views/mou/index.html.erb
tommotaylor: Looks like the db lookup is happening 3x here for `current_organisation.signed_mou`. Perhaps this could be put into an instance variable in the controller? Or does `current_organisation.signed_mou` not actually load the object?

spec/features/upload_download_mous_spec.rb
```diff
@@ -23,14 +23,23 @@
...
         attach_file("signed_mou", Rails.root + "spec/fixtures/mou.pdf")
         click_on 'Upload'
+      end
 
+      it 'can upload and download a version of the mou' do
         expect(page).to have_content("MOU uploaded successfully.")
       end
+
+      it 'allows the user to see when the mou was uploaded' do
+        expect(page).to have_content("A signed MoU was uploaded on")
```
emileswarts: I know it's not good to have multiple assertions per it block but we're executing the before block 3 times here to check the text on the same page.

What are your thoughts?

georgeschena: Condense it into one IT block sounds good

app/views/mou/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% mou_attached = @current_org_signed_mou %>
```
emileswarts: Don't see this intermediate variable being needed

georgeschena: Good spot

tommotaylor:
One thing, and I'm being pretty picky here... as you correctly saw, theres no unneccesary db lookups, but you are performing lookups from the view. You can see this because the lookup in the logs happens after the view render. Eg:
screenshot 2019-03-05 at 14 27 59

By changing to the following, you end up doing all the lookups in the controller, and the view remains dumb:
screenshot 2019-03-05 at 14 28 44
screenshot 2019-03-05 at 14 28 48

PR #513, opened 2019-03-07
Replace support email with contact us link

Renny4Real:
**WHY:**
When a user signs up to create an account for an existing organisation. The error message includes the support email. 

It would be better to change this to a link to the offline support form with the app.

**IN THIS PR:**
Add a link within the message in the `en.yml` file.

**BEFORE:**

**AFTER:**
screenshot 2019-03-07 at 10 46 55

### DIFF
--- a/app/views/layouts/_form_errors.html.erb
+++ b/app/views/layouts/_form_errors.html.erb
@@ -1,11 +1,11 @@
 <% if defined? resource %>
   <% if resource&.errors&.any? %>
-    <div class="govuk-error-summary">
+    <div class="govuk-error-summary" id="error-summary">
       <h2 class="govuk-error-summary__title">There is a problem</h2>
       <div class="govuk-error-summary__body">
         <ul class="govuk-list govuk-error-summary__list">
           <% resource.errors.full_messages.each do |message| %>
-            <li><%= message %></li>
+            <li><%= message.html_safe %></li>
           <% end %>
         </ul>
       </div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -12,6 +12,6 @@ en:
           attributes:
             name:
               taken: " is already registered. If you wish to administer an \n
-              existing GovWifi installation, you must be invited to the \n
-              account. <a class='govuk-link' href='/help/new'>Contact us</a> \n
+              existing GovWifi installation, you must be invited to that \n
+              organisation's team. <a class='govuk-link' href='/help/new'>Contact us</a> \n
               if you need help."
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -13,5 +13,5 @@ en:
             name:
               taken: " is already registered. If you wish to administer an \n
               existing GovWifi installation, you must be invited to the \n
-              account. Please contact govwifi-support@digital.cabinet-office.gov.uk \n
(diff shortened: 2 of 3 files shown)


config/locales/en.yml
```diff
@@ -13,5 +13,5 @@ en:
             name:
               taken: " is already registered. If you wish to administer an \n
```
kjdchapman: Some thoughts on the content of this message:
- It's quite long, I worry people will have trouble understanding it
- It uses `please` - GOV.UK convention is to not
- It's annoying that this has to start with "email" - I think it makes more sense to talk about *accounts* here (I think this is supported by research from other GDS services too). 

Suggested reword:
> You already have a GovWifi account.
> If you want to join the team managing #{org_name}'s GovWifi installation, you need to be invited by someone already on that team.
> If you can't find a member of that team, [contact us](link)

This re-word doesn't take into account technical feasibility. If putting the org name / links in there is hard, could skip some of it.

kjdchapman:
image

Is this link supposed to be red?

config/locales/en.yml
tommotaylor: @kjdchapman the reason it has "Email" at the start is because its an ActiveRecord::Base error that applies to the email attribute on the user model. To get rid of email requires doing this as a unique validator and declaring the error message in the model rather than the `yml`.

Renny4Real:
@kjdchapman Yeah, it's from the `.gov-uk-summary__links` class.

Renny4Real:
**UPDATE:**

screenshot 2019-03-07 at 12 02 32

config/locales/en.yml
kjdchapman: Yeah, sounds like a lot of trouble to sort out - let's skip it.

kjdchapman:
I like this new messaging better!

One tweak - I would say:
> invited to that organisation's team

rather than
> invited to the account

Renny4Real:
**UPDATE 2**:

screenshot 2019-03-07 at 14 54 49

app/views/layouts/_form_errors.html.erb
```diff
@@ -1,11 +1,11 @@
 <% if defined? resource %>
   <% if resource&.errors&.any? %>
-    <div class="govuk-error-summary">
+    <div class="govuk-error-summary" id="error-summary">
       <h2 class="govuk-error-summary__title">There is a problem</h2>
       <div class="govuk-error-summary__body">
         <ul class="govuk-list govuk-error-summary__list">
           <% resource.errors.full_messages.each do |message| %>
-            <li><%= message %></li>
+            <li><%= message.html_safe %></li>
```
kjdchapman: [Having a discussion about this](https://github.com/alphagov/govwifi-admin/pull/514#discussion_r263419156) on the other open PR!

cybojenix: `html_safe` is being used too broadly here, and will lead to potential XSS exploits.

cybojenix: this has been discussed, no data actually gets into this message currently, but it would be nice if this is flagged in the tech debt board and have it noted somewhere that we can't put data into these flash messages

kjdchapman: Added a card to the tech debt board.

PR #514, opened 2019-03-07
Better error messaging on signup 

georgeschena:
Empty or invalid email domain:
screenshot 2019-03-07 at 12 19 17

Already existing email:
screenshot 2019-03-07 at 12 19 23

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -27,7 +27,7 @@ protected
 
   def set_user_object_with_errors
     @user = User.new(sign_up_params)
-    @user.errors.add(:email, 'must be from a government domain')
+    @user.errors.add(:email, "must be from a government domain. If you're having trouble signing up, <a href='/help/new'>contact us</a>.")
   end
 
   def return_user_to_registration_page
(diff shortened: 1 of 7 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -27,7 +27,7 @@ def validate_email_on_whitelist
 
   def set_user_object_with_errors
     @user = User.new(sign_up_params)
-    @user.errors.add(:email, 'must be from a government domain')
+    @user.errors.add(:email, "must be from a government domain. If you're having trouble signing up, <a href='/help/new'>contact us</a>.")
```
cybojenix: is there a way to load the href from the routes, instead of it being hard coded?

app/views/layouts/_form_errors.html.erb
```diff
@@ -5,7 +5,7 @@
       <div class="govuk-error-summary__body">
         <ul class="govuk-list govuk-error-summary__list">
           <% resource.errors.full_messages.each do |message| %>
-            <li><%= message %></li>
+            <li><%= message.html_safe %></li>
```
cybojenix: Please don't do this. it will lead to potential for a XSS attack, as this is being set broadly.
Instead, apply `html_safe` on the string with html in itself (`./app/controllers/users/registrations_controller.rb:30`)

kjdchapman: Notably, https://github.com/alphagov/govwifi-admin/pull/513 is committing this sin as well - we should block both if we're going to have this discussion.

kjdchapman:
Mentioned it on the other PR, but wrote a card on our tech debt board so we can decide if we want to spend the time trying to fix this.

PR #521, opened 2019-03-12
Format date in the unconfirmed user list in super admin

Renny4Real:
**WHY:**
Better to change the `created_at` time to a more human readable format.

**IN THIS PR:**
- Add unconfirmed users to the `seeds.rb` file
- Use `strftime` method to format the time in the `index.html.erb`.

**BEFORE:**

Screenshot 2019-03-12 at 16 31 00

**AFTER:**

Screenshot 2019-03-12 at 16 31 12


cybojenix:
Merging as I know you're not in tomorrow

PR #523, opened 2019-03-14
Filter tables 

georgeschena:


spec/features/admin/view_unconfirmed_users_spec.rb
```diff
@@ -24,5 +25,9 @@
       expect(page).to have_content(unconfirmed_user_2.email)
       expect(page).to_not have_content(confirmed_user.email)
     end
+
+    it 'displays the list of all unconfirmed users in alphabetical order' do
+      expect(page.body).to match(/test2@gov.uk.*test3@gov.uk/m)
```
cybojenix: I'm unsure on how much this test proves.

It's possible for the code to be ordering it based on when a row was created.

Suggest having 3 unconfirmed users, and assert that the creation order is different to the email address ordering

spec/features/super_admin/add_custom_organisation_spec.rb
```diff
@@ -6,11 +6,19 @@
...
+    let!(:org2) { CustomOrganisationName.create(name: 'Custom Org 2') }
+    let!(:org3) { CustomOrganisationName.create(name: 'Custom Org 3') }
+
     before do
       sign_in_user admin_user
       visit admin_custom_organisations_path
     end
 
+    it 'displays the list of all custom organisations in alphabetical order' do
+      expect(page.body).to match(/Custom Org 1.*Custom Org 2.*Custom Org 3/m)
```
cybojenix: as above, aim for the alphabetical order to not align with the creation order

app/controllers/admin/authorised_email_domains_controller.rb
```diff
@@ -47,4 +49,16 @@ def destroy
   def authorised_email_params
     params.require(:authorised_email_domain).permit(:name)
   end
+
+  def sortable_columns
+    %w[name]
+  end
+
+  def sort_column
+    sortable_columns.include?(params[:sort]) ? params[:sort] : "name"
```
cybojenix: While it's not a lot of duplication, would be good to use `sortable_columns` for the default:

```suggestion
    sortable_columns.include?(params[:sort]) ? params[:sort] : sortable_columns.first
```

app/controllers/admin/authorised_email_domains_controller.rb
```diff
@@ -47,4 +49,16 @@ def destroy
...
+  end
+
+  def sort_direction
+    %w[asc].include?(params[:direction]) ? params[:direction] : "asc"
```
cybojenix: can we also have `sort_directions` so we match the above pattern please?

app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -1,11 +1,13 @@
...
+
   def index
-    @custom_organisation_names = CustomOrganisationName.all
+    order_custom_orgs
     @custom_organisation = CustomOrganisationName.new
   end
 
   def create
-    @custom_organisation_names = CustomOrganisationName.all
+    order_custom_orgs
```
cybojenix: thinking in terms of readability of the code here:

I feel that it would be good to define all template parameters within the controller, as apposed to within a private method.

```suggestion
    @custom_organisation_names = ordered_custom_orgs
```

app/views/admin/manage_users/index.html.erb
```diff
@@ -5,7 +5,7 @@
     <thead class="govuk-table__head">
       <tr class="govuk-table__row">
         <th class="govuk-table__header govuk-!-width-one-third">Email</th>
-        <th class="govuk-table__header govuk-!-width-one-third">Created at</th>
+        <th class="govuk-table__header govuk-!-width-one-third">Created on</th>
```
cybojenix: shouldn't be in this PR, but meh

spec/features/super_admin/add_custom_organisation_spec.rb
```diff
@@ -6,11 +6,20 @@
   let!(:admin_user) { create(:user, super_admin: true) }
 
   context 'when visiting the custom organisations page' do
+    let!(:org2) { CustomOrganisationName.create(name: 'Custom Org 2') }
```
cybojenix: someone is definitely going to come and put these in order

georgeschena: ?

georgeschena: Haha

app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -1,11 +1,13 @@
 class Admin::CustomOrganisationsController < AdminController
+  helper_method :sort_column, :sort_direction
+
   def index
-    @custom_organisation_names = CustomOrganisationName.all
+    order_custom_orgs
```
cybojenix: same change as below please?

app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -30,4 +32,24 @@ def destroy
   def custom_org_params
     params.require(:custom_organisations).permit(:name)
   end
+
+  def order_custom_orgs
```
cybojenix: now that this is going to be returning the sorted names, this can be changed to `ordered_custom_orgs`.

Also no longer needs to set the variable

app/controllers/admin/custom_organisations_controller.rb
georgeschena: You mean same as the create method? @cybojenix

PR #528, opened 2019-03-15
Allow super admin to delete an unconfirmed user

Renny4Real:
**WHY:**
We've encountered a frequent problem:
- A user begins the sign up for the admin portal and realises that their organisation already has an account.
- They ask their network engineers to invite them to the organisation's account.
- Their network engineers login to the account and try invite to  their colleague.
-  However, their colleague is already in the database (as an unconfirmed user) and so cannot be invited to the admin platform. This shows as an error message to the network engineers.
- They then email support.
- A way to correct this is to delete the unconfirmed user directly from the database, and then email them explaining that they can now be invited.

There is a need for this to be solved by a non-technical person with no direct access to the database.

**IN THIS PR:**
- Add a delete functionality to the manage users page.

#### **EXAMPLE: Deleting test user: `travis@gov.uk`**
**STEP ONE:**

Screenshot 2019-03-15 at 11 43 07

**STEP TWO:**

Screenshot 2019-03-15 at 11 43 32

**STEP THREE:**

Screenshot 2019-03-15 at 11 43 42

**STILL TO DO:**
- Make the flash message more descriptive.

Screenshot 2019-03-15 at 11 43 32

The flash message doesn't make it clear which user is about to be deleted. 
Although the message can be altered to be more descriptive by passing the user's email as params in the URL, this is regarded as exposing PII which is not allowed in this project.

**Any suggestions/feedback on this would be appreciated.**

### DIFF
--- a/app/views/admin/manage_users/index.html.erb
+++ b/app/views/admin/manage_users/index.html.erb
@@ -1,3 +1,5 @@
+<%= render "confirm_remove_unconfirmed_user" if params[:user_to_remove] %>
+
 <h1 class="govuk-heading-l">Manage Users</h1>
 
 <table class="govuk-table">
@@ -6,6 +8,7 @@
       <tr class="govuk-table__row">
         <th class="govuk-table__header govuk-!-width-one-third">Email</th>
         <th class="govuk-table__header govuk-!-width-one-third">Created at</th>
+        <th class="govuk-table__header govuk-!-width-one-third"></th>
       </tr>
     </thead>
   <tbody class="govuk-table__body">
@@ -16,6 +19,9 @@
         </td>
         <td class="govuk-table__cell govuk-!-width-one-third" scope="row">
           <%= user.created_at.strftime('%e %b %Y') %>
+        </td>
+        <td class="govuk-table__cell govuk-!-width-one-third" scope="row">
+          <%= link_to 'Remove user', admin_manage_users_path(id: user.id, user_to_remove: true), class: "govuk-link" %>
         </td>
       </tr>
     <% end %>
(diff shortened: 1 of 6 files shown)


app/views/admin/manage_users/index.html.erb
```diff
@@ -17,6 +20,9 @@
         <td class="govuk-table__cell govuk-!-width-one-third" scope="row">
           <%= user.created_at.strftime('%e %b %Y') %>
         </td>
+        <td class="govuk-table__cell govuk-!-width-one-third" scope="row">
+          <%= link_to 'Remove user', admin_manage_users_path(id: user.id, user_to_remove: true), class: "govuk-link" %>
```
georgeschena: I feel this text should be red to match the other places where we have a destructive remove button/link

Renny4Real: I don't think it needs to be red. It's not immediate destruction.
And we haven't kept to a pattern:
IPs' remove link isn't red but team members' is.
Unless this is going to change?

cybojenix: There does appear to be some inconsistency across blue vs red text for destructive actions.

Given there's no guidance around link colours on types of actions beyond "links are blue", and "buttons are green", let's stick with blue for now (unless we already have some designer input around this)

cybojenix:
I feel the confirmation box should include the user's email, as currently it's not giving the user the opportunity to confirm they've selected the right user to remove.

Renny4Real:
@cybojenix I agree, I'm just not sure how to do this without passing the email as params. 
Wanted to get some suggestions.

tommotaylor:
I don't think this is the cleanest way to solve this problem. Could we not solve it without a whole MVC feature by simply allowing unconfirmed users to be invited to organisations?

tommotaylor:
Closed in favour of https://github.com/alphagov/govwifi-admin/pull/540

PR #533, opened 2019-03-18
Sort team members by name or email 

georgeschena:
Screenshot 2019-03-18 at 15 54 17

### DIFF
--- a/app/controllers/team_members_controller.rb
+++ b/app/controllers/team_members_controller.rb
@@ -4,7 +4,7 @@ class TeamMembersController < ApplicationController
   helper_method :sort_column, :sort_direction
 
   def index
-    @team_members = current_user.organisation.users.order(build_order_query)
+    @team_members = current_organisation.users.order(build_order_query)
   end
 
   def destroy
--- a/app/controllers/team_members_controller.rb
+++ b/app/controllers/team_members_controller.rb
@@ -4,7 +4,7 @@ class TeamMembersController < ApplicationController
   helper_method :sort_column, :sort_direction
 
   def index
-    @team_members = current_user.organisation.users.order("#{sort_column} #{sort_direction}")
+    @team_members = current_user.organisation.users.order(build_order_query)
   end
 
   def destroy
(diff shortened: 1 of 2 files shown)


app/controllers/team_members_controller.rb
```diff
@@ -4,7 +4,7 @@ class TeamMembersController < ApplicationController
   helper_method :sort_column, :sort_direction
 
   def index
-    @team_members = current_user.organisation.users.order("#{sort_column} #{sort_direction}")
+    @team_members = current_user.organisation.users.order(build_order_query)
```
tommotaylor: `current_user.organisation` already exists as `@current_organisation`.

app/controllers/team_members_controller.rb
```diff
@@ -26,6 +26,13 @@ def update
 
 private
 
+  def build_order_query
+    Arel::Nodes::NamedFunction.new("COALESCE", [
+      User.arel_table['name'],
+      User.arel_table['email']
+    ]).asc
+  end
```
tommotaylor: Nice

spec/features/team/view_team_members_spec.rb
```diff
@@ -35,14 +35,14 @@
...
-      let!(:user_3) { create(:user, email: 'zara@example.gov.uk', organisation: organisation) }
+      let!(:user_2) { create(:user, name: 'amada', organisation: organisation) }
+      let!(:user_3) { create(:user, name: 'zara', organisation: organisation) }
 
       it 'renders all team members within my organisation in alphabetical order' do
         sign_in_user user
         visit team_members_path
 
-        expect(page.body).to match(/#{user_2.email}.*#{user_1.email}.*#{user.email}.*#{user_3.email}/m)
+        expect(page.body).to match(/#{user_2.name}.*#{user_1.email}.*#{user.email}.*#{user_3.name}/m)
```
tommotaylor: I don't think this spec covers the feature, which is "sort by username, or email if username is missing". This spec only sorts by name, as `user_1` has an auto generated name of "bob" (https://github.com/alphagov/govwifi-admin/blob/master/spec/factories/users.rb#L7). 

For full coverage you would need to ensure at least one of the users in the list here doesn't have a name. You could just set it to an empty string in the declaration.

PR #534, opened 2019-03-18
Lint SCSS

tommotaylor:
Lint scss with govuk-lint: https://github.com/alphagov/govuk-lint

This PR implements the following:
- Moves the one external stylesheet we have to `vendors/assets/`
- Adds the scss linting step to make lint
- refactors the `application.scss` files
- Uses scss variables for all colours
- Fixes all other scss linting errors

Reviewer will need to pull this and check the new organisation signup flow as the "autocomplete" has been moved to a new folder. Also worth checking correct colours etc to ensure the new variables haven't broken anything.


Makefile
```diff
@@ -25,6 +25,8 @@ serve: build
 lint:
 	$(MAKE) build
 	$(DOCKER_COMPOSE) run --rm app bundle exec govuk-lint-ruby app lib spec Gemfile*
+	$(DOCKER_COMPOSE) run --rm app bundle exec govuk-lint-sass app/assets/stylesheets
```
cybojenix: suggest changing to this:

```mk
lint: lint-ruby lint-sass
lint-ruby: build
	$(DOCKER_COMPOSE) run --rm app bundle exec govuk-lint-ruby app lib spec Gemfile*
lint-sass: build
	$(DOCKER_COMPOSE) run --rm app bundle exec govuk-lint-sass app/assets/stylesheets
```

`build` will still only run once.

tommotaylor: Nice

app/assets/stylesheets/components/beta_banner.scss
```diff
@@ -0,0 +1,8 @@
+.govuk-phase-banner {
+  margin-bottom: 25px;
+}
+
+.beta-logo {
```
cybojenix: This wasn't originally on the ie8 stylesheet.

2 questions here:
- Should this now be applied to ie8?
- Do we still need ie8? (basically does this go up to ie11)

tommotaylor: ie8 uses a different govuk frontend package. And unfortunately there are users out there on ie8. GA would show us exactly how many, but our user testing found a number. Without this ie8 package, the footer is in the header and all sorts of other problems like that.

The reason it was not originally in the ie8 file was an error.

PR #535, opened 2019-03-18
Order locations logs search alphabetically 

georgeschena:
Screenshot 2019-03-19 at 12 08 45

### DIFF
--- a/app/controllers/logs_searches_controller.rb
+++ b/app/controllers/logs_searches_controller.rb
@@ -1,6 +1,7 @@
 class LogsSearchesController < ApplicationController
   def create
     @search = LogsSearch.new(search_params)
+    @locations = current_organisation.locations.order([:address])
 
     @search.first_step ? filter_choice : term_choice
   end
--- a/app/views/logs_searches/location.html.erb
+++ b/app/views/logs_searches/location.html.erb
@@ -11,7 +11,7 @@
     <%= form_for @search do |f| %>
       <div class='govuk-form-group'>
         <%= f.label :term, "Select one of your organisation's locations", class: 'govuk-label' %>
-        <%= collection_select :logs_search, :term, current_organisation.locations, :id, :full_address, {}, { class: 'govuk-select' } %>
+        <%= collection_select :logs_search, :term, current_organisation.locations.order([:address]), :id, :full_address, {}, { class: 'govuk-select' } %>
       </div>
       <%= f.hidden_field :filter %>
       <%= f.submit 'Show logs', class: 'govuk-button' %>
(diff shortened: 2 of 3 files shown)


app/views/logs_searches/location.html.erb
```diff
@@ -11,7 +11,7 @@
     <%= form_for @search do |f| %>
       <div class='govuk-form-group'>
         <%= f.label :term, "Select one of your organisation's locations", class: 'govuk-label' %>
-        <%= collection_select :logs_search, :term, current_organisation.locations, :id, :full_address, {}, { class: 'govuk-select' } %>
+        <%= collection_select :logs_search, :term, current_organisation.locations.order([:address]), :id, :full_address, {}, { class: 'govuk-select' } %>
```
tommotaylor: I would suggest putting this into the controller so the view just uses an instance variable such as `@locations`

spec/features/logging/view_auth_requests_for_a_location_spec.rb
```diff
@@ -11,6 +11,19 @@
...
+
+      let!(:organisation) { create(:organisation, locations: [location_2, location_3, location_4]) }
+
+      it 'shows me the locations in alpabetical order' do
+        locations = find('#logs_search_term').all('option').collect(&:text)
+        expect(locations).to match_array ["Abbey Street, HA7 2BL", "Garry Road, HA7 4BL", "Zeon Grove, HA7 3BL"]
+      end
+    end
+
     it 'shows me my locations' do
```
tommotaylor: This spec seems redundant now, the above spec tests for all the locations.

spec/features/logging/view_auth_requests_for_a_location_spec.rb
```diff
@@ -11,6 +11,19 @@
       click_on 'Go to search'
     end
 
+    context 'With a new set of locations' do
```
tommotaylor: Not a very descriptive context. Doesn't fit in with the rest of the spec. I wonder if we need this context at all.

app/controllers/logs_searches_controller.rb
```diff
@@ -1,6 +1,7 @@
 class LogsSearchesController < ApplicationController
   def create
     @search = LogsSearch.new(search_params)
+    @locations = current_organisation.locations.order([:address])
```
cybojenix: how come this isn't in the `location` controller?

georgeschena: Because we are using the LogsSearch controller to retrieve the logs for each filter

georgeschena: I do feel this LogsSearchController is a tad confusing imo

tommotaylor: Looks like its in the wrong controller action but doesn't work the other way. I'm also a little confused about how this controller is setup and how the routes work.

PR #539, opened 2019-03-19
Lint spec files

tommotaylor:
While we were calling the `/spec` folder to be linted, it wasn't being linted because we needed a `.rubocop.yml` file with `require: rubocop-rspec`


cybojenix:
@tommotaylor, this is going to take a while to do, and we've already got merge conflicts.
Let's get these changes passing, and remove the actual rubocop changes so we can merge this.

Then we can iteratively tackle this between us a branch at a time.


Also going to note that `bin`, `db`, and `config` need linting. Suggest we just run `govuk-lint-ruby` outright without any parameters, which will do ruby detection for us recursively.

spec/features/admin/view_unconfirmed_users_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'support/notifications_service'
```
cybojenix: merge conflict, this file shouldn't exist

PR #540, opened 2019-03-19
Invite an unconfirmed user by deleting them first

Renny4Real:
**WHY:**
We've encountered a frequent problem:
- A user begins the sign up for the admin portal and realises that their organisation already has an account.
- They ask their network engineers to invite them to the organisation's account.
- Their network engineers login to the account and try invite to  their colleague.
-  However, their colleague is already in the database (as an unconfirmed user) and so cannot be invited to the admin platform. This shows as an error message to the network engineers.
- They then email support.
- A way to correct this is to delete the unconfirmed user directly from the database, and then email them explaining that they can now be invited.

**IN THIS PR:**
Delete an unconfirmed user when they are invited and send them an invite anew.
(**Devise:Invitable** creates a new user instance when it sends an invite.)

**Any feedback on this would be appreciated**


spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -49,7 +49,7 @@
         end
       end
 
-      context "with non gov.uk email address" do
+      context "for new user with non gov.uk email address" do
```
tommotaylor: Anything not related to the actual issue should be in a separate PR. This just makes the PR unnecessarily more complicated.

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +64,30 @@
...
+        end
 
-        it "tells the user that the email has already been taken" do
-          expect {
-            click_on "Send invitation email"
-          }.to change { User.count }.by(0)
-          expect(InviteUseCaseSpy.invite_count).to eq(0)
-          expect(page).to have_content("Email is already associated with an account. If you can't sign in, reset your password")
+        context "and is not confirmed" do
+          let(:invited_user) { create(:user, :unconfirmed) }
```
tommotaylor: creating this user here creates it with an organisation. In the real life problem the user will not have an organisation so this test isn't quite right.

If the user already has an organisation, it means they have signed up or been invited so they should be skipped, land should probably get the message ""Email is already associated with an account. If you can't sign in, reset your password""

app/controllers/users/invitations_controller.rb
```diff
@@ -26,6 +26,7 @@ def validate_invited_user
   end
 
   def user_is_invalid?
+    delete_user_record if invited_user_already_exists? && invited_user_not_confirmed?
```
cybojenix: it isn't really the responsibility of the validation check to delete the user record.

It might be worth moving this to the same level as the `resending_invite?` is at:

```rb

def user_should_be_cleared?
  resending_invite? or invited_user_not_confirmed?
end
```

app/controllers/users/invitations_controller.rb
```diff
@@ -38,6 +39,10 @@ def invited_user_already_exists?
     !!User.find_by(email: invite_params[:email])
   end
 
+  def invited_user_not_confirmed?
+    User.find_by(email: invite_params[:email]).confirmed_at.nil?
```
cybojenix: Should we also check they're not associated with an organisation too?

app/controllers/users/invitations_controller.rb
tommotaylor: This looks like correct logic but in a weird part of this file. The method name `def user_is_invalid?` doesn't relate to checking if this user is an unconfirmed user.

I would suggest pulling it out so its more explicit, kind of how `before_action :delete_user_record, if: :resending_invite?, only: :create` makes a lot of sense.

One idea could be:
```
before_action :delete_user_record, if: :user_is_unconfirmed_and_without_org?, only: :create

def user_is_unconfirmed_and_without_org?
  invited_user_already_exists? && invited_user_not_confirmed?
end
```

This is just an example. The point is its currently quite hard to understand whats going on and `def user_is_invalid?` is the wrong place for the logic

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +66,40 @@
       end
 
       context "with an email that already exists" do
-        let(:invited_user_email) { user.email }
+        context "and is confirmed" do
```
tommotaylor: contexts should always start with "when", "with" or "without": http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +66,40 @@
...
+          let(:invited_user) { User.find_by(email: invited_user_email) }
+
+          it "sends them an invite to the organisation's account" do
+            expect {
+              click_on "Send invitation email"
+            }.to change { User.count }.by(0)
+            expect(InviteUseCaseSpy.invite_count).to eq(1)
+            expect(invited_user.confirmed?).to eq(false)
+            expect(invited_user.organisation_id).to eq(user.organisation_id)
+          end
```
tommotaylor: Still missing the example when the user already has an organisation. We don't want to be deleting users that have already been invited to an organisation.

app/controllers/users/invitations_controller.rb
tommotaylor: `invited_user_already_exists? && invited_user_not_confirmed?` doesn't cover the case of a user that is unconfirmed but has an organisation.

app/controllers/users/invitations_controller.rb
```diff
@@ -38,6 +47,14 @@ def invited_user_already_exists?
...
```
tommotaylor: `.confirmed?` would be a better method to use here: https://github.com/plataformatec/devise/blob/master/lib/devise/models/confirmable.rb#L106

eg: `!User.find.....confirmed?`
```suggestion
    User.find_by(email: invite_params[:email]).confirmed_at.nil?
```

app/controllers/users/invitations_controller.rb
```diff
@@ -25,6 +26,14 @@ def validate_invited_user
     return_user_to_invite_page if user_is_invalid?
   end
 
+  def user_should_be_cleared?
+    resending_invite? || unconfirmed_user_with_no_org?
+  end
+
+  def unconfirmed_user_with_no_org?
+    invited_user_already_exists? && invited_user_not_confirmed? && invited_user_has_no_org?
```
tommotaylor: We are doing three separate user lookups here, thats very inefficient. We should just be doing one.

spec/features/team/invite_a_team_member_spec.rb
tommotaylor: @Renny4Real these still need to be removed from this PR. Changes to line 54 and line 40.

spec/features/invite_existing_user_spec.rb
```diff
@@ -65,34 +65,9 @@
         click_on "Send invitation email"
       end
 
-      it "does not send an invitation" do
-        expect(InviteUseCaseSpy.invite_count).to eq(0)
-        expect(page).to have_content("Email is already associated with an account. If you can't sign in, reset your password")
-      end
-
-      context 'when user uses the confirmation link after invitation attempt' do
```
cybojenix: I think a user should still be able to click on the confirmation link, so this context should remain

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +66,55 @@
...
-          expect(InviteUseCaseSpy.invite_count).to eq(0)
-          expect(page).to have_content("Email is already associated with an account. If you can't sign in, reset your password")
+        context "when user is not confirmed and has no organisation" do
+          before do
+            sign_out
+            sign_up_for_account(email: invited_user_email)
+            sign_in_user user
+            visit root_path
+            click_on "Team"
+            click_on "Invite team member"
```
tommotaylor: This all seems very verbose:
```
visit root_path
click_on "Team"
click_on "Invite team member"
```
You can either visit the path directly or even manually generate the invitation?

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +66,55 @@
...
+            expect(InviteUseCaseSpy.invite_count).to eq(1)
+            expect(invited_user.confirmed?).to eq(false)
+            expect(invited_user.organisation_id).to eq(user.organisation_id)
+          end
+        end
+
+        context "when user is not confirmed but has a organisation" do
+          let(:organisation) { create(:organisation) }
+          let(:invited_user) { create(:user, organisation: organisation, confirmed_at: nil) }
+          let(:invited_user_email) { invited_user.email }
```
tommotaylor: This doesn't quite hit the proper flow. Here you are creating an unconfirmed user with an organisation. The real life example is that they would be invite to an org, so they would also have all the "invite" data on the user, such as `invited_at` etc.

I don't see this being an issue for now, its just not best practice. We should aim to make the testing mimic the actual user flow.

spec/features/invite_existing_user_spec.rb
cybojenix: understand now that this relates to the user being recreated. While it's not an ideal user journey, we don't expect people to use previous confirmation emails when a new one has just been sent to them

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +66,55 @@
...
+
+          it "does not send them an invite to the organisation's account" do
+            expect {
+              click_on "Send invitation email"
+            }.to change { User.count }.by(0)
+            expect(InviteUseCaseSpy.invite_count).to eq(0)
+            expect(invited_user.confirmed?).to eq(false)
+            expect(invited_user.organisation_id).to_not eq(user.organisation_id)
```
tommotaylor: Needs the error message expectation:
```
expect(page).to have_content("Email is already associated with an account. If you can't sign in, reset your password")
```

spec/features/invite_existing_user_spec.rb
Renny4Real: The user record(s) get deleted when they are sent an invite so the previous confirmation link(s) would have been lost.

spec/features/team/invite_a_team_member_spec.rb
```diff
@@ -64,14 +66,55 @@
...
+          let(:organisation) { create(:organisation) }
+          let(:invited_user) { create(:user, organisation: organisation, confirmed_at: nil) }
+          let(:invited_user_email) { invited_user.email }
+
+          it "does not send them an invite to the organisation's account" do
+            expect {
+              click_on "Send invitation email"
+            }.to change { User.count }.by(0)
+            expect(InviteUseCaseSpy.invite_count).to eq(0)
+            expect(invited_user.confirmed?).to eq(false)
```
tommotaylor: Whats the point of this line? Exists in L100 above as well. I don't think it adds anything to these specs.

app/controllers/users/invitations_controller.rb
cybojenix: would argue this should be done as part of a separate refactor, as it's a pre-existing issue for this class

spec/features/team/invite_a_team_member_spec.rb
cybojenix: While I agree that it's verbose, I disagree with your proposed solutions.

The tests that matter are around inviting the team member, so these should remain.

I think the part that should be broken out to be programmatic should be signing up the new user.
```rb
User.create(email: invited_user_email, confirmed_at: nil)
```

cybojenix: Given that "inviting team" is performed in the [outer context](https://github.com/alphagov/govwifi-admin/blob/c3d2cdb16ba0da5803f09ba35a1d9462c8e16b5f/spec/features/team/invite_a_team_member_spec.rb#L22-L25), you should be able to drop the 3 lines @tommotaylor mentioned, and just leave this `before` block for signing up a new user

app/controllers/users/invitations_controller.rb
tommotaylor: Cool, added to the sprint

PR #544, opened 2019-03-20
ensure @locations is set on the location specific controller

cybojenix:
This controller is used for the 'search again' route. Ensure we're testing it, and add locations to it.


app/controllers/logs_searches_controller.rb
```diff
@@ -33,11 +33,16 @@ def username
...
+    @locations = ordered_locations
   end
 
 private
 
   def search_params
     params.require(:logs_search).permit(:filter, :term, :first_step)
   end
+
+  def ordered_locations
```
georgeschena: I like this extraction, feels good

PR #546, opened 2019-03-20
Alphabetize organisations locations 

georgeschena:


app/controllers/admin/organisations_controller.rb
```diff
@@ -22,14 +23,25 @@ def destroy
 private
 
   def sortable_columns
-    %w[name created_at active_storage_attachments.created_at]
+    %w[name created_at active_storage_attachments.created_at address]
   end
 
   def sort_column
-    sortable_columns.include?(params[:sort]) ? params[:sort] : "created_at"
+    sortable_columns.include?(params[:sort]) ? params[:sort] : sortable_columns[1]
```
cybojenix: would be better to leave this one hard coded, as going for the second item is less intuitive than going for the first.

georgeschena: Was trying to stay consistent but you make a valid point

app/controllers/admin/organisations_controller.rb
```diff
@@ -22,14 +23,25 @@ def destroy
...
+  end
+
+  def sort_column_show
+    sortable_columns.include?(params[:sort]) ? params[:sort] : sortable_columns.last
```
cybojenix: I think all of these `sort_*` methods have been misnamed.
One set relates to MoUs, another set relates to `locations`.

They're also using a central definition of sortable columns.
What happens if I try to sort MoUs by address?
What happens if I try to sort locations by `active_storage_attachments.created_at`.

Can these be fully broken up and made clear what's going on please?

spec/features/super_admin/view_organisation_page_spec.rb
```diff
@@ -2,11 +2,16 @@
   let(:organisation) { create(:organisation) }
 
   context 'when logged in as a super-admin' do
-    let(:location) { create(:location, organisation: organisation) }
+    let(:location_1) { create(:location, organisation: organisation, address: 'Aarry Street', postcode: 'HA7 3BL') }
+    let(:location_3) { create(:location, organisation: organisation, address: 'Carry Street', postcode: 'HA7 2BL') }
```
cybojenix: parameters out of sequence, the big lint change incoming will complain about this.

cybojenix: Also, they're not actually created here, so the order of creation is dictated within the `before` block where they're referenced

app/controllers/admin/organisations_controller.rb
georgeschena: I made the change as it was more readable imo

spec/features/super_admin/view_organisation_page_spec.rb
```diff
@@ -43,20 +48,17 @@
 
     it 'shows the number of locations' do
       within('#location-count') do
-        expect(page).to have_content('1')
+        expect(page).to have_content('3')
```
cybojenix: just commenting on capybara's design here...
does it really modify `page` when entering a `within` block? I'd have expected it to pass through a fragment to the block.

:roll_eyes:

spec/features/super_admin/view_organisation_page_spec.rb
cybojenix: noting also that the postcodes are different.
If they have no effect on ordering, please keep them the same so we reduce potential scope of this test

PR #547, opened 2019-03-21
ERB lint by default

cybojenix:
rebase after #545 

Hook up the erb linter to our lint command.

This also adds an `autocorrect` command to aid with fixing issues.

Also run `autocorrect`.


app/views/admin/organisations/show.html.erb
```diff
@@ -10,7 +10,7 @@
 
     <p class="govuk-body govuk-!-margin-bottom-8">
       GovWifi organisation since <%= @organisation.created_at.strftime(
-        "#{@organisation.created_at.day.ordinalize} of %B, %Y"
+                                       "#{@organisation.created_at.day.ordinalize} of %B, %Y"
```
georgeschena: What is this massive whitespace haha?

cybojenix: looks like it's aligning it with the opening of `<%=`.
Fix for this would be to move the `<%=` to a new line.

Going to leave it as is though.

PR #549, opened 2019-03-21
Refactor invitation controller

cybojenix:
This introduces caching for querying an already invited user
We also lean more on the base controller, which means we can remove our validation and authorisation checking

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -7,6 +7,8 @@ private
 
   def find_user
     @found_user ||= User.find_by(email: invite_params[:email])
+    @found_user = User.find_by(email: invite_params[:email]) if @found_user&.destroyed?
+    @found_user
   end
 
   def delete_user_record
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -3,11 +3,14 @@ class Users::InvitationsController < Devise::InvitationsController
   before_action :delete_user_record, if: :user_should_be_cleared?, only: :create
   before_action :add_organisation_to_params, :validate_invited_user, only: :create
 
-
 private
 
+  def find_user
+    @found_user ||= User.find_by(email: invite_params[:email])
+  end
+
   def delete_user_record
-    User.find_by(email: invite_params[:email]).destroy!
+    find_user.destroy!
   end
 
   def resending_invite?
@@ -44,15 +47,15 @@ private
   end
 
   def invited_user_already_exists?
-    !!User.find_by(email: invite_params[:email])
+    !!find_user
   end
 
   def invited_user_not_confirmed?
-    !User.find_by(email: invite_params[:email]).confirmed?
+    !find_user.confirmed?
   end
 
   def invited_user_has_no_org?
-    User.find_by(email: invite_params[:email]).organisation_id.nil?
+    find_user.organisation_id.nil?
   end
 
   # Overrides https://github.com/scambra/devise_invitable/blob/master/app/controllers/devise/invitations_controller.rb#L105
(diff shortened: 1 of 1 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -1,13 +1,25 @@
 class Users::InvitationsController < Devise::InvitationsController
-  before_action :authorise_manage_team, only: %i(create new)
   before_action :delete_user_record, if: :user_should_be_cleared?, only: :create
-  before_action :add_organisation_to_params, :validate_invited_user, only: :create
-
+  before_action :add_organisation_to_params, only: :create
 
 private
 
+  def authenticate_inviter!
```
tommotaylor: I'm guessing this is a Devise method that we are overriding. It would be good to link directly to the definition of this method in a comment, similar how we've done it on L53 below.

app/controllers/users/invitations_controller.rb
```diff
@@ -1,13 +1,25 @@
...
+
   def delete_user_record
-    User.find_by(email: invite_params[:email]).destroy!
+    invited_user.destroy!
+  end
+
+  def invited_user
+    @invited_user ||= User.find_by(email: invite_params[:email])
+    @invited_user = User.find_by(email: invite_params[:email]) if @invited_user&.destroyed?
+    @invited_user
```
tommotaylor: L21 & L22 not required.

app/controllers/users/invitations_controller.rb
```diff
@@ -1,13 +1,25 @@
 class Users::InvitationsController < Devise::InvitationsController
-  before_action :authorise_manage_team, only: %i(create new)
   before_action :delete_user_record, if: :user_should_be_cleared?, only: :create
-  before_action :add_organisation_to_params, :validate_invited_user, only: :create
-
+  before_action :add_organisation_to_params, only: :create
```
tommotaylor: We've lost `before_action validate_invited_user, only: :create`. Where is the user being validated now?

cybojenix: Validation is run as part of the invite steps, so part of the base `Devise::InvitiationsController`

app/controllers/users/invitations_controller.rb
cybojenix: aye.. This was me trying to avoid some unexpected behaviour as we do have a destroy step in here.

Will remove

app/controllers/users/invitations_controller.rb
tommotaylor: 👍

app/controllers/users/invitations_controller.rb
```diff
@@ -1,29 +1,44 @@
...
+    respond_with_navigational(resource) { render :new }
   end
 
-  def after_invite_path_for(_resource)
-    team_members_path
+  def user_is_invalid?
+    # This is an indirect solution to preventing a user being re-invited when they belong
+    # to another organisation.
+    self.resource = resource_class.new(invite_params)
+    resource.invalid?
```
tommotaylor: I preferred the old syntax where it was explicitly testing a user. The `resource` abstraction is a Devise one because Devise allows you to name your user model whatever you like. We use `User` so I think we should be explicit. Eg:
```
  @user = User.new(invite_params)
  !@user.validate
```
or `@user.invalid?` as you've used.

app/controllers/users/invitations_controller.rb
```diff
@@ -1,29 +1,44 @@
 class Users::InvitationsController < Devise::InvitationsController
-  before_action :authorise_manage_team, only: %i(create new)
```
tommotaylor: Where are we doing this now that its removed?

cybojenix: this is part of `authenticate_inviter!`, as documented below

PR #551, opened 2019-03-21
ensure we bind the IP address when serving in development

cybojenix:
a recent change to puma means the default bind is 'localhost' instead of '0.0.0.0' for development.
This breaks when running rails inside docker.

move to using '0.0.0.0' as default. This does not affect hosted environments, which already defaults to '0.0.0.0'


cybojenix:
cause for this was due to a previous bug in puma: https://github.com/puma/puma/pull/1700

PR #552, opened 2019-03-21
change to using sassc-rails

cybojenix:
This is [next in line](https://github.com/rails/sass-rails/pull/424) for replacing sass-rails following the `sass` gem deprecation, and is a drop in replacement.

note that sass is still installed due to govuk-lint, however rails will now use sassc


cybojenix:
This takes my initial page render from 27s to 8s

cybojenix:
It looks like our development asset pipeline behaves differently to test/staging/production.

By the looks of it, it's preprocessing the sass twice, but not in development.

Closing this until we can do further investigation.

PR #568, opened 2019-03-25
Lint the upload download mou spec

tommotaylor:
Lint `spec/features/upload_download_mous_spec.rb`

Extract the super admin context into `spec/features/super_admin/upload_download_mou_spec.rb`


spec/features/super_admin/upload_download_mou_spec.rb
```diff
@@ -0,0 +1,42 @@
+describe 'Upload and download the MOU template', type: :feature do
```
cybojenix: both of these titles differ. Can they be aligned please?

PR #561, opened 2019-03-22
Allow super admin to view logs

georgeschena:
This PR gives a super admin the exposure to search logs  by a particular username to allow the super admin to diagnose issues with users connecting to GovWifi.

PR #570, opened 2019-03-25
Allow super admin to search logs with an IP

georgeschena:
This PR builds upon #561 and allows a super admin to search the logs by a particular IP to allow the super admin to diagnose issues with users connecting to GovWifi.


spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+    Session.create!(
+      start: 3.days.ago,
+      username: username,
+      siteIP: ip,
+      success: true
+    )
+
+    create(:ip, location_id: location.id, address: ip)
```
tommotaylor: This ip creation seems a bit convoluted to me. You are doing something like:
- setting a variable which has an ip address: `let(:ip) { '1.2.3.4' }`
- creating three sessions referencing that variable Session.create!(siteIP: ip)`
- then creating an `Ip` with that variable: `create(:ip, location_id: location.id, address: ip)`

You could do this like:
- Create an Ip: `let(:ip) { create(:ip) }`
- creating three sessions referencing that variable Session.create!(siteIP: ip.address)`

Less code, more logical and much easier to follow

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    context "with a correct IP" do
+      let(:search_string) { ip }
+
+      it 'displays all results for the searched IP' do
+        expect(page).to have_content("Found 3 results for \"#{ip}\"")
+      end
+    end
+
+    context 'when a user from an organisation searches logs for an IP from another organisation' do
```
tommotaylor: This doesn't make sense to me, this currently reads like:
```
in `/super_admin/logging/`
'View authentication requests for an IP'
  'when searching for an IP address'
    'when a user from an organisation searches logs for an IP from another organisation'
      'displays all results for the searched IP'
```
When reading the nested contexts all in a row it should make sense. This is actually the output you see when you run `rspec` in the command line.

I'm confused by:
- the fact this is a user without admin privileges, yet its nested in the same context as a signed in super admin user and its in the super_admin folder.
- The wording
- Also it suggests that a non-super admin user should be able to see logs from another organisation which is incorrect.

I would suggest that this is better in the non-super admin spec. As an idea something like:
```
in `specs/features/logging/view_auth_requests_for_an_ip_spec.rb`
"when searching for an IP address"
  "when entering an IP address that belongs to another organisation"
    "does not display any logs"
```

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      let(:search_string) { ip }
+
+      before do
+        sign_in_user other_user
+        visit ip_new_logs_search_path
+        fill_in "IP address", with: search_string
+        click_on "Show logs"
+      end
+
+      it 'displays all results for the searched IP' do
```
tommotaylor: incorrect statement

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      let(:search_string) { "1" }
+
+      it_behaves_like "errors in form"
+
+      it "displays an error summary" do
+        expect(page).to have_content("Search term must be a valid IP address")
+      end
+
+      it "prompts to the user for a valid IP address" do
+        expect(page).to have_content("Enter a valid IP address")
```
tommotaylor: Is this necessary? What does it test?

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
georgeschena: I need to set an IP with an address as I need to search for that IP

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  let(:user_2) { create(:user, organisation: user_2_organisation) }
+  let(:user_2_organisation) { create(:organisation) }
+  let(:user_2_location) { create(:location, organisation: user_2.organisation) }
+
+  let(:super_admin) { create(:user, super_admin: true) }
+  let(:super_admin_organisation) { create(:organisation) }
+  let(:super_admin_location) { create(:location, organisation: super_admin.organisation) }
+
+  let(:ip_1) { create(:ip, location_id: user_location.id, address: '1.2.3.4') }
```
cybojenix: nit: can these be grouped with their respective collections? (so with `user` and `user_2` respectively)

georgeschena: Will do

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,78 @@
+describe 'View authentication requests for an IP', type: :feature do
+  let(:user) { create(:user, organisation: user_organisation) }
+  let(:user_organisation) { create(:organisation) }
```
tommotaylor: These two can be combined to one as the user factory creates an org with the user. See `factories/users.rb`.
So it becomes:
```
let(:user) { create(:user)
```

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  let(:user_location) { create(:location, organisation: user.organisation) }
+  let(:ip_1) { create(:ip, location_id: user_location.id, address: '1.2.3.4') }
+
+  let(:user_2) { create(:user, organisation: user_2_organisation) }
+  let(:user_2_organisation) { create(:organisation) }
+  let(:user_2_location) { create(:location, organisation: user_2.organisation) }
```
tommotaylor: Why not just `user_2`, `location_2`, `organisation_2`? Simpler and more understandable

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  let(:ip_2) { create(:ip, location_id: user_2_location.id, address: '1.2.3.5') }
+
+  let(:super_admin) { create(:user, super_admin: true) }
+  let(:super_admin_organisation) { create(:organisation) }
+  let(:super_admin_location) { create(:location, organisation: super_admin.organisation) }
```
tommotaylor: this isn't used, can probably be deleted

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  let(:user_location) { create(:location, organisation: user.organisation) }
+  let(:ip_1) { create(:ip, location_id: user_location.id, address: '1.2.3.4') }
+
+  let(:user_2) { create(:user, organisation: user_2_organisation) }
+  let(:user_2_organisation) { create(:organisation) }
+  let(:user_2_location) { create(:location, organisation: user_2.organisation) }
+  let(:ip_2) { create(:ip, location_id: user_2_location.id, address: '1.2.3.5') }
+
+  let(:super_admin) { create(:user, super_admin: true) }
+  let(:super_admin_organisation) { create(:organisation) }
```
tommotaylor: This is unnecessary as line 12 creates an organisation. See previous comment.

tommotaylor: Also not used

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    Session.create!(
+      start: 3.days.ago,
+      siteIP: ip_2.address,
+    )
+
+    sign_in_user super_admin
+    visit ip_new_logs_search_path
+  end
+
+  context "when searching for an IP from a location" do
```
tommotaylor: "when searching for an IP from a location" isn't what is being tested here. The " from a location part" seems unnecessary.

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      fill_in "IP address", with: ip_1.address
+      click_on "Show logs"
+    end
+
+    it 'displays all found results for that searched IP' do
+      expect(page).to have_content("Found 2 results for \"#{ip_1.address}\"")
+    end
+  end
+
+  context "when looking for an IP from a different location" do
```
tommotaylor: Why is the " from a different location" important? We aren't searching locations here, just IPs.

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
tommotaylor: Also `let(:user_2_organisation) { create(:organisation) }` not required

spec/features/super_admin/logging/view_logs_for_ip_spec.rb
```diff
@@ -0,0 +1,73 @@
+describe 'View authentication requests for an IP', type: :feature do
+  let(:user) { create(:user) }
+  let(:user_location) { create(:location, organisation: user.organisation) }
+  let(:ip_1) { create(:ip, location_id: user_location.id, address: '1.2.3.4') }
```
tommotaylor: Use ActiveRecords association helpers, not ids. Eg: `location: user_location` not `location_id: user_location.id`

tommotaylor: Also worth checking other places you've used it in this spec

PR #578, opened 2019-03-27
Super admin dynamic MoU upload

georgeschena:
Before MoU uploaded:
Screenshot 2019-03-27 at 15 07 35
After MoU uploaded: 
Screenshot 2019-03-27 at 15 57 37


Renny4Real:
Perhaps change `Upload Template` bold text to be `Replace Template` when an MOU template has been submitted.

PR #582, opened 2019-03-28
set up paths + redirects to help with building flows in analytics

cybojenix:


spec/features/ips/add_ip_to_existing_location_spec.rb
```diff
@@ -22,6 +22,10 @@
         it 'adds to the correct location' do
           expect(location.reload.ips.map(&:address)).to include("141.0.149.130")
         end
+
+        it 'redirects to the "after IP created" path for Analytics' do
```
georgeschena: I feel mentioning "for Analytics" is not really needed, this change is also good for best practice not just for GA. But i'm open to convincing.

cybojenix: I think for "best practice" it doesn't really make any difference as in generally we don't really care unless we're wanting to make multi-step journeys.

The sole reason this exists is to improve our Analytics. This makes it so the test explains why the feature is in here, as apposed to just stating it's functionality.

Happy to change to another way of explaining in the tests why we need it though

georgeschena: Im convinced

PR #583, opened 2019-03-28
Allow super admin to invite user to any organisation

georgeschena:
- [x] Change redirect after adding user to an organisation
- [x] don't display hidden filed when not super admin
- [x] Info around what team your adding too
- [x] Change back to list link

Screenshot 2019-04-02 at 12 05 51

Screenshot 2019-04-02 at 12 06 01

Screenshot 2019-04-02 at 12 06 13


app/controllers/users/invitations_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Users::InvitationsController < Devise::InvitationsController
   before_action :delete_user_record, if: :user_should_be_cleared?, only: :create
   before_action :return_user_to_invite_page, if: :user_is_invalid?, only: :create
-  before_action :add_organisation_to_params, only: :create
+  before_action :add_organisation_to_params, unless: :super_admin?, only: :create
 
 private
```
cybojenix: can `authenticate_inviter!` (below this line) be explicit about allowing super admins please?
being a super_admin doesn't automatically give the permission to manage a team, which will lead to some unusual potentially hard to diagnose issues down the line

app/controllers/users/invitations_controller.rb
```diff
@@ -38,7 +38,11 @@ def resending_invite?
   end
 
   def after_invite_path_for(_resource)
-    resending_invite? ? recreated_invite_team_members_path : created_invite_team_members_path
+    if super_admin?
+      admin_organisation_path(invite_params[:organisation_id])
```
cybojenix: do we need to set up analytics paths for super admins too?

app/views/users/invitations/new.html.erb
```diff
@@ -2,19 +2,27 @@
...
-    <h2 class="govuk-heading-l">
-      Invite a team member
-    </h2>
-
-    <div class="govuk-details">
-      <div class="govuk-details__text">
-        <p class="govuk-body">
-          Team members will receive an email invitation to join your organisation's GovWifi Admin dashboard.
-        </p>
+    <% if current_user.super_admin? %>
```
cybojenix: can we use the method in the controller for this?

app/views/users/invitations/new.html.erb
```diff
@@ -2,19 +2,27 @@
...
+      <%= link_to "Back to organisation", admin_organisation_path(params[:id]), class: "govuk-back-link" %>
+      <h2 class="govuk-heading-l">Invite a team member to <%= params[:name] %> </h2>
```
cybojenix: can the target organisation be exposed as an object in the controller? This will allow for not needing to pass the org name around in the path

spec/features/super_admin/invite_user_to_organisation_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'support/invite_use_case'
+require 'support/notifications_service'
+
+describe "Inviting a team member as a super admin", type: :feature do
```
cybojenix: I think there are some missing tests here, most can be copied from the admin invitation tests:
I think some of them will also be `request` tests instead of `feature` tests too.

```
without super admin privs, it ignores the provided parameters
with invalid email, it has the same behaviour as invalid email handling in admin invitations
```

What's the behaviour when tampering with parameters, (unset, set to non-existant org).
What's the behaviour of a re-invitation? Can a super-admin send a re-invitation?
If not, assert that it can't.

There's lots of undefined behaviour here.

app/controllers/users/invitations_controller.rb
cybojenix: Thinking about this further, we've been talking about read-only super admins.
I'll leave it as is for now, but may consider adding a test to make it explicit a super admin without the team management perm can't invite team members.

app/controllers/users/invitations_controller.rb
cybojenix: Confirmed we do not need this yet

spec/features/super_admin/invite_user_to_organisation_spec.rb
cybojenix: I've expanded on the tests, however in the future we should look at using a shared example to test common usecases when branching is involved

app/controllers/users/invitations_controller.rb
```diff
@@ -61,6 +67,18 @@ def invited_user_has_no_org?
...
 
+  def super_admin?
+    current_user.super_admin?
+  end
+
+  def target_organisation
+    @target_organisation ||= if super_admin?
+                               Organisation.find_by(id: params[:id] || invite_params[:organisation_id])
+                             else
+                               current_user.organisation
```
tommotaylor: This structure handles both super_admin and non super_admin users, however the method `target_organisation` is only used for super admins.

Makes me think that having a `helper_method` is one extraction too far where a simple instance variable `@target_organisation` would do.

spec/features/super_admin/invite_user_to_organisation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  before do
+    sign_in_user super_admin
+    visit admin_organisation_path(organisation)
+    click_on 'Add team member'
+    fill_in 'Email address', with: email
+  end
+
+  include_examples 'notifications service'
+  include_examples 'invite use case spy'
+
```
tommotaylor: Whitespace

spec/features/super_admin/invite_user_to_organisation_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    it "does not send an invite" do
+      expect(InviteUseCaseSpy.invite_count).to eq(0)
+    end
+
+    it "displays the correct error message" do
+      expect(page).to have_content("Email can't be blank")
+    end
+
+    context "when retying with a valid email address" do
```
tommotaylor: spelling

app/controllers/users/invitations_controller.rb
cybojenix: moving to using an instance variable would require changes at 2 points.

This contains the branching to one place.

If you'd like to still go down the instance variable route, can you recommend a way, as I'm getting odd behaviour when modifying the `new` and `create` controllers

app/controllers/users/invitations_controller.rb
```diff
@@ -22,6 +25,10 @@ def return_user_to_invite_page
     respond_with_navigational(resource) { render :new }
   end
 
+  def set_target_organisation
+    @target_organisation = Organisation.find_by(id: params[:id] || invite_params[:organisation_id])
```
tommotaylor: This doesn't need to be `find_by(id: ID)`, if you are doing a lookup by id then `.find(ID)` is better.

PR #584, opened 2019-03-28
Fix whitelist organisation error

Renny4Real:
**WHY:**
When a user doesn't select an organisation from the drop down list of whitelisted organisations and tries to finish creating their account at the confirmation stage. The page would display an error that `Organisation base isn't a whitelisted organisation`. 

This is a Rails error message that hasn't displayed the whitelist error as intended/desired.

The method by which this error message is added should be changed so that a better worded message would be shown to the user.

This is one approach to making the errors we display to users more readable and understandable.
Another approach can be seen in PR #586

**IN THIS PR:**
- Change the method from adding the error to `:base` to adding it to the `:name` attribute.

- Edit the select functionality in the signup helper method: `update_user_details` to be conditional. This needed to be done as the capybara would error if it couldn't select anything from the drop down and testing the display of the whitelist error message depends on no selection being made before submission.

**BEFORE:**

Screenshot 2019-03-28 at 11 30 44

**AFTER:**

Screenshot 2019-03-28 at 11 31 29

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -10,7 +10,7 @@ class Organisation < ApplicationRecord
 
   def validate_in_register?
     unless Organisation.fetch_organisations_from_register.include?(name)
-      errors.add(:base, "#{name} isn't a whitelisted organisation")
+      errors.add(:name, "isn't whitelisted")
     end
   end
 
--- a/spec/features/registration/sign_up_as_an_organisation_spec.rb
+++ b/spec/features/registration/sign_up_as_an_organisation_spec.rb
@@ -188,4 +188,16 @@ describe 'Sign up as an organisation' do
       expect(page).to have_content("Email is already associated with an account. If you can't sign in, reset your password")
     end
   end
+
+  context 'when no whitelisted organisation is selected' do
+    let(:non_whitelist_org_name) {'Not Whitelisted'}
+
+    before { sign_up_for_account }
+
+    it 'will tell the user the organisation is not whitelisted' do
+      update_user_details(organisation_name: non_whitelist_org_name)
+      expect(page).to have_content("name isn't whitelisted")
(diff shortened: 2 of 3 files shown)


app/models/organisation.rb
```diff
@@ -10,7 +10,7 @@ class Organisation < ApplicationRecord
 
   def validate_in_register?
     unless Organisation.fetch_organisations_from_register.include?(name)
```
cybojenix:
```suggestion
    unless !name.blank? && Organisation.fetch_organisations_from_register.include?(name)
```

cybojenix: might also be able to do this instead:

```rb
validate :validate_in_register?, unless: -> org { org.name.blank? }
```

Renny4Real: Will try this in a separate PR 👍

spec/features/registration/sign_up_as_an_organisation_spec.rb
```diff
@@ -188,4 +188,16 @@
...
     end
   end
+
+  context 'when no whitelisted organisation is selected' do
+    let(:non_whitelist_org_name) {'Not Whitelisted'}
+
+    before { sign_up_for_account }
+
+    it 'will tell the user the organisation is not whitelisted' do
+      update_user_details(organisation_name: non_whitelist_org_name)
```
cybojenix: this amounts to not actually selecting an organisation name, as apposed to trying to set `Not Whitelisted` as the name.
This means the test is passing due to an indirect reason (blank is not whitelisted).

This is due to the conditional added in `sign_up_helpers`.
`expect(page).not_to have_content("Organisation name can't be blank")` will fail if this is the case.

Renny4Real:
Closing in favour of PR #586

PR #586, opened 2019-03-28
Don't display whitelist error when no organisation is selected

Renny4Real:
**WHY:**
When a user doesn't select an organisation from the drop down list of whitelisted organisations and tries to finish creating their account at the confirmation stage. The page would display both these errors: 
- `Organisation name can't be blank`
- `Organisation base isn't a whitelisted organisation`

The organisation name field can only be filled by selecting from the drop-down menu of whitelisted organisations.

The field does not receive/pass any user text inputted directly.

So an instance of a user 'filling in' the field with a name not on the list and an instance of a user leaving the field blank result in the same thing: 
- No organisation is selected and so, no name is received for the organisation name value.
- The subsequent validation of the name value (based on presence) would lead to the `Organisation name can't be blank` error.

Displaying both errors is unneccessary when no organisation has been selected from the list.

**IN THIS PR:**
The `Organisation name can't be blank` error is the only error that would display, if no organisation is selected since the field is not filled unless you select from the list

This is one approach to making the errors we display to users more readable and understandable.

The user should not need to see a whitelist error since there is already an instruction for users on what to do if their organisation is not on the list.

Another approach could be seen in PR: #584 

**BEFORE:**

Screenshot 2019-03-28 at 12 27 24

**AFTER:**

Screenshot 2019-03-28 at 12 27 36

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -6,7 +6,7 @@ class Organisation < ApplicationRecord
 
   validates :name, presence: true, uniqueness: { case_sensitive: false }
   validates :service_email, presence: true
-  validate :validate_in_register?
+  validate :validate_in_register?, unless: Proc.new { |org| org.name.blank? }
 
   def validate_in_register?
     unless Organisation.fetch_organisations_from_register.include?(name)
--- a/spec/features/registration/sign_up_as_an_organisation_spec.rb
+++ b/spec/features/registration/sign_up_as_an_organisation_spec.rb
@@ -188,4 +188,24 @@ describe 'Sign up as an organisation' do
       expect(page).to have_content("Email is already associated with an account. If you can't sign in, reset your password")
     end
   end
+
+  context 'when no organisation has been selected' do
+    let(:org_name_left_blank) { '' }
+
+    before { sign_up_for_account }
+
+    it 'will tell the user the organisation name cannot be left blank' do
+      update_user_details(organisation_name: org_name_left_blank)
+      within('div#error-summary') do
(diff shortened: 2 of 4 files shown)


spec/features/registration/sign_up_as_an_organisation_spec.rb
```diff
@@ -188,4 +188,24 @@
...
+    before { sign_up_for_account }
+
+    it 'will tell the user the organisation name cannot be left blank' do
+      update_user_details(organisation_name: org_name_left_blank)
+      within('div#error-summary') do
+        expect(page).to have_content("Organisation name can't be blank")
+      end
+    end
+
+    it 'will not display a non-whitelist error for a blank entry' do
```
tommotaylor: Not sure if this is the best way to test this. Its kind of like saying, "make sure a page doesn't show some totally irrelevant content". It might be better to say something like `it "only displays one error message"` and just check that the `"Organisation name can't be blank"` is the only error message displayed.

Does that make sense?

tommotaylor:
I prefer this approach, much cleaner than the other approach @Renny4Real

app/models/organisation.rb
```diff
@@ -6,7 +6,7 @@ class Organisation < ApplicationRecord
 
   validates :name, presence: true, uniqueness: { case_sensitive: false }
   validates :service_email, presence: true
-  validate :validate_in_register?
+  validate :validate_in_register?, unless: Proc.new { |org| org.name.blank? }
```
cybojenix: Can this be tested in the model tests please?

PR #593, opened 2019-04-01
Alphabetise team list in super admin

tommotaylor:
Alphabetising the list on the "show organisation page" in super admin. Must have been missed off the original task.

Involves a refactor using Clean Architecture to DRY up the code. Not 100% sure this is the right way to do it with CA so welcome feedback.

Screenshot 2019-04-01 at 11 38 03

### DIFF
--- a/spec/gateways/organisation_users_spec.rb
+++ b/spec/gateways/organisation_users_spec.rb
@@ -15,6 +15,10 @@ describe Gateways::OrganisationUsers do
     it 'fetches the users for the organisation' do
       expect(gateway.fetch).to eq(result)
     end
+
+    it 'returns an ActiveRecord collection' do
+      expect(gateway.fetch.class.ancestors).to include(ActiveRecord::Relation)
+    end
   end
 
   context 'without users' do
--- a/spec/gateways/organisation_users_spec.rb
+++ b/spec/gateways/organisation_users_spec.rb
@@ -1,0 +1,27 @@
+describe Gateways::OrganisationUsers do
+  subject(:gateway) { described_class.new(organisation: organisation) }
+
+  context 'with users' do
+    let(:organisation) { create(:organisation) }
+    let(:result) do
+      organisation.users
+    end
+
+    before do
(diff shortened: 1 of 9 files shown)


spec/gateways/organisation_users_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    it 'fetches the users for the organisation' do
+      expect(gateway.fetch).to eq(result)
+    end
+  end
+
+  context 'without users' do
+    let(:organisation) { create(:organisation) }
+
+    it 'returns an empty collection' do
+      expect(gateway.fetch).to eq([])
```
cybojenix: We're expecting this gateway to always yield an active record collection. Can it be tested as such please?

I think we're also missing a test to assert that further queries can be run.
I'm unsure on the "rails way" for this, however I think asserting that it's an `ActiveRecord::Collection` should suffice

spec/use_cases/administrator/sort_users_spec.rb
```diff
@@ -0,0 +1,26 @@
+describe UseCases::Administrator::SortUsers do
```
cybojenix: can there also be a test to ensure values fetched from the database get sorted?
This allows us to assert ordering outside of a feature test

tommotaylor: I guess what you are talking about is the integration test that ties this all together. It currently exists in a few feature specs.

spec/gateways/organisation_users_spec.rb
tommotaylor: Agree to a point. Heres the behaviour:
```
gateway.fetch
=> []
```
```
gateway.fetch.class
=> User::ActiveRecord_Associations_CollectionProxy
```

spec/use_cases/administrator/sort_users_spec.rb
tommotaylor: Done @cybojenix

PR #597, opened 2019-04-01
[WIP] Spike manage super admin privileges Option 1

tommotaylor:
Spike for allowing GDS to invite a super admin. 

Screenshot 2019-04-01 at 16 54 28

Other options include:
- Option 2: Doing "super admin" permissions on a team, eg: "GovWifi Super Administrators". If you belong to this team you are a super admin. View here: https://github.com/alphagov/govwifi-admin/pull/598
OR
- Option 1b:Adding a "super admin" checkbox to an edit permissions page, eg:
Screenshot 2019-04-01 at 17 11 54
Screenshot 2019-04-01 at 17 13 04
OR
Screenshot 2019-04-01 at 17 14 18

PR #598, opened 2019-04-01
Move super admin privileges to the organisation

tommotaylor:
Moves the `super_admin` boolean to organisation. So now to invite a super admin, the super admin would just need to invite them to the "GovWifi Super Administrators" organisation.

Screenshot 2019-04-02 at 14 20 09

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -2,6 +2,6 @@ class AdminController < ApplicationController
   before_action :authorise_admin
 
   def authorise_admin
-    redirect_to root_path unless current_user.super_admin?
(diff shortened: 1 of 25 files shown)


Renny4Real:
I feel like this a cleaner approach to the task. 
Allowing a super admin to be regarded as an organisation could make it easier to work with for future features. 
PR #597 (the alternative version) could also mean that the `revoke/grant super admin links` would be attached to every user in every organisation in the super admin view. 
If by accident, a link was pressed unintentionally, it might be difficult to track the user and increase the risk of information exposure.
It could also increase the ambiguity surrounding which organisation a super admin can belong to, although this might be unlikely.

PR #603, opened 2019-04-03
GovWifi map setup

georgeschena:


spec/features/super_admin/view_map_of_locations_for_govwifi_spec.rb
```diff
@@ -0,0 +1,14 @@
+describe 'Signing in as a super admin', type: :feature do
+  let(:user) { create(:user, :super_admin) }
+
+  context 'when viewing the GovWifi map' do
+    before do
+      sign_in_user user
+      visit admin_govwifi_map_index_path
+    end
+
+    it 'shows list of signed organisations on the home page' do
```
cybojenix: the assertion in this test doesn't align with what's being test

Gemfile
```diff
@@ -11,6 +11,7 @@ gem 'aws-sdk-route53', '~> 1.21.0'
 gem 'aws-sdk-s3', '~> 1'
 gem 'devise', '~> 4.6.2'
 gem 'devise_invitable', '~> 2.0.0'
+gem 'geocoder', '~> 1.5', '>= 1.5.1'
```
cybojenix: This dependency isn't used yet, Can you introduce the dependency when you come to implement the code please?

PR #604, opened 2019-04-03
Track new organisation first IP creation pathway v2.0

Renny4Real:
**WHY:**
_This PR is version 2 of second step to allowing GA to track first IP creation activity paths. PR #602_

First step is PR #595 

A new organisation is an organisation with no IPs set up regardless of adding other things e.g location, team members.

We want to follow a new organisation's conversion i.e adding their first IP address and thus making their installation of GovWifi count as an active one.

We need to follow the difference between new and existing organisations through the "create IP" pathway.

**IN THIS PR:**
- Add data layer script with `signed_in` and `hasIPs` variables:

**_Added  a `signed_in` variable because `hasIPs` is false when the user is not signed in._** 

Screenshot 2019-04-03 at 14 56 06

**_`hasIPs` is false when the organisation has no IPs_** 

Screenshot 2019-04-03 at 14 57 09

**_`hasIPs` is false when the organisation has no IPs (even if locations are added)_** 

Screenshot 2019-04-03 at 14 55 33

**_`hasIPs` is true when the organisation had an IP address_** 

Screenshot 2019-04-03 at 14 55 56

**Any feedback or suggestions on this approach would be appreciated**

### DIFF
--- a/app/views/layouts/_google_analytics.html.erb
+++ b/app/views/layouts/_google_analytics.html.erb
@@ -3,8 +3,8 @@
 
   <script>
     dataLayer = [{
-      'signed_in': '<%= user_signed_in?.to_json %>'
-      'hasIPs': '<%= (!current_user.nil? && current_organisation.ips.any?).to_json %>'
+      'signedIn': '<%= user_signed_in?.to_json %>'
+      'hasIPs': '<%= (!!current_user && current_organisation.ips.any?).to_json %>'
     }];
   </script>
 
--- a/app/views/layouts/_google_analytics.html.erb
+++ b/app/views/layouts/_google_analytics.html.erb
@@ -3,7 +3,7 @@
 
   <script>
     dataLayer = [{
-      'signedIn': '<%= user_signed_in?.to_json %>'
+      'signedIn': '<%= user_signed_in?.to_json %>',
       'hasIPs': '<%= (!!current_user && current_organisation.ips.any?).to_json %>'
     }];
   </script>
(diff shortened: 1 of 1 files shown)


app/views/layouts/_google_analytics.html.erb
```diff
@@ -1,6 +1,13 @@
 <!-- Global site tag (gtag.js) - Google Analytics -->
 <script async src="https://www.googletagmanager.com/gtag/js?id=<%= SITE_CONFIG['google_analytics_script_id'] %>"></script>
 
+  <script>
+    dataLayer = [{
+      'signed_in': '<%= user_signed_in?.to_json %>'
+      'hasIPs': '<%= (!current_user.nil? && current_organisation.ips.any?).to_json %>'
```
tommotaylor: You won't need either of the `.to_json` as `false.to_json` => 'false' and  '<%= false %>' => 'false'.

also `!current_user.nil?` can just be written as `!!current_user` as `!!nil` => `false`

Would also use `signedIn` javascript style rather than `signed_in` ruby style considering its js.

Renny4Real: Added `.to_json` because of the ERB linter.

tommotaylor: also needs a comma in between the two hash key value pairs.

Renny4Real:
UPDATE: Changed `signed_in` to `signedIn` to match javascript naming style.

Screenshot 2019-04-03 at 16 02 30

PR #605, opened 2019-04-03
Lint specs in models part one

Renny4Real:
**WHY:**

**IN THIS PR:**
Lint and refactor the following files in the `spec/models folder`:

- `spec/models/logs_search_spec.rb`
- `spec/models/ip_spec.rb`
- `spec/models/location_spec.rb`

### DIFF
--- a/spec/models/ip_spec.rb
+++ b/spec/models/ip_spec.rb
@@ -2,40 +2,37 @@ describe Ip do
   subject(:ip_address) { described_class.new }
 
   it { is_expected.to belong_to(:location) }
+  it { is_expected.to validate_presence_of(:address) }
+  it { is_expected.to validate_uniqueness_of(:address) }
 
-  context "with validations" do
-    it { is_expected.to validate_presence_of(:address) }
-    it { is_expected.to validate_uniqueness_of(:address) }
-  end
-
-  context "with address" do
+  context "when validating" do
     let(:location) { create(:location, organisation: create(:organisation)) }
 
-    context "when invalid" do
+    it 'does not allow the address 0.0.0.0' do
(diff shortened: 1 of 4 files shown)


spec/models/ip_spec.rb
```diff
@@ -1,51 +1,56 @@
 describe Ip do
+  subject(:ip_address) { described_class.new }
+
   it { is_expected.to belong_to(:location) }
+  it { is_expected.to validate_presence_of(:address) }
+  it { is_expected.to validate_uniqueness_of(:address) }
+
+  context "when validating" do
```
tommotaylor: "when validating address" might be better. We are only validating the address in this context.

spec/models/ip_spec.rb
```diff
@@ -1,51 +1,56 @@
...
+        expect(ip.errors.full_messages).to eq([
+          "Address 'invalidIP' is not valid"
+        ])
+      end
+    end
+
+    context "with a valid address" do
+      let!(:ip) { described_class.create(address: "141.0.149.130", location: location) }
+
+      it "saves the address" do
```
tommotaylor: Actually it saves the `Ip`, not the `address`. `address` is an attribute on the `Ip` model.

spec/models/location_spec.rb
```diff
@@ -68,9 +62,14 @@
           "1" => ActionController::Parameters.new(address: 'wrong')
         ).permit!
       )
+    }
 
-      expect(location.save).to be_falsey
+    it 'is invalid' do
```
tommotaylor: Needs to be more descriptive for instance "does not create the location" or "does not pass validation"

PR #609, opened 2019-04-04
Lint remaining specs in request folder

Renny4Real:
**WHY:**

**IN THIS PR:**
Lint and refactor the following files in the `spec/requests` folder:

- `spec/requests/ips/delete_spec.rb`
- `spec/requests/email_domain/delete_spec.rb`
- `spec/requests/monitoring/healthcheck_spec.rb`


spec/requests/email_domain/delete_spec.rb
```diff
@@ -1,5 +1,11 @@
 describe "DELETE /authorised_email_domains/:id", type: :request do
   let!(:email_domain) { create(:authorised_email_domain, name: 'some.domain.org.uk') }
+  let(:gateway_instance) { instance_double(Gateways::S3) }
+
+  before do
+    allow(Gateways::S3).to receive(:new).and_return(gateway_instance)
+    allow(gateway_instance).to receive(:write)
```
tommotaylor: Having this `allow(gateway_instance).to receive(:write)` and also `expect(gateway_instance).to have_received(:write)` on line 24 seems odd. Like we are telling rspec to do something and then testing it did it.

spec/requests/ips/delete_spec.rb
```diff
@@ -11,15 +11,22 @@
   end
 
   context "when the user owns the IP" do
+    let(:publish_instance) { instance_double(Facades::Ips::Publish) }
+
+    before do
+      allow(Facades::Ips::Publish).to receive(:new).and_return(publish_instance)
+      allow(publish_instance).to receive(:execute)
```
tommotaylor: Same comment as above.

spec/requests/email_domain/delete_spec.rb
tommotaylor: A better approach would be something like:

L3: `let(:gateway) { instance_spy(Gateways::S3, write: nil) }`
Delete L7
L24: `expect(gateway).to have_received(:write)`

PR #610, opened 2019-04-04
Run PRs in Concourse

cybojenix:
This sets up the PR pipeline, and part of the deployment pipeline for Concourse.

The PR pipeline is currently running on the GDS concourse, as well as local concourse environments.

While a deploy pipeline does exist, it currently is not set up to run yet.
This will occur at a later time.


Makefile
```diff
@@ -1,17 +1,30 @@
 DOCKER_COMPOSE = docker-compose -f docker-compose.yml
 BUNDLE_FLAGS=
+
 ifdef DEPLOYMENT
   BUNDLE_FLAGS = --without test development
 endif
 
 ifndef JENKINS_URL
```
emileswarts: Should we just remove all references to Jenkins?
Are we keeping this backwards compatible as we make the switch to concourse?

ci/pipelines/deploy.yml
```diff
@@ -0,0 +1,78 @@
+resources:
+  - name: govwifi-admin
+    type: git
+    source:
+      uri: "https://github.com/alphagov/govwifi-admin.git"
+      branch: "concourse"
+
+  - name: ecr
+    # This needs to be docker-image, until registry-image supports ECR
```
emileswarts: Not entirely sure what this comments means.
ECR isn't a supported registry?

cybojenix: https://github.com/concourse/registry-image-resource/pull/27

feature is being worked on.
Concourse recommends `registry-image` over `docker-image`

Makefile
cybojenix: leaving them in to make the switch easier

tommotaylor:
Looks good to me. Will wait for @emileswarts to approve

ci/pipelines/deploy.yml
dan-hill2802: Hoping to get it merged soon. In the meantime you could use:
```
- name: registry-image-resource
  type: docker-image
  source:
    repository: dwpdigital/registry-image-resource
```

cybojenix: Thanks for that @dan-hill2802. 
For now, we'll stick with using `docker-image` until the functionality is mainlined.
There's no pressing need to use `registry-image`

PR #611, opened 2019-04-04
Lint support specs part one

Renny4Real:
**WHY:**

**IN THIS PR:**
Lint and refactor the following files in the `spec/support` folder:
    - `spec/support/notifications_service.rb`
    - `spec/support/confirmation_use_case.rb`
    - `spec/support/unlock_account_use_case_spy.rb`

Rename shared examples where relevant


spec/features/authentication/logging_in_after_signing_up_spec.rb
```diff
@@ -4,8 +4,8 @@
 describe 'Logging in after signing up', type: :feature do
   let(:correct_password) { 'f1uffy-bu44ies' }
 
-  include_examples 'confirmation use case spy'
-  include_examples 'notifications service'
+  include_examples 'with confirmation use case spy'
+  include_examples 'with notifications service'
```
tommotaylor: I made a comment in https://github.com/alphagov/govwifi-admin/pull/612 that also applies here. Namely:
1. Using `shared_context` rather than `shared_examples`.
2. Renaming these shared contexts to be more descriptive

PR #612, opened 2019-04-04
Lint support specs part two

Renny4Real:
**WHY:**
Part two of PR #611 

**IN THIS PR:**
Lint and refactor the following files in the `spec/requests` folder:
 - `spec/support/reset_password_use_case.rb`
 - `spec/support/invite_use_case.rb`

Rename shared example usage were relevant


spec/features/invite_existing_user_spec.rb
```diff
@@ -6,7 +6,7 @@
   let(:betty) { create(:user) }
   let(:confirmed_user) { create(:user) }
 
-  include_examples 'invite use case spy'
+  include_examples 'with invite use case spy'
```
tommotaylor: Firstly, I don't think this really makes sense. "include examples with invite use case spy"????

Secondly, here we are calling `include_examples` yet `'with invite use case spy'` is a `shared_context`.

I think we we should be using `include_context` rather than `include_examples`, and it seems to work.

Also when renaming it, perhaps something more descriptive like `include_context 'when sending an invite email'`

PR #622, opened 2019-04-05
Lint remaining specs gateway folder

Renny4Real:
**WHY:**

**IN THIS PR:**
Lint and refactor the following files in the `spec/gateways` folder:
    - `spec/gateways/authorised_email_domains_spec.rb`
    -  `spec/gateways/email_spec.rb`
    - `spec/gateways/sessions_spec.rb`
    - `spec/gateways/route53_spec.rb`


spec/gateways/email_gateway_spec.rb
```diff
@@ -1,14 +1,16 @@
...
   before do
     ENV['NOTIFY_API_KEY'] = 'dummy_key-00000000-0000-0000-0000-000000000000-00000000-0000-0000-0000-000000000000'
-
-    expect_any_instance_of(Notifications::Client).to \
-      receive(:send_email).with(notifications_payload)
+    allow(Notifications::Client).to receive(:new).and_return(notification)
   end
 
-  context 'confirmation emails' do
+  context 'with confirmation emails' do
```
tommotaylor: This isn't very clear to me, it says "EmailGateway with confirmation emails calls notifications" when you put it all together.

After reading the spec it seems something better could be "EmailGateway when sending confirmation emails calls the Notify client"

PR #632, opened 2019-04-09
Lint remaining feature specs

tommotaylor:
Lint:
```
spec/features/ips/permissions_spec.rb
spec/features/ips/view_ip_address_readiness_spec.rb
spec/features/ips/add_ip_to_existing_location_spec.rb
spec/features/ips/view_ip_addresses_spec.rb
spec/features/ips/remove_an_ip_spec.rb
```

### DIFF
--- a/app/views/ips/_table.html.erb
+++ b/app/views/ips/_table.html.erb
@@ -1,9 +1,9 @@
 <table class="govuk-table govuk-!-margin-bottom-8">
   <caption class="govuk-table__caption text-left govuk-!-font-size-24">
-    <span id="location-address"><%= location.full_address %></span>
+    <%= location.full_address %>
     <div class="govuk-body govuk-!-margin-bottom-2">
       RADIUS secret key:
-      <div class="inline-block" id="radius-secret-key">
+      <div class="inline-block">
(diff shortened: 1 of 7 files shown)


app/views/ips/_table.html.erb
```diff
@@ -23,7 +23,7 @@
   <tbody class="govuk-table__body" id="ips-table">
     <% location.ips.each do |ip| %>
       <tr class="govuk-table__row">
-        <td class="govuk-table__cell govuk-!-width-one-third"><%= ip.address %></td>
+        <td class="govuk-table__cell govuk-!-width-one-third" id="ip-address"><%= ip.address %></td>
```
emileswarts: Is this going to print #ip-address to the page multiple times? A class instead perhaps?

tommotaylor: Its just giving the `td` an `id` attribute, so nothing changes in the view. I went for `id` selector over `class` as I consider classes for css and ids for javascript and testing.

emileswarts: Think it's against the rules to have more than one id with the same name on a page no?

PR #638, opened 2019-04-10
Allow a super admin to check the register before adding an organisation

georgeschena:
Screenshot 2019-04-10 at 13 31 09

### DIFF
--- a/app/controllers/admin/custom_organisations_controller.rb
+++ b/app/controllers/admin/custom_organisations_controller.rb
@@ -56,11 +56,7 @@ private
     sortable_directions.include?(params[:direction]) ? params[:direction] : sortable_directions.first
   end
 
-  def fetch_organisations_from_register
-    @register_organisations = Organisation.fetch_organisations_from_register
-  end
-
   def set_organisations_from_register
-    @fetched_organisations_from_register = fetch_organisations_from_register
+    @fetched_organisations_from_register = Organisation.fetch_organisations_from_register
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -12,6 +14,7 @@ def create
 
     if @custom_organisation.save
       flash.now[:notice] = "Custom organisation has been successfully added"
+      set_organisations_from_register
```
emileswarts: Isn't this run in the before action already?

georgeschena: Hahah let me explain

app/controllers/admin/custom_organisations_controller.rb
```diff
@@ -1,6 +1,8 @@
 class Admin::CustomOrganisationsController < AdminController
   helper_method :sort_column, :sort_direction
 
+  before_action :set_organisations_from_register, only: %i[create index]
```
emileswarts: Where is this instance variable being used for the create action?

app/controllers/admin/custom_organisations_controller.rb
georgeschena: Easier face to face

georgeschena: As its bit odd

app/views/admin/custom_organisations/index.html.erb
```diff
@@ -37,3 +39,18 @@
...
+  accessibleAutocomplete.enhanceSelectElement({
+    selectElement: document.querySelector('#organisations_register'),
+    required: true,
+    displayMenu: 'overlay',
+    defaultValue: '',
+    showNoOptionsFound: true,
+    showAllValues: true,
+    confirmOnBlur: true
+  })
+</script>
```
emileswarts: Could we pull this out into a partial so it's not duplicated?

PR #639, opened 2019-04-10
Refactor dataLayer script

Renny4Real:
**WHY:**
Based on feedback from Merkle, refactored the dataLayer script.
Also, the variable `hasIPs` is always false for super admins and this might lead to a false negative.
Another variable i.e `superAdmin` was added to the data layer.

**IN THIS PR:**
- Changed `dataLayer` to `window.dataLayer`

- Added superAdmin variable

_Super Admin view (note: `hasIPs` has a false value for a super administrator):_

Screenshot 2019-04-10 at 14 04 01

Should we be considering using the Data Layer with HTML Event Handlers, we currently don't?
We assign methods as the value to our dataLayer variables.

**Any suggestions/feedback on this would be appreciated.**

### DIFF
--- a/app/views/layouts/_google_analytics.html.erb
+++ b/app/views/layouts/_google_analytics.html.erb
@@ -5,7 +5,6 @@
     window.dataLayer = [{
       'signedIn': '<%= user_signed_in?.to_json %>',
       'hasIPs': '<%= (!!current_user && current_organisation.ips.any?).to_json %>'
-      'superAdmin': '<%= (user_signed_in? && super_admin?).to_json %>'
     }];
   </script>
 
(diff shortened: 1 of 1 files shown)


app/views/layouts/_google_analytics.html.erb
```diff
@@ -2,9 +2,10 @@
 <script async src="https://www.googletagmanager.com/gtag/js?id=<%= SITE_CONFIG['google_analytics_script_id'] %>"></script>
 
   <script>
-    dataLayer = [{
-      'signedIn': '<%= user_signed_in?.to_json %>',
+    window.dataLayer = [{
+      'signedIn': '<%= user_signed_in?.to_json %>',
       'hasIPs': '<%= (!!current_user && current_organisation.ips.any?).to_json %>'
+      'superAdmin': '<%= (user_signed_in? && super_admin?).to_json %>'
```
tommotaylor: I'm wondering how necessary this is as Super Admins will never visit the setup instructions page, they'll never add a location and they'll never add an IP.

They do visit the `/logs` and all other `/logs/something` endpoints, but everything else has `/admin/` preceeding it. So I would say we don't need it.

Renny4Real:
UPDATE: Removed superAdmin variable.

PR #640, opened 2019-04-10
Improve messaging when adding a location and IP 

georgeschena:
Screenshot 2019-04-10 at 15 09 17

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -6,12 +6,13 @@ class LocationsController < ApplicationController
 
   def create
     @location = Location.new(location_params_without_blank_ips)
+    number_of_ips_added = @location.ips.length
 
     if @location.save
       Facades::Ips::Publish.new.execute
       redirect_to(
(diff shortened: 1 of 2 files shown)


app/controllers/locations_controller.rb
```diff
@@ -6,12 +6,14 @@ def new
 
   def create
     @location = Location.new(location_params_without_blank_ips)
+    number_of_ips_added = @location.ips.length
+    pluralize_adress = 'IP address'.pluralize(number_of_ips_added)
```
emileswarts: I wouldn't be opposed to just inlining these.
No need for variables.

As it stands we may declare these variables and never use them if the location failed to save.

spec/features/locations/add_new_location_spec.rb
```diff
@@ -38,6 +38,20 @@
       end
     end
 
+    context 'when adding one or more ips' do
```
emileswarts: when adding multiple ips ?

app/controllers/locations_controller.rb
georgeschena: Fair point sir

PR #643, opened 2019-04-11
Add custom IPv6 error message

Renny4Real:
**WHY:**
An IPv4 address is what we take from the user. 
We don't accept other types of IP addresses and we let the user know through error messaging when an invalid IP address was inputted.

However, the error message displayed to the user is vague.
When a user enters a valid IPv6 address they see:
 `'1200:0000:AB00:1234:0000:2552:7777:1313' is not valid`
but are not informed why it is not valid.

IN THIS PR:
- Add hash key-value to the use case that checks the validity of submitted IP addresses. This would describe whether the IP address is IPv6 or not.
- Add custom error message for when the value of the IPv6 hash key is true.
- Refactor modified files.

**BEFORE:**

Screenshot 2019-04-11 at 12 42 46

**AFTER:**
Screenshot 2019-04-11 at 12 41 59

**Any feedback/suggestions would be appreciated. Especially on the refactoring.**

### DIFF
--- a/app/models/ip.rb
+++ b/app/models/ip.rb
@@ -12,12 +12,7 @@ private
 
   def address_must_be_valid_ip
     checker = UseCases::Administrator::CheckIfValidIp.new
-    valid_ipv4 = checker.execute(self.address)[:success]
-    valid_ipv6 = checker.execute(self.address)[:ipv6?]
-    ipv6_message = "'#{self.address}' is an IPv6 address. Only IPv4 addresses can be added."
-    return errors.add(:address, ipv6_message) if valid_ipv6
-
-    message = "'#{self.address}' is not valid"
-    errors.add(:address, message) unless valid_ipv4
+    results = checker.execute(self.address)
+    errors.add(:address, results[:error_message]) unless results[:success]
   end
 end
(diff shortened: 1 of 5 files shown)


lib/use_cases/administrator/check_if_valid_ip.rb
```diff
@@ -3,14 +3,14 @@ module Administrator
     class CheckIfValidIp
       def execute(address)
         @address = address
-        { success: valid_address? }
+        { success: valid_ipv4_address?, ipv6?: valid_ipv6_address? }
```
tommotaylor: Could we have an `error_message:` in this object so that its set here rather than in the `ip.rb` model?

It would make it more like the `ActiveRecord` pattern where errors are added to the object.

It would look something like:
```
{ success: valid_ipv4_address?, error_message: whatever_method_generates_the_error_message }
``` 

then the model can just attach the error to the `Ip` object, eg: 
```
errors.add(:address, checker[:error_message])
``` 
which can then pass it up to the view.

app/models/ip.rb
```diff
@@ -12,8 +12,12 @@ def available?
 
   def address_must_be_valid_ip
     checker = UseCases::Administrator::CheckIfValidIp.new
-    valid_ip = checker.execute(self.address)[:success]
+    valid_ipv4 = checker.execute(self.address)[:success]
+    valid_ipv6 = checker.execute(self.address)[:ipv6?]
```
tommotaylor: Calling `checker.execute` twice is a bit of a code smell. Try refactoring so that it only needs to be called once.

tommotaylor:
Nice work @Renny4Real

PR #648, opened 2019-04-12
Whitelisting flow

tommotaylor:
Still needs some design love but works and is tested at this stage. I suggest we merge and then improve the rest in separate PRs.

Test by visiting `/admin/whitelist/new` and going through the flow:
Screenshot 2019-04-16 at 12 50 57

### DIFF
--- a/app/controllers/admin/whitelists_controller.rb
+++ b/app/controllers/admin/whitelists_controller.rb
@@ -9,7 +9,8 @@ class Admin::WhitelistsController < AdminController
       email_domain: whitelist_params[:email_domain]
     )
     if @whitelist.save
-      redirect_to new_admin_whitelist_path, notice: "Saved"
+      redirect_to new_admin_whitelist_path,
+        notice: "Organisation has been whitelisted"
     else
       redirect_to new_admin_whitelist_path(step: "fourth")
     end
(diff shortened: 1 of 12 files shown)


app/controllers/admin/whitelists_controller.rb
```diff
@@ -1,3 +1,52 @@
 class Admin::WhitelistsController < AdminController
-  def new; end
+  def new
+    @whitelist ||= Whitelist.new
+
+    if organisation_name_submitted? && organisation_name_needs_validation?
+      @organisation_name = CustomOrganisationName.new(name: whitelist_params.dig(:organisation_name))
+      if @organisation_name.invalid?
+        params[:whitelist][:step] = "fourth"
```
Renny4Real: You could refactor this to be a guard statement:
```
      return params[:whitelist][:step] = "fourth" if @organisation_name.invalid?

      params[:whitelist][:organisation_name_valid] = true
    end
```

Renny4Real: or a ternary? e.g.
`@organisation_name.invalid? ? params[:whitelist][:step] = "fourth" : params[:whitelist][:organisation_name_valid]`

PR #649, opened 2019-04-12
Get spec coverage back up to 100%

emileswarts:
Screenshot 2019-04-12 at 13 20 01

### DIFF
--- a/app/controllers/admin/whitelists/organisation_names_controller.rb
+++ b/app/controllers/admin/whitelists/organisation_names_controller.rb
@@ -23,7 +23,7 @@ class Admin::Whitelists::OrganisationNamesController < AdminController
     custom_org = CustomOrganisationName.find(params.fetch(:id))
 
     custom_org.destroy
-    redirect_to admin_custom_organisations_path, notice: "Successfully removed #{custom_org.name}"
+    redirect_to admin_whitelist_organisation_names_path, notice: "Successfully removed #{custom_org.name}"
   end
 
 private
--- a/app/controllers/admin/whitelists/organisation_names_controller.rb
+++ b/app/controllers/admin/whitelists/organisation_names_controller.rb
@@ -22,12 +22,8 @@ class Admin::Whitelists::OrganisationNamesController < AdminController
   def destroy
     custom_org = CustomOrganisationName.find(params.fetch(:id))
 
-    if custom_org.destroy
-      notice = "Successfully removed #{custom_org.name}"
-    else
-      redirect_to admin_whitelist_organisation_names_path, notice: custom_org.errors.full_messages
-    end
-    redirect_to admin_whitelist_organisation_names_path, notice: notice
+    custom_org.destroy
+    redirect_to admin_custom_organisations_path, notice: "Successfully removed #{custom_org.name}"
   end
 
 private
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,7 +4,7 @@ module ApplicationHelper
   end
 
   def active_tab(identifier)
-    if request.path.include?(identifier) && identifier != root_path
+    if request.path.include?(identifier)
       "active"
     else
       ""
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,9 +4,7 @@ module ApplicationHelper
   end
 
   def active_tab(identifier)
-    if (request.path == root_path) && (request.path == identifier)
-      "active"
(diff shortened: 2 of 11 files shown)


app/controllers/admin/whitelists/organisation_names_controller.rb
```diff
@@ -22,12 +22,8 @@ def create
...
     custom_org = CustomOrganisationName.find(params.fetch(:id))
 
-    if custom_org.destroy
-      notice = "Successfully removed #{custom_org.name}"
-    else
-      redirect_to admin_whitelist_organisation_names_path, notice: custom_org.errors.full_messages
-    end
-    redirect_to admin_whitelist_organisation_names_path, notice: notice
+    custom_org.destroy
+    redirect_to admin_custom_organisations_path, notice: "Successfully removed #{custom_org.name}"
```
emileswarts: This model has no relations.
I cannot see a scenario where deleting one would ever raise an error.

Showed up as untested code in SimpleCov.

app/jobs/application_job.rb
```diff
@@ -1,2 +0,0 @@
-class ApplicationJob < ActiveJob::Base
-end
```
emileswarts: We don't use this, and probably never will.

emileswarts:
Screenshot 2019-04-12 at 13 20 01

spec/features/super_admin/govwifi_map_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    create(:location, postcode: 'SE10HS')
+    user = create(:user, :super_admin)
+    sign_in_user user
+  end
+
+  it 'shows a list' do
+    visit admin_govwifi_map_index_path
+
+    expect(page).to have_content('GovWifi Map of Locations')
+    assert_requested :post, 'http://api.postcodes.io/postcodes', body: 'postcodes[]=SE10HS'
```
cybojenix: I would expect this to post to the `https` site, not the unencrypted site

spec/spec_helper.rb
```diff
@@ -1,6 +1,10 @@
 require 'simplecov'
 
-SimpleCov.start
+SimpleCov.start 'rails'
+SimpleCov.add_filter [
```
cybojenix: Given that both of these have behaviour, they shouldn't be filtered out

tommotaylor:
@emileswarts We need to add a check to `SimpleCov` that ensures that not only are we hitting a minimum testing threshold, but that each file is also hitting this coverage threshold independently. We need it to block PR merge if it doesn't.

spec/spec_helper.rb
```diff
@@ -1,6 +1,7 @@
 require 'simplecov'
 
-SimpleCov.start
+SimpleCov.start 'rails'
+SimpleCov.minimum_coverage 100
```
cybojenix: also need by file:

```suggestion
SimpleCov.minimum_coverage 100
SimpleCov.minimum_coverage_by_file 100
```

emileswarts: Is per file not covered by this?
"minimum_coverage 100"

If a file isn't 100% covered, then this won't be either.

tommotaylor: True

spec/features/edit_organisation_details_spec.rb
```diff
@@ -32,4 +32,17 @@
...
+
+  context 'when inputting invalid details' do
+    before do
+      sign_in_user user
+      visit edit_organisation_path(organisation)
+      fill_in 'Service email', with: ''
+      click_on 'Save'
+    end
+
+    it 'will cause routing error' do
```
tommotaylor: Definitely being pedantic here but `"it will cause routing error"`, isn't whats being asserted here. The assertion is that the page shows the error message, so a better spec description would be `it "displays an error message to the user"`

emileswarts: Yeah makes sense, updated

app/helpers/application_helper.rb
```diff
@@ -4,9 +4,7 @@ def field_error(resource, key)
   end
 
   def active_tab(identifier)
-    if (request.path == root_path) && (request.path == identifier)
-      "active"
-    elsif request.path.include?(identifier) && identifier != root_path
+    if request.path.include?(identifier) && identifier != root_path
```
cybojenix: I think the latter condition also needs removing here, as this seems to be a defence against the identifier being `/` and marking all tabs as active.

If we're assuming the identifier is never `/`, then we can remove this too
```suggestion
    if request.path.include?(identifier)
```

PR #645, opened 2019-04-11
Dynamic search filter for locations/Ips

georgeschena:
Adds a JS powered search bar to IPs list

PR #654, opened 2019-04-15
[WIP]: Try out javascript testing in capybara

Renny4Real:
**WHY:**
We currently do not have a test to describe the filter added in PR #645.

This PR is an attempt to add a test for this feature so that additional changes to the dynamic filter feature can be made with more confidence.

**IN THIS PR:**
- Add `selenium-webdriver` for javascript testing
- Add failing test to describe behaviour.

Currently, having difficulty with testing the javascript as adding `js: true` to the context description causes web mock to error:

Screenshot 2019-04-15 at 15 50 31

The URLs in the stub request suggestions aren't the same and without the `js: true` tag, the test runs as normal.

NOTE: The test currently fails because the javascript code is not executed within the test.

**Would appreciate any feedback on the current test setup and/or how to test javascript with the selenium-webdriver and/or any suggestions on a better way to test javascript in our code.**

### DIFF
--- a/spec/features/ips/filter_ip_search_by_location_spec.rb
+++ b/spec/features/ips/filter_ip_search_by_location_spec.rb
@@ -6,11 +6,20 @@ describe 'filtering IP addresses by dynamic search' do
   let!(:location_three) { create(:location, organisation: organisation, address: 'Citrus Lane') }
 
   before do
+    # allowed_sites = [
+    #   "https://chromedriver.storage.googleapis.com",
+    #   "https://github.com/mozilla/geckodriver/releases",
+    #   "https://selenium-release.storage.googleapis.com",
+    #   "https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver"
+    # ]
+
+    WebMock.disable_net_connect!(allow_localhost: true, allow: ["app:55555", "selenium:4444", %r{github.com/mozilla/geckodriver/releases}])
+
     sign_in_user user
     visit ips_path
   end
 
-  context 'when user types in the search field' do
+  context 'when user types in the search field', js: true do
     it 'displays the location that closely matches the text field input' do
       fill_in 'filterInput', with: 'B'
       # page.execute_script(filterLocations())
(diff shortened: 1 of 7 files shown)


spec/features/ips/filter_ip_search_by_location_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  before do
+    sign_in_user user
+    visit ips_path
+  end
+
+  context 'when user types in the search field' do
+    it 'displays the location that closely matches the text field input' do
+      fill_in 'filterInput', with: 'B'
+      # page.execute_script(filterLocations())
```
Renny4Real: Pseudocode for executing the javascript function.

spec/features/ips/filter_ip_search_by_location_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      within('div#wrapper')do
+        expect(page).to have_selector("table", count: 1)
```
Renny4Real: The desired behaviour is that:
Since `Banana street` has B in it and no other location does, there should only be one table, not three. 
It fails because all three are still present.

spec/features/ips/filter_ip_search_by_location_spec.rb
emileswarts: Could you not type in the search and click buttons through the UI to trigger this?

spec/features/ips/filter_ip_search_by_location_spec.rb
emileswarts: Assertion could be that the page doesn't have the content that's been filtered?

spec/features/ips/filter_ip_search_by_location_spec.rb
Renny4Real: The filter doesn't have a button that triggers the search. 
You type in the field and it automatically locates a match.

spec/features/ips/filter_ip_search_by_location_spec.rb
Renny4Real: I could add that an additional test e.g when there are multiple matches.

I chose this assertion because I wanted to replicate the view of the filtered page as much as possible i.e a user mimicking the same actions in the current test setup should also only see one table (with Banana Street in it).

Renny4Real:
The next commits are all work in progress attempts at finding a solution.
Pushing them up for visibility.

Renny4Real:
Closed in favour of [WIP] PR #660.

PR #656, opened 2019-04-16
 Print no results message when filtering in the UI 

emileswarts:
This is a better user experience than not displaying any error messages.

Screenshot 2019-04-16 at 15 29 58

### DIFF
--- a/app/views/shared/_filter_search.html.erb
+++ b/app/views/shared/_filter_search.html.erb
@@ -4,22 +4,32 @@
 
   function filterLocations(){
     let filterValue = document.getElementById('filterInput').value.toUpperCase();
-    let table = document.getElementById('wrapper')
-    let location_name = table.querySelectorAll('table.govuk-table');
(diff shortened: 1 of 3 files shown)


app/views/shared/_filter_search.html.erb
```diff
@@ -4,22 +4,32 @@
...
-        location_name[i].style.display = 'none';
-        ips[i].style.display = 'none';
       }
     }
+
+    displayResultsMessage(location_name);
+  }
+
+function displayResultsMessage(location_name) {
+  var noResultsMessage = document.getElementById('noLocationResults');
```
Renny4Real: Should this be changed to `let` instead of `var`?
Feel like I've read that it's better to use `let` or `const` rather than `var`

emileswarts: These are very new.  Not sure all browsers will support this (given these are mainly government orgs)

PR #658, opened 2019-04-17
Display when an IP is inactive

emileswarts:
Administrators find it challenging to identify which IPs are no longer
serving traffic.  Label an IP as inactive so they can safely remove
them.

Screenshot 2019-04-17 at 14 45 12

### DIFF
--- a/app/models/ip.rb
+++ b/app/models/ip.rb
@@ -4,6 +4,14 @@ class Ip < ApplicationRecord
   validates :address, presence: true, uniqueness: true
   validate :address_must_be_valid_ip
 
+  def inactive?
+    Session
+      .where(siteIP: self.address)
+      .where("start > #{Date.today - 10.days}")
+      .limit(1)
+      .empty?
+  end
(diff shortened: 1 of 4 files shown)


app/models/ip.rb
```diff
@@ -4,6 +4,14 @@ class Ip < ApplicationRecord
   validates :address, presence: true, uniqueness: true
   validate :address_must_be_valid_ip
 
+  def inactive?
+    Session
+      .where(siteIP: self.address)
+      .where("start > #{Date.today - 10.days}")
+      .limit(1)
+      .empty?
+  end
```
emileswarts: This is a very naive implementation that has the potential to put a lot of load on the sessions read replica.
Before optimising this, I want to see how much load in production exactly.  Maybe it's fine as nothing else uses this database.

emileswarts: This is not performance optimised. Want to see how it behaves in production before making looking into this.

PR #659, opened 2019-04-17
Use FactoryBot for creating sessions in specs

emileswarts:


spec/factories/sessions.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :session do
+    sequence :siteIP do |n|
```
Renny4Real: What is the sequence method doing?

spec/factories/sessions.rb
```diff
@@ -0,0 +1,9 @@
...
+      fourth = n % 255
+      third = n / 255
+      "141.0.#{third}.#{fourth}"
```
Renny4Real: Are we returning an IP address?

emileswarts: A sequence is every time you call it, you will get a different IP address.

PR #660, opened 2019-04-17
[WIP]: Allow javascript testing v2

Renny4Real:
**WHY:**
VERSION TWO OF PR #654 
We currently do not have a test to describe the filter added in PR #645.

This PR is an attempt to add a test for this feature so that additional changes to the dynamic filter feature can be made with more confidence.

**IN THIS PR:**
- Add `selenium-webdriver` gem.
- Add `webdriver` gem.
- Add failing test to describe behaviour.
- Set up using chromium package and chrome driver in the `Dockerfile`.
- Add selenium chrome image to `docker-compose.yml`

Currently experiencing this error in the test ([on this line](https://github.com/alphagov/govwifi-admin/compare/add-javascript-testing-v2?expand=1#diff-cb47a38d8225ef7ac9caadc38233ee6fR45)):

Screenshot 2019-04-17 at 15 33 37

NOTE: The test currently fails because the page is trying to redirect to localhost and can't.

**Would appreciate any feedback on the current test setup and/or how to test javascript with the selenium-webdriver and/or any suggestions on a better way to test javascript in our code.**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ENV RR_DB_NAME rr_govwifi
 WORKDIR /usr/src/app
 
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
-  apt-get update && apt-get install -y apt-transport-https nodejs libuv1 && \
+  apt-get update && apt-get install -y apt-transport-https nodejs libuv1 chromium chromedriver && \
   curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.9.4 && \
   rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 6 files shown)


Renny4Real:
**UPDATE:**

Currently, the connection seems to work but the page loads with nothing in it:

`<html xmlns=\"http://www.w3.org/1999/xhtml\"><head></head><body></body></html>`

Which leads to these errors in these failing tests:

Screenshot 2019-04-18 at 11 50 25

This was an attempt at allowing javascript testing and the results are better than the attempt in PR #654 

However, will leave this as [WIP] for now.

Renny4Real:
Closing this for now:
- Current consensus is this is not necessary and hard to implement. 
- We don't want to to encourage adding Javascript to the codebase as well. 
- We also want to keep Javascript in the code to a minimum.

PR #662, opened 2019-04-18
Allow filtering by organisation for super admins

emileswarts:
This existed on locations, modify this to be general enough to be used
in more than one place.

Screenshot 2019-04-18 at 13 21 35

### DIFF
--- a/app/views/ips/index.html.erb
+++ b/app/views/ips/index.html.erb
@@ -25,7 +25,7 @@
 
   <% else %>
     <p class="govuk-body" id="no-results" style="display: none">
-      No results found
+      <strong>No results found</strong>
     </p>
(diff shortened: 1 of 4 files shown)


app/views/ips/index.html.erb
```diff
@@ -26,11 +24,12 @@
     <p class="govuk-body">If your authenticators are allocated IPs dynamically, you can use Firewall NAT rules, so that your requests come from consistent IPs.</p>
 
   <% else %>
-    <p class="govuk-body" id="noLocationResults" style="display: none">
+    <p class="govuk-body" id="no-results" style="display: none">
       No results found
```
Renny4Real: Wondering if we could style this to be bold text or larger font size?

PR #663, opened 2019-04-18
Allow users to clear search terms

emileswarts:
This prints an x icon in the input and when users click it, it resets
the search.

Screenshot 2019-04-18 at 16 07 56

### DIFF
--- a/app/views/admin/organisations/index.html.erb
+++ b/app/views/admin/organisations/index.html.erb
@@ -12,7 +12,7 @@
 
 <%= link_to 'GovWifi map', admin_govwifi_map_index_path, class: "govuk-link" %>
 
-<p><input type='search' id='filter-input' placeholder='Search by organisation name' class='govuk-input'></p>
+<p><input type="search" id="filter-input" placeholder="Search by organisation name" class="govuk-input"></p>
 
 <table class="govuk-table">
   <thead class="govuk-table__head">
--- a/app/views/admin/organisations/index.html.erb
+++ b/app/views/admin/organisations/index.html.erb
@@ -12,7 +12,7 @@
 
 <%= link_to 'GovWifi map', admin_govwifi_map_index_path, class: "govuk-link" %>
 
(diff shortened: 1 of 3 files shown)


app/views/admin/organisations/index.html.erb
```diff
@@ -12,7 +12,7 @@
 
 <%= link_to 'GovWifi map', admin_govwifi_map_index_path, class: "govuk-link" %>
 
-<p><input type='text' id='filter-input' placeholder='Search by organisation name' class='govuk-input'></p>
```
chrisblackburn: Should these be double quotes?

app/views/admin/organisations/index.html.erb
```diff
@@ -12,7 +12,7 @@
...
+<p><input type="search" id="filter-input" placeholder="Search by organisation name" class="govuk-input"></p>
```
cybojenix: `search` is non-standard. Suggest not using it

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/search_event

app/views/shared/_filter_search.html.erb
```diff
@@ -1,6 +1,7 @@
 <script>
   var filterInput = document.getElementById('filter-input');
   filterInput.addEventListener('keyup', filterResults);
+  filterInput.addEventListener('search', filterResults);
```
cybojenix: for browsers that support the `search` event, this will end up filtering the results twice.
while this won't technically cause an issue, it will result in the code running twice, resulting in noticeable slowdowns with larger lists

app/views/admin/organisations/index.html.erb
emileswarts: Old browsers will fall back to using a text field here.
Won't prevent them from using this functionality minus some nice features of the UI?

Notify are using this input as well.

app/views/shared/_filter_search.html.erb
emileswarts: Search only fires when you hit the x button.
Can confirm that they don't both fire when typing.

PR #664, opened 2019-04-18
Give no indication of the existence of a user's account.

Renny4Real:
**WHY:**
We should not display messages that could give an attacker any information about the existence or status of a user account. 
e.g When entering a username/password combo, we would potentially be telling an attacker that the username entered was correct if the error message only displays `password is incorrect` or vice-versa.
Good read on this is the [OWASP Cheatsheet.](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Authentication_Cheat_Sheet.md)

We should be responding with generic messages that give no indication as to whether:
- An account exists or doesn't.
- Any sign-in credentials are correct or incorrect.

Devise has a [paranoid feature](https://github.com/plataformatec/devise/wiki/How-To:-Using-paranoid-mode,-avoid-user-enumeration-on-registerable). 
This sends/displays instructions that are the same regardless of whether, the user credentials provided were right or wrong.

**IN THIS PR:**

- Set [`config.paranoid`](https://github.com/alphagov/govwifi-admin/compare/reword-exposing-messages?expand=1#diff-a8a4112d0760c0fd11f691c6a5fb22adR84) in [`devise.rb`](config/initializers/devise.rb) to true.

- Edit assertions in example tests that describe [locking a user account](spec/features/authentication/lock_user_account_spec.rb), [resending a confirmation email](spec/features/registration/resend_confirmation_instructions_spec.rb) & [resetting a password](spec/features/authentication/reset_password_spec.rb) to expect the generic response.

- Refactor the examples in the above spec files to describe/explain the user flow better since specific messages are no longer being used.

**Password reset request**
After providing an email address in the text field and clicking on `Send me reset password instructions`, it displays:

Screenshot 2019-04-18 at 13 51 33

**Locked Account**
After 10 consecutive failed sign in attempts, it will send the unlock email but will also continue to display:

Screenshot 2019-04-18 at 15 33 56

and will not change even if the username/password combo entered on the next attempt is correct, i.e account is now locked.

**Resend Confirmation Email**
After providing an email address in the text field and clicking on `Resend confirmation instructions`, it displays:

Screenshot 2019-04-18 at 13 51 15

**Any feedback/suggestions on this approach would be appreciated.**

### DIFF
--- a/spec/features/registration/resend_confirmation_instructions_spec.rb
+++ b/spec/features/registration/resend_confirmation_instructions_spec.rb
@@ -8,7 +8,7 @@ describe 'Resending confirmation instructions', type: :feature do
   include_context 'when sending a confirmation email'
   include_context 'when using the notifications service'
 
-  context 'when entering an email address that has signed up' do
+  context 'when entering an email address that has signed up but not confirmed' do
     before do
       sign_up_for_account(email: unconfirmed_email)
       visit new_user_confirmation_path
@@ -50,7 +50,7 @@ describe 'Resending confirmation instructions', type: :feature do
       fill_in 'user_email', with: new_user_email
     end
 
-    it 'displays generic response message to the user' do
+    it 'displays a generic response to the user' do
       click_on 'Resend confirmation instructions'
       expect(page).to have_content 'If your email address exists in our database, you will receive an email with instructions for how to confirm your email address in a few minutes.'
     end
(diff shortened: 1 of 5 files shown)


emileswarts:
Nice.  Just had a quick scan around other GDS apps to see how they do this and they seem to always say something like "click the link in your inbox".  

Whether it was a real email address or not.

Wonder if this is just as effective but less openly defensive?

Renny4Real:
NOTE: 
Change paranoid instructions in [`devise.en.yml`](config/locales/devise.en.yml) to be more polite. e.g Click the link sent to this email.

**Any suggestions/feedback on wording this would be appreciated.**

spec/features/registration/resend_confirmation_instructions_spec.rb
```diff
@@ -3,77 +3,83 @@
...
 
     it 'resends the confirmation link' do
       expect {
         click_on 'Resend confirmation instructions'
       }.to change(ConfirmationUseCaseSpy, :confirmations_count).by(1)
     end
+
+    it 'displays generic response message to the user' do
+      click_on 'Resend confirmation instructions'
+      expect(page).to have_content 'If your email address exists in our database, you will receive an email with instructions for how to confirm your email address in a few minutes.'
```
emileswarts: Maybe this?

If you have an account with us, you will receive an email with a confirmation link.

Renny4Real:
**UPDATE:**

**Resend Confirmation Email:**

After providing an email address in the text field and clicking on `Resend confirmation instructions`, it displays:

Screenshot 2019-04-23 at 10 21 46

**Password reset request:**

After providing an email address in the text field and clicking on `Send me reset password instructions`, it displays:

Screenshot 2019-04-23 at 10 21 56

Renny4Real:
Added this commit: fb14a3d to get the pipeline to rebuild and test the PR again. 
(The PR checks were still yellow  for a while even though the checks were green in the pipeline).

PR #671, opened 2019-04-23
Spike rails 6

georgeschena:
5 tests failed as validate_presence_of relies on the constant ActiveModel::SecurePassword::InstanceMethodsOnActivation which has been removed in Rails 6. However they have fixed this issue but have not released the new gem version at the time of this PR. It should be released when rails 6 is live so won't need the extra line in the spec helper. Overall pretty frictionless from what I can see


cybojenix:
closing this as we're not doing this until rails 6 is released

PR #672, opened 2019-04-23
Validate service email format when creating or updating it

Renny4Real:
**WHY:**
When a user adds or edits their organisation's service email, it only validates the email's presence i.e. `email can't be blank` but it doesn't validate the email's format.
As a result, this allowed any form of text to be accepted and saved to the organisation, even if the text was not an email address.

**IN THIS PR:**
- Add validation on the service email to match Devise's email regex.
- Add tests on the model and on the feature to describe validating a service email's format.

**BEFORE:**

Adding any text and clicking on save, redirected to the Team members page and displayed:
Screenshot 2019-04-23 at 11 40 08

**AFTER:**

Adding any text and clicking on save, check's the format of the text and displays an error message, if the text does not match devise's email regexp:

Screenshot 2019-04-23 at 11 40 35

### DIFF
--- a/spec/models/organisation_spec.rb
+++ b/spec/models/organisation_spec.rb
@@ -68,4 +68,34 @@ describe Organisation do
       ])
     end
   end
+
+  context 'when organisation service email is left blank' do
+    let(:organisation) { build(:organisation, service_email: '') }
+
+    it 'is not valid' do
+      expect(organisation).not_to be_valid
+    end
+
+    it 'explains why it is invalid' do
+      organisation.valid?
+      expect(organisation.errors.full_messages).to eq([
+        "Service email can't be blank"
+      ])
+    end
+  end
+
+  context 'when organisation service email is not an email address' do
+    let(:organisation) { build(:organisation, service_email: 'IncorrectFormat') }
+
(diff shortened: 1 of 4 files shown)


spec/models/organisation_spec.rb
```diff
@@ -68,4 +68,34 @@
...
+    let(:organisation) { build(:organisation, service_email: '') }
+
+    it 'is not valid' do
+      expect(organisation).not_to be_valid
+    end
+
+    it 'explains why it is invalid' do
+      organisation.valid?
+      expect(organisation.errors.full_messages).to eq([
+        "Service email can't be blank"
```
emileswarts: I see we protect from nil but empty string is ok.
When does this happen as a user is using the site?

Renny4Real: @emileswarts Not sure if I've understood your comment/question correctly, but a user can also update their service email after confirming their account.

Renny4Real:
Suggestion was made to validate service email only on format and not on presence since an empty field still fails the validation on format.

UPDATE:
- Remove validation on presence for an Organisation's service email.

PR #673, opened 2019-04-23
Fix google maps inconsistent loading

georgeschena:


app/views/admin/govwifi_map/index.html.erb
```diff
@@ -1,10 +1,9 @@
 <%= link_to "Back to organisations", admin_organisations_path, class: "govuk-back-link" %>
  <h1 class="govuk-heading-l">GovWifi Map of Locations</h1>
-  <script async defer src="https://maps.googleapis.com/maps/api/js?key=<%= ENV['GOOGLE_MAPS_PUBLIC_API_KEY'] %>&callback=initMap"></script>
 
   <div class="map"></div>
    <script>
-     function initMap() {
+   window.initMap = function(){
```
emileswarts: Indenting is a bit off if you look at the script tag start.
No space after closing )

app/views/admin/govwifi_map/index.html.erb
```diff
@@ -25,3 +24,4 @@
        }
      }
    </script>
+<script src="https://maps.googleapis.com/maps/api/js?key=<%= ENV['GOOGLE_MAPS_PUBLIC_API_KEY'] %>&callback=initMap" async defer></script>
```
emileswarts: This script tag should be aligned with the script tag above it?

app/views/admin/govwifi_map/index.html.erb
cybojenix: Can you define as `window.initMap = function initMap()` please, which will help with debugging purposes.

with `window.initMap = function()`, it shows up in traces as an anonymous function

cybojenix: Also unsure if it needs attaching to `window`. The problem will have been from the fetched script being cached, which loads before we reach this `script` block defining `initMap`.
Moving the `script` block to above the async script should fix the issue

PR #677, opened 2019-04-26
Update copy on sign in page

tommotaylor:
Update copy on `/users/sign_in`

Screenshot 2019-04-26 at 11 41 11


app/views/users/sessions/new.html.erb
```diff
@@ -1,10 +1,15 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">Sign in</h2>
+    <h2 class="govuk-heading-l">Network administrator account</h2>
```
emileswarts: NICE
Just this change is going to have a big impact on confused end users

PR #682, opened 2019-04-30
convert admin to use same test/lint standard as the main pipelines

cybojenix:


ci/pipelines/pr.yml
```diff
@@ -46,48 +46,58 @@ jobs:
...
+        - get: docker-cache/mysql-image
+          resource: mysql-image
           params: {format: oci}
-        - get: ruby-image
+        - get: docker-cache/ruby-image
+          resource: ruby-image
           params: {format: oci}
-        - get: nginx-image
+        - get: docker-cache/nginx-image
+          resource: nginx-image
```
emileswarts: Do we need Nginx?  Pretty sure we just use Puma.

cybojenix: it's used by the `fake-register`

emileswarts: Ah yes ok

ci/tasks/scripts/with-docker.sh
```diff
@@ -2,19 +2,28 @@
...
+
+  index=0
+  for cached_image in docker-cache/*/image.tar; do
+    docker load -qi "${cached_image}" & pids[${index}]=$!
+    index="$(( "${index}" + 1 ))"
+  done
 
   for pid in ${pids[*]}; do
-    wait "$pid" || true  # we don't care about the return code of docker loading
+    wait "$pid"
```
emileswarts: Just out of curiosity, we now care about the return code of docker loading?

cybojenix: ah, explanation for this one.

This occurred when the cache wasn't populated, and relates to the old line 14.

```
[[ -f "govwifi-admin-prebuilt/image.tar" ]] && docker load -qi "govwifi-admin-prebuilt/image.tar" & pids[3]=$!
```

The `& pids[3]=$!` is run on the whole statement, not just the part after the condition check.
This means that for when the file doesn't exist, the process of that statement will return an exit code of 1 due to the `[[ ... ]]` returning falsely.

adding `wait || true` was a hack to work around the issue as I didn't fully understand the cause before.

With the move to globbing, the only issues we'll get are ones we want to raise

PR #691, opened 2019-05-01
Add uuid to organisations

emileswarts:
When creating an organisation, it will create its own uuid and persist
it in the database. All existing organisations will be manually
backfilled to contain new uuids.

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -8,6 +8,8 @@ class Organisation < ApplicationRecord
   validates :service_email, format: { with: Devise.email_regexp, message: "must be a valid email address" }
   validate :validate_in_register?, unless: Proc.new { |org| org.name.blank? }
 
+  after_initialize :generate_uuid
+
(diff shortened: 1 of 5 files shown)


app/models/organisation.rb
```diff
@@ -19,4 +21,10 @@ def self.fetch_organisations_from_register
       organisations_gateway: Gateways::GovukOrganisationsRegisterGateway.new
     ).execute.sort
   end
+
+  def generate_uuid
+    return if uuid.present?
+
+    self.uuid = SecureRandom.uuid
+  end
```
tommotaylor: Can achieve this without the need for ruby I think:
```
class AddUuidToBooks < ActiveRecord::Migration
  def change
    add_column :organisations, :uuid, :uuid, default: 'uuid_generate_v4()'
  end
end
```

emileswarts: 👍

PR #696, opened 2019-05-03
Add MOU & IP routes to Organisation/UUID scope

Renny4Real:
**WHY:**
As part of allowing users to register multiple organisations, we want to edit the URL structure to have the current organisation's uuid within it, this is an iterative slice towards that objective.

This slice involves moving the MOU & IP resource routes into `scope '/organisations/:organisation'`

The new URL would look like, e.g: `https://admin.wifi.service.gov.uk/organisations/aaaaaa-bbbbb-cccc-dddd-eeee1111/ips`

**IN THIS PR:**
- Move `resources :ips, only: %i[index new create destroy]` into the `scope '/organisations/:organisation'`

- Move `resources :mou, only: %i[index create]` into the `scope '/organisations/:organisation'`

- Fix failing specs
- Refactor route paths
- Refactor test

### DIFF
--- a/app/controllers/mou_controller.rb
+++ b/app/controllers/mou_controller.rb
@@ -6,9 +6,9 @@ class MouController < ApplicationController
 
   def create
     redirect_path = if replacing?
-                      replaced_mou_index_path(organisation_uuid: current_organisation.uuid)
+                      replaced_mou_index_path(organisation: current_organisation)
                     else
-                      created_mou_index_path(organisation_uuid: current_organisation.uuid)
+                      created_mou_index_path(organisation: current_organisation)
                     end
 
     if params[:signed_mou]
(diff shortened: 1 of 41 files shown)


app/controllers/mou_controller.rb
```diff
@@ -5,7 +5,12 @@ def index
   end
 
   def create
-    redirect_path = replacing? ? replaced_mou_index_path : created_mou_index_path
+    redirect_path = if replacing?
+                      replaced_mou_index_path(organisation: current_organisation)
+                    else
+                      created_mou_index_path(organisation: current_organisation)
+                    end
```
tommotaylor: @Renny4Real Looks like these changes are already in another PR: https://github.com/alphagov/govwifi-admin/pull/695/files

Did you branch off a feature branch rather than master?

emileswarts: Yeah, this branch contains everything we need so closed the feature branch.

PR #698, opened 2019-05-07
Find current organisation by session

emileswarts:
Users can belong to multiple organisations.
Ensure users cannot switch to other organisations by setting a session
manually.


app/controllers/application_controller.rb
```diff
@@ -7,7 +7,11 @@ class ApplicationController < ActionController::Base
   helper_method :current_organisation, :super_admin?
 
   def current_organisation
-    @current_organisation ||= current_user.organisation
+    if session[:organisation_id] && session[:organisation_id] == current_user.organisation.id
+      Organisation.find(session[:organisation_id])
```
emileswarts: This will be of more use when users have a many to many relationship with organisations

Gemfile.lock
```diff
@@ -377,7 +381,7 @@ DEPENDENCIES
   zendesk_api
 
 RUBY VERSION
-   ruby 2.6.2p47
+   ruby 2.6.3p62
```
tommotaylor: Is this going to blowup because of our `.ruby-version` file and whatever ruby version we have in docker?

spec/features/switch_organisation_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let(:other_organisation) { create(:organisation) }
+
+  let(:user) { create(:user, organisation: create(:organisation)) }
+
+  before do
+    sign_in_user user
+  end
+
+  context "when the user doesn't belong to the organisation" do
+    it 'dissallows allows switching to that organisation' do
```
tommotaylor: Doesn't read well / make sense, but I get it.

app/controllers/application_controller.rb
```diff
@@ -7,7 +7,11 @@ class ApplicationController < ActionController::Base
   helper_method :current_organisation, :super_admin?
 
   def current_organisation
-    @current_organisation ||= current_user.organisations.first
+    @organisation ||= if session[:organisation_id] && current_user.organisations.map(&:id).include?(session[:organisation_id])
```
emileswarts: This query above is inefficient and will be addressed in a future PR

Gemfile.lock
emileswarts: This is a dependabot upgrade, we've not updated the reference here yet.
So shouldn't blow up for anyone

spec/features/switch_organisation_spec.rb
emileswarts: Agreed, have updated it

app/controllers/application_controller.rb
georgeschena: Have we intentionally lost the `@current_organisation` cacheing?

emileswarts: Yes this memoization breaks the new implementation.
Old implementation was fine because it never changed.
Have a note to come back to this.

Renny4Real:
@georgeschena Yeah, it was breaking tests.

PR #701, opened 2019-05-07
Adding many to many relationship between Orgs and Users

emileswarts:


app/controllers/application_controller.rb
```diff
@@ -7,7 +7,7 @@ class ApplicationController < ActionController::Base
   helper_method :current_organisation, :super_admin?
 
   def current_organisation
-    @current_organisation ||= current_user.organisation
+    @current_organisation ||= current_user.organisations.first
```
emileswarts: This will change in the future, just preserving existing functionality for now

app/models/organisation.rb
```diff
@@ -1,6 +1,6 @@
 class Organisation < ApplicationRecord
   has_one_attached :signed_mou
-  has_many :users, inverse_of: :organisation, dependent: :destroy
+  has_and_belongs_to_many :users, inverse_of: :organisation
```
tommotaylor: This means we might end up with users with no orgs. Not an issue to solve now though I don't think

PR #709, opened 2019-05-10
Download CSV file of all service emails

georgeschena:
This PR allows a super admin to download all the organisations service emails in CSV format.
Screenshot 2019-05-13 at 12 30 59


app/controllers/admin/contact_all_orgs_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::ContactAllOrgsController < AdminController
```
tommotaylor: Not a very RESTful name.

app/models/organisation.rb
```diff
@@ -22,6 +22,16 @@ def self.fetch_organisations_from_register
...
   end
 
+  def to_csv
+    attributes = %w{service_email}
+    CSV.generate(headers: true) do |csv|
+      csv << attributes
+      all.each do |location|
+        csv << location.attributes.values_at(*attributes)
+      end
+    end
```
tommotaylor: Looks good but curious as to why you chose the model over the controller

config/application.rb
```diff
@@ -1,6 +1,7 @@
 require_relative 'boot'
 
 require "rails"
+require 'csv'
```
tommotaylor: Is this needed? `CSV` doesn't come with rails?

config/routes.rb
```diff
@@ -69,6 +69,7 @@
 
   namespace :admin do
     resources :govwifi_map, only: %i[index]
+    resources :contact_all_orgs, only: %i[index]
```
tommotaylor: Not a very resourceful route

spec/features/super_admin/download_all_service_emails_spec.rb
```diff
@@ -0,0 +1,17 @@
+describe 'Downloading all the service emails', type: :feature do
+  let(:user) { create(:user, :super_admin) }
```
tommotaylor: The latest changes to the user factory means you need to explicitly pass in the trait for organisation, eg:
```
let(:user) { create(:user, :super_admin, ;with_organisation) }
```

spec/requests/organisations/get_spec.rb
```diff
@@ -0,0 +1,17 @@
+describe "GET /locations", type: :request do
```
emileswarts: /locations old description

PR #715, opened 2019-05-13
Remove organisation id from user table

Renny4Real:
**WHY:**
We're removing `organisation.id ` from the user table as it exists in its own table as a many to many relationship.

**IN THIS PR:**
- Create & apply migration to remove the `organisation_id` column from the `users` table.
- Refactor invitations controller as `organisation_id` is no longer to available to pass as params within the `invite_params` method.

**Any suggestions/feedback on this approach would be appreciated.**


georgeschena:
Closed as this has now been completed

PR #721, opened 2019-05-14
Allow users to create additional organisations

georgeschena:
Allow users to create additional organisations:
Screenshot 2019-05-14 at 15 18 50
Screenshot 2019-05-14 at 15 20 29


app/controllers/organisations_controller.rb
```diff
@@ -29,7 +47,15 @@ def user_belongs_to_same_org
     current_organisation == @organisation
   end
 
+  def assign_user_to_organisation(organisation)
+    current_user.organisations << organisation
```
emileswarts: We only want to do this if the organisation has no users associated with it right?
Should we guard for that here?

emileswarts: NM

tommotaylor: Yeah it can only be a brand new organisation at this point

PR #727, opened 2019-05-15
Pending team list

emileswarts:
Display pending cross-organisation invited users in the team list alongside members.

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -22,12 +22,6 @@ class Organisation < ApplicationRecord
   end
 
   def team_list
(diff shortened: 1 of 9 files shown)


app/models/organisation.rb
```diff
@@ -19,4 +20,8 @@ def self.fetch_organisations_from_register
       organisations_gateway: Gateways::GovukOrganisationsRegisterGateway.new
     ).execute.sort
   end
+
+  def team_list
```
phantas: Would you mind walking me through this? This data model looks a bit windy to manage the user x organisation relation

PR #730, opened 2019-05-17
View and search locations

georgeschena:
This PR involves adding a searchable table of all the locations using GovWifi. This allows super admins to dynamically search through the table to find a desired location, the searching works either by location name or via postcode

Screenshot 2019-05-20 at 10 29 02

Searching via postcode: 
Screenshot 2019-05-20 at 10 30 27

### DIFF
--- a/app/controllers/admin/locations_controller.rb
+++ b/app/controllers/admin/locations_controller.rb
@@ -1,0 +1,4 @@
+class Admin::LocationsController < AdminController
+  def index
+  end
+end
--- a/app/controllers/admin/locations_controller.rb
+++ b/app/controllers/admin/locations_controller.rb
@@ -1,7 +1,6 @@
 class Admin::LocationsController < AdminController
   def index
-    @locations = Location.all
-    .order("#{sort_column} #{sort_direction}")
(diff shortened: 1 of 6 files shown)


app/views/admin/locations/index.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-half">
+    <h1 class="govuk-heading-l">GovWifi locations </h1>
```
emileswarts: extra space before closing </h1>

emileswarts: h1

app/controllers/admin/locations_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Admin::LocationsController < AdminController
+  def index
+    @locations = Location.all
```
emileswarts: Does calling .order without .all do the same thing?
I'm not sure, can you check?

georgeschena: Can confirm works the same. So removing `.all`

PR #731, opened 2019-05-17
Simplify user to organisation association

tommotaylor:
A simpler approach to users <=> organisations.

Replaces: https://github.com/alphagov/govwifi-admin/pull/727

- Uses a `Membership` model as the join between users and organisations. 
- Renames `CrossOrganisationInvitation` to `Membership`
- User has many organisations through memberships
- Drops the `has_and_belongs_to_many` association and table

To Do:
- [ ] needs a data migration to migrate all existing organisation data over to the memberships table
- [x] need to confirm if all invited members of an organisation are appearing in the team list
- [x] Need to see if we can move the invitation information that currently lives on users and put it on memberships. This would require getting devise to save this info to the new membership model rather than the user on first invite.

Migration table rename works locally and does not loose the existing data -'George;

### DIFF
--- a/db/migrate/20190517152315_remove_null_false_constraint_from_invitation_token_on_memberships.rb
+++ b/db/migrate/20190517152315_remove_null_false_constraint_from_invitation_token_on_memberships.rb
@@ -1,0 +1,7 @@
+class RemoveNullFalseConstraintFromInvitationTokenOnMemberships < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :memberships, :invitation_token, true
+    change_column_null :memberships, :invited_by_id, true
+    change_column_null :memberships, :confirmed_at, true
+  end
+end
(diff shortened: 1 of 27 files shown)


db/migrate/20190517152315_remove_null_false_constraint_from_invitation_token_on_memberships.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveNullFalseConstraintFromInvitationTokenOnMemberships < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :memberships, :invitation_token, true
+    change_column_null :memberships, :invited_by_id, true
+    change_column_null :memberships, :confirmed_at, true
```
tommotaylor: This is a bit ugly. Rails was trying to create the membership object first when creating it through the user, this was causing db validation errors on the `NOT NULL` constraint.

Had a go at fixing it in code like so but failed: https://thoughtbot.com/blog/accepts-nested-attributes-for-with-has-many-through

So in the end just removed the validations.

PR #748, opened 2019-05-22
Fix pending user issue 

georgeschena:
The PR fixes the pending user issue when a user that already belongs to an organisation is then invited to another organisation.

### DIFF
--- a/spec/features/pending_user_spec.rb
+++ b/spec/features/pending_user_spec.rb
@@ -19,7 +19,7 @@ describe 'Inviting an existing user which belongs to an organisation', type: :fe
       click_on 'Send invitation email'
     end
 
-    it 'will pending users which have been invited to an organisation' do
+    it 'will show pending users which have been invited to an organisation' do
       expect(page).to have_content('inviteme@gov.uk (invited')
     end
   end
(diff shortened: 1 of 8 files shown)


spec/features/team/view_team_members_spec.rb
```diff
@@ -52,7 +52,7 @@
       end
 
       it 'renders all team members within my organisation in alphabetical order' do
-        expect(page.body).to match(/#{user_2.name}.*#{user_1.email}.*#{user.name}.*#{user_3.name}/m)
+        expect(page.body).to match(/#{user_2.email}.*#{user_1.email}.*#{user.email}.*#{user_3.email}/m)
```
emileswarts: Can't see why this assertion has changed.

spec/features/pending_user_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    before do
+      sign_in_user user
+      visit team_members_path
+      click_on 'Invite team member'
+      fill_in 'Email', with: user_2.email
+      click_on 'Send invitation email'
+    end
+
+    it 'will show pending users which have been invited to an organisation' do
+      expect(page).to have_content('inviteme@gov.uk (invited')
```
emileswarts: partial matching on "(invited" ?

emileswarts: Still unsure about this (invited assertion.
Is this on purpose?

PR #761, opened 2019-05-31
Redirect users without memberships to orgs

steventux:
https://trello.com/c/0g3MYHja/28-allow-a-orphaned-user-to-login

Screenshot from 2019-06-06 09-15-41

When a user is no longer associated with any organisation (this is possible if an admin deletes a user's only org) we need to ensure that, once authenticated, they associate themselves with an organisation.
This PR makes sure the orphaned user is redirected to the support form with an additional explanation of why they are there.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,7 +3,7 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate_user!, except: :error
   before_action :configure_devise_permitted_parameters, if: :devise_controller?
-  before_action :ensure_user_belongs_to_organisation, unless: :current_action_is_valid?
+  before_action :redirect_user_with_no_organisation, unless: :current_action_is_valid?
 
   helper_method :current_organisation, :super_admin?
 
@@ -33,23 +33,18 @@ protected
     devise_parameter_sanitizer.permit(:accept_invitation, keys: [:name])
   end
 
-  def ensure_user_belongs_to_organisation
-    if current_user && current_user.organisations.empty?
-      message = 'Please select your organisation'
-
-      redirect_to new_organisation_path, notice: message
+  def redirect_user_with_no_organisation
+    if current_user&.organisations&.empty?
+      msg = "You do not belong to an organisation. Please mention this in your support request."
+      redirect_to signed_in_new_help_path, notice: msg
     end
   end
 
   def current_action_is_valid?
-    valid_actions_for_orphaned_user.fetch(controller_name, []).include?(action_name)
+    controller_name == 'help' && valid_help_actions.include?(action_name)
   end
 
-  def valid_actions_for_orphaned_user
-    @valid_actions_for_orphaned_user ||= {
-      'current_organisation' => %w(edit),
-      'organisations' => %w(new create),
-      'help' => %w(new create technical_support user_support admin_account)
-    }.freeze
+  def valid_help_actions
+    @valid_help_actions ||= %w(signed_in create).freeze
   end
 end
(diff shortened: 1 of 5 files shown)


app/views/layouts/_google_analytics.html.erb
```diff
@@ -4,7 +4,7 @@
   <script>
     window.dataLayer = [{
       'signedIn': '<%= user_signed_in?.to_json %>',
-      'hasIPs': '<%= (!!current_user && current_organisation.ips.any?).to_json %>'
+      'hasIPs': '<%= (!!current_user && current_organisation && current_organisation.ips.any?).to_json %>'
```
tommotaylor: Rather than `current_organisation && current_organisation.ips.any?` can we use `current_organisation&.ips&.any?`

Same applies anywhere we're using `current_organisation && current_organisation.ips.any?`

spec/features/user_with_no_organisation_memberships_spec.rb
```diff
@@ -0,0 +1,28 @@
+describe 'A confirmeduser with no organisation memberships logs in', type: :feature do
```
tommotaylor: Spelling `confirmeduser`

steventux:
@tommotaylor I think I've covered all your review comments now, I've redirected to the change organisation page with the suggested message.

app/controllers/application_controller.rb
```diff
@@ -31,4 +32,15 @@ def authorise_manage_locations
   def configure_devise_permitted_parameters
     devise_parameter_sanitizer.permit(:accept_invitation, keys: [:name])
   end
+
+  def ensure_user_belongs_to_organisation
+    if current_user&.organisations&.empty?
+      msg = "You do not belong to an organisation. Please mention this in your support request."
+      redirect_to user_support_new_help_path, notice: msg
```
emileswarts: The public help page is 
```
new_help_path
```

app/controllers/application_controller.rb
```diff
@@ -31,4 +32,19 @@ def authorise_manage_locations
   def configure_devise_permitted_parameters
     devise_parameter_sanitizer.permit(:accept_invitation, keys: [:name])
   end
+
+  def ensure_user_belongs_to_organisation
+    if current_user&.organisations&.empty?
+      msg = "You do not belong to an organisation. Please mention this in your support request."
+      redirect_to new_help_path, notice: msg
```
emileswarts: Cool looks good

app/controllers/application_controller.rb
```diff
@@ -31,4 +32,19 @@ def authorise_manage_locations
...
+    end
+  end
+
+  def current_action_is_valid?
+    controller_name == 'help' && valid_help_actions.include?(action_name)
+  end
+
+  def valid_help_actions
+    @valid_help_actions ||= %w(admin_account technical_support user_support signed_in new create).freeze
+  end
```
emileswarts: Given the new_help_path is public, we won't need these two methods above.

app/views/application/_subnavigation.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class='govuk-grid-row subnav govuk-!-margin-0'>
   <div class='govuk-grid-column-one-third govuk-grid-column-one-half-from-desktop govuk-!-padding-left-0 organisation-name'>
     <p class='govuk-body govuk-!-margin-0 govuk-!-padding-0'>
-      <strong class="govuk-!-margin-right-1"><%= current_organisation.name %></strong>
+      <strong class="govuk-!-margin-right-1"><%= current_organisation.name if current_organisation %></strong>
```
emileswarts: Also no need to modify these files as we will redirect to the public page.
We can safely assume that anyone who is logged in has an organisation at this point.

PR #763, opened 2019-06-03
Refactor LogsSearch validation

steventux:
Part of https://trello.com/c/oCYLBJ0y/9-reduce-sentry-errors-to-zero

Spotted there might be validation problems from [a production Sentry error](https://sentry.io/organizations/government-digital-services/issues/855105327/?environment=production&project=1259357&query=is%3Aunresolved)

The fix involved a bit of a refactor to use Rails validation helpers.

### DIFF
--- a/app/models/logs_search.rb
+++ b/app/models/logs_search.rb
@@ -1,45 +1,19 @@
 class LogsSearch
   include ActiveModel::Model
 
-  attr_accessor :filter, :first_step
-  attr_writer :term
+  attr_accessor :filter, :first_step, :search_term
 
-  validate :validate_term, :validate_presence
-
-  def term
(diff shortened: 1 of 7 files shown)


app/models/logs_search.rb
```diff
@@ -1,45 +1,19 @@
...
 
-  validate :validate_term, :validate_presence
-
-  def term
-    @term&.strip
-  end
+  validates :search_term, presence: { message: 'cannot be empty' }
+  validates :search_term,
+            length: { in: 5..6, message: 'must be 5 or 6 characters' },
+            if: -> { filter == 'username' }
```
emileswarts: very nice

PR #764, opened 2019-06-03
WIP: Improve DB queries

georgeschena:


georgeschena:
@steventux Yeah I agree with you, I will add it into the test environment so when we run tests we can catch these N+1 queries during testing.

PR #766, opened 2019-06-04
Fix duplicate email sent on first invite

tommotaylor:
Fixes the issue where a new user was receiving both an invitation email based on them being a new user and a membership invitation email based on them being invited to an organisation.

Eventually it would be nice to only have one way of being invited to an organisation. Currently you can either be invited via the `User` (devise, for new users) or via the `Membership` (ours, for existing users). But that can be another PR

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -31,17 +31,19 @@ private
   def add_user_to_organisation(organisation)
     membership = invited_user.memberships.find_or_create_by(invited_by_id: current_user.id, organisation: organisation)
     membership.update_attributes(can_manage_team: params[:can_manage_team], can_manage_locations: params[:can_manage_locations])
-    send_membership_invite_to_confirmed_users_only(membership)
+    send_invite_email(membership) if user_has_confirmed_account?
   end
 
-  def send_membership_invite_to_confirmed_users_only(membership)
-    if invited_user.confirmed?
-      AuthenticationMailer.membership_instructions(
-        invited_user,
-        membership.invitation_token,
-        organisation: current_organisation
-      ).deliver_now
-    end
+  def send_invite_email(membership)
+    AuthenticationMailer.membership_instructions(
+      invited_user,
+      membership.invitation_token,
+      organisation: current_organisation
+    ).deliver_now
+  end
+
+  def user_has_confirmed_account?
+    invited_user.confirmed?
   end
 
   def after_path(organisation)
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -21,22 +21,27 @@ class Users::InvitationsController < Devise::InvitationsController
       return
     end
 
-    invite_user(organisation)
+    add_user_to_organisation(organisation)
 
     redirect_to(after_path(organisation), notice: "#{invited_user.email} has been invited to join #{organisation.name}")
   end
 
 private
 
-  def invite_user(organisation)
+  def add_user_to_organisation(organisation)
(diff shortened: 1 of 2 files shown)


spec/features/memberships/invite_user_to_their_first_organisation_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      before do
+        membership = user.membership_for(organisation)
+        visit confirm_new_membership_path(token: membership.invitation_token)
+        fill_in "name", with: "Invitee"
+        fill_in "password", with: "password"
+        click_on "Save"
+      end
+
+      it 'allows the user to sign in successfully' do
+        expect(page).to have_content 'Sign out'
```
tommotaylor: This is failing:
```
Failure/Error: invitation.confirm!

     NoMethodError:
       undefined method `confirm!' for nil:NilClass
     # ./app/controllers/users/memberships_controller.rb:4:in `create'
```

app/controllers/users/invitations_controller.rb
```diff
@@ -21,22 +21,27 @@ def create
       return
     end
 
-    invite_user(organisation)
+    add_user_to_organisation(organisation)
```
emileswarts: I didn't mind the invite_user method name.
add_user_to_organisation sounds to me like it's a done deal and no further action is required.

Keen to hear your thoughts.

app/controllers/users/invitations_controller.rb
```diff
@@ -21,22 +21,27 @@ def create
...
     membership = invited_user.memberships.find_or_create_by(invited_by_id: current_user.id, organisation: organisation)
     membership.update_attributes(can_manage_team: params[:can_manage_team], can_manage_locations: params[:can_manage_locations])
-
-    send_invitation_email(membership)
+    send_membership_invite_to_confirmed_users_only(membership)
   end
 
-  def send_invitation_email(membership)
-    AuthenticationMailer.membership_instructions(invited_user, membership.invitation_token, organisation: current_organisation).deliver_now
+  def send_membership_invite_to_confirmed_users_only(membership)
```
emileswarts: The plural 'users' sounds like it could send this out to more than one user.
What about 'notify_invited_to_team' ?

Just thinking out loud here, but this name applies only to existing users.
You won't be invited to a team if you created the account yourself.

spec/features/memberships/invite_user_to_their_first_organisation_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'support/invite_use_case'
```
emileswarts: The spec I was just working on:
spec/features/team/accept_an_invitation_and_sign_up_spec.rb

Seems to assert on this too.  Wonder if we can combine these specs?
I know there is a large name refactor pending.

app/controllers/users/invitations_controller.rb
tommotaylor: Yeah makes sense. But I want to have really well explained methods in this controller. Its already super confusing.

app/controllers/users/invitations_controller.rb
tommotaylor: Yeah I don't really mind either way. My thinking here was that we've got  `invitations` for new users and `membership invites` for existing users. These are the two pathways. So rather than rely on a word that could be misconstrued, I went for something super literal which is the creation of the relationship.

spec/features/memberships/invite_user_to_their_first_organisation_spec.rb
tommotaylor: Theres loads we can do here. Lets leave it for now, fix the edge cases, then improve the specs.

PR #775, opened 2019-06-06
Lookup user contact details from username

steventux:
https://trello.com/c/ydPR0kkx/33-l-allow-gds-super-admin-to-find-user-details-of-a-username-and-vice-versa-7

Screenshot from 2019-06-06 13-24-01

Allows a super admin to search for users by email, phone number or username and retrieve useful contact information.
Provides a link to search logs for the resulting user.

See https://github.com/alphagov/govwifi-terraform/pull/259 for terraform changes to allow connection to user db.

### TODO

This search feature doesn't have any navigation yet, we need to decide where it lives.

### DIFF
--- a/app/models/wifi_user.rb
+++ b/app/models/wifi_user.rb
@@ -1,0 +1,10 @@
+class WifiUser < ApplicationRecord
+  establish_connection WIFI_USER_DB
+  self.table_name = "userdetails"
+
+  def self.search(search_term)
+    search_attr = search_term =~ /^[a-z]{5,6}$/i ? :username : :contact
+
+    find_by(search_attr => search_term)
+  end
+end
--- a/mysql_user/Dockerfile
+++ b/mysql_user/Dockerfile
@@ -1,0 +1,3 @@
+FROM mysql:5.7
+
+ADD wifi_user_schema.sql /docker-entrypoint-initdb.d
--- a/spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
+++ b/spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
@@ -1,0 +1,60 @@
+describe 'Lookup wifi user contact details', type: :feature do
+  let(:user) { create(:user, :super_admin) }
+  let(:search_term) { '' }
+  let(:contact) { 'wifi.user@govwifi.org' }
+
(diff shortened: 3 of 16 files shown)


emileswarts:
Looks good!

app/models/wifi_user.rb
```diff
@@ -0,0 +1,10 @@
+class WifiUser < ActiveRecord::Base
```
tommotaylor: This should inheret from `ApplicationRecord` it's a Rails 5 thing.

Or it could just be a PORO that includes `ActiveModel:Model` like the `LogsSearch` class:
```
class LogsSearch
  include ActiveModel::Model
```

spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
```diff
@@ -0,0 +1,60 @@
+describe 'Lookup wifi user contact details', type: :feature do
+  let(:user) { create(:user, :super_admin) }
+  let(:search_term) { '' }
+  let(:contact) { 'wifi.user@govwifi.org' }
+
+  after { WifiUser.destroy_all }
```
tommotaylor: I think this will automatically be cleaned up by RSpec, or is it specifically needed to avoid errors?

steventux: I found I was getting duplicate key errors on username. I agree it's a bit of a smell that something isn't right in the setup.

app/models/wifi_user.rb
steventux: Good point, I'll update to inherit from `ApplicationRecord` as this will still give me the AR finder methods needed for `find_by` and should ensure the resulting SQL query is clean.

spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
tommotaylor: 👍

mysql_user/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM mysql:5.7
```
cybojenix: user database is actually mysql 8.0

steventux: :+1: OK, I'll amend in another PR

PR #778, opened 2019-06-10
setup DatabaseCleaner to properly clean all our databases

cybojenix:
DatabaseCleaner will only clean the default database by default, so expose it to our other databases


emileswarts:
Nice!

PR #783, opened 2019-06-11
Refactor invite and membership specs

tommotaylor:
Improves two things:
- Naming of spec files so that they are easier to understand
- Removes some duplication we had


spec/features/memberships/invite_user_to_second_organisation_spec.rb
```diff
@@ -44,17 +50,21 @@
     end
   end
 
-  context 'when a super admin does not have can_manage_team permissions' do
```
tommotaylor: This was moved to `specs/super_admin/`

spec/features/confirm_membership_invitation_spec.rb
```diff
@@ -1,26 +0,0 @@
-describe 'Confirming an invitation to a new organisation', type: :feature do
-  let(:organisation) { create(:organisation) }
```
tommotaylor: Duplicate of `'Confirming an invite to a second or subsequent organisation'` below

spec/features/invite_existing_user_spec.rb
```diff
@@ -1,65 +0,0 @@
-require 'support/invite_use_case'
```
tommotaylor: Duplicate file

emileswarts:
Nice, looks good

PR #797, opened 2019-06-27
Update logs search to incorporate user details search in radio group

steventux:
https://trello.com/c/8VpExf4X/145-update-design-for-log-search-by-username-or-contact-details

Incorporates the option to search user details into the logs search radio group (super admins only).

Screenshot from 2019-06-27 16-58-56


app/controllers/super_admin/wifi_user_search_controller.rb
```diff
@@ -6,7 +6,7 @@ def search
       @wifi_user = WifiUser.search(search_term.gsub(' ', ''))
     end
 
-    render action: 'lookup'
+    render 'logs_searches/contact'
```
phantas: you are missing `action`, is that an issue?

app/views/logs_searches/contact.html.erb
```diff
@@ -1,8 +1,16 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+<%= render "layouts/form_errors", resource: @search %>
+
+<div class='govuk-grid-row'>
+  <div class='govuk-grid-column-two-thirds'>
+    <%= link_to "Back to search options", new_logs_search_path, class: "govuk-back-link" %>
```
phantas: 😻

app/views/logs_searches/contact.html.erb
```diff
@@ -1,8 +1,16 @@
...
+  </div>
+
+  <h1 class='govuk-heading-l govuk-grid-column-full'>Search for user details</h1>
+
+  <div class='govuk-grid-column-two-thirds'>
     <%= form_with url: super_admin_wifi_user_search_path do |f| %>
       <p>
-        <%= f.label :search_term, 'Username or contact details', class: 'govuk-label', for: :search_term %>
+        <%= f.label :search_term, 'Username, email address or phone number', class: 'govuk-label', for: :search_term %>
```
phantas: it's a bit academic, but isn't it missing IP address?

app/views/logs_searches/contact.html.erb
```diff
@@ -1,8 +1,16 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+<%= render "layouts/form_errors", resource: @search %>
```
phantas: 😻 that we are consolidating super_admin and admin here

phantas: (This was for the file move)

app/controllers/super_admin/wifi_user_search_controller.rb
steventux: Not in this case as it's telling Rails to render a particular ERB partial.
https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials

app/views/logs_searches/contact.html.erb
steventux: The original feature didn't include finding a user by IP address, not sure if we have this on a per user basis though I know we do have MAC address. It's worth considering as an enhancement though.

PR #805, opened 2019-07-02
Provide network admins with clearer support options

vgphb19:
We need to provide network admins with clearer options regarding the support we provide:
-set up GovWifi support

manage Govwifi support
user information support


steventux:
This was merged in https://github.com/alphagov/govwifi-admin/pull/813

PR #812, opened 2019-07-09
Add two factor authentication for Super Admin users

steventux:
https://trello.com/c/s3l4mIVJ/30-enable-2fa-for-govwifi-admin-login-at-least-for-super-admins

Enforces 2FA setup and authentication for super admin users.
The setup step is activated once for super admin users as they log in.

Screenshot from 2019-07-09 15-11-17

They are then asked to authenticate via TOTP on every login (this is configurable).

Screenshot from 2019-07-09 15-12-19

Do not merge until https://github.com/alphagov/govwifi-build/pull/367 and https://github.com/alphagov/govwifi-terraform/pull/269 have been applied.


config/initializers/devise.rb
```diff
@@ -328,4 +330,19 @@
...
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  #
+  # Two factor authentication config
+  #
+  config.max_login_attempts = 3 # Maximum second factor attempts count.
+  config.allowed_otp_drift_seconds = 30 # Allowed TOTP time drift between client and server.
+  config.otp_length = 6 # TOTP code length
+  config.direct_otp_valid_for = 5.minutes # Time before direct OTP becomes invalid
```
robinmitra: We only seem to be using TOTP. Is this for future?

spec/features/authentication/set_up_two_factor_authentication_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  let(:super_admin_organisation) { create(:organisation, super_admin: true) }
+  let(:user) { create(:user, organisations: [super_admin_organisation]) }
+
+  before do
+    sign_in_user(user, pass_through_two_factor: false)
+    visit root_path
+  end
+
+  context "with a super admin user" do
+    it "2FA setup is enforced" do
```
robinmitra: Consider rephrasing:
```suggestion
    it "enforces 2FA setup" do
```

robinmitra:
LGTM but I've observed that after entering the correct TOTP, I can still go back to `http://localhost:8080/users/two_factor_authentication` route for entering TOTP again. This doesn't seem right as I'm already logged in. Ideally, I would expect it to redirect me to the dashboard.

steventux:
OK thanks @robinmitra that's a really helpful review, a few things to fix up here.

steventux:
@robinmitra having looked at the `two_factor_authentication` gem I can see nothing would prevent a user explicitly navigating to the 2FA page and re-entering a TOTP code after setup or auth. We would need to patch gem functionality to avoid this. My feeling is that if someone wants to manually navigate to a 2FA page after authenticating and re-authenticate, we shouldn't worry too much, I'm not sure it's worth the additional effort in preventing this.

PR #821, opened 2019-07-16
Move sub navigation to left hand side

quis:
This is consistent with what GOV.UK Notify does, and what GOV.UK Pay is moving towards.

Our research has shown it has two main advantages:
- more clearly associates which navigation items are associated with the organisation, and which are about the platform generally
- the vertical arrangement is quicker to scan, so new users can get an idea of what they can do once signed in

This commit also makes a bunch of spacing and style tweaks to generally tidy things up and make them clearer. Sorry I couldn’t split them into cleaner, atomic commits.

# Before 

image

# After 

image

### DIFF
--- a/app/assets/stylesheets/components/left_nav.scss
+++ b/app/assets/stylesheets/components/left_nav.scss
@@ -1,0 +1,38 @@
+.leftnav {
+
+  ul {
+
+    margin: 0;
+    padding: govuk-spacing(3) 0 0 0;
+
+    li {
+      list-style: none;
+      margin: 0;
+    }
+
+  }
+
+  a {
+
+    @include govuk-link-common;
+    @include govuk-link-style-no-visited-state;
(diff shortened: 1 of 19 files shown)


quis:
I had to do a bit of work to resolve the conflicts with #824

I think we can achieve the same intent by only linking to the support form when the user is signed in, and by signposting the sign in page from the non-signed-in support page for network admins (see https://github.com/alphagov/govwifi-product-page/pull/199).

I think the distinction between support and admin support is a bit confusing.

The IA for a signed-in user looks a bit like this:
```
            Support
                |
    —————————————————————————
    |           |           |
End user   Admin user   Raise ticket
    |           |           
Connecting   Tech docs
```

For a non-signed-in user it looks like this:
```
       Support
          |
    ——————————————————
    |                |
End user        Admin user —— Tech docs
    |                |
Connecting         Sign in
                     |
               Raise a ticket
```

PR #823, opened 2019-07-17
Allow super admins to destroy team memberships

steventux:
https://trello.com/c/3yPlxHcR/43-allow-superadmins-to-remove-team-members

Screenshot from 2019-07-17 16-07-35

Screenshot from 2019-07-17 16-13-38

### DIFF
--- a/app/assets/stylesheets/components/buttons_and_links.scss
+++ b/app/assets/stylesheets/components/buttons_and_links.scss
@@ -32,3 +32,9 @@
   box-sizing: content-box;
   width: 6ch;
 }
+
+.govuk-table__cell {
+  .govuk-button {
+    margin: 0;
+  }
+}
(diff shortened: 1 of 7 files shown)


robinmitra:
Clicking on the "Back to list" link takes me to "/memberships" route which doesn't actually contain a list of users - it looks like this:

/membership

Same thing happens when the "Save" button after change a permission for a user - it takes me to "/memberships/updated/permissions" route, which doesn't contains a list of users. It looks like this:

memberships/updated/permissions

I'm not sure whether this is intentional though.

steventux:
@robinmitra I think these are valid bugs, I'm not sure they were introduced in this PR though.

steventux:
@robinmitra looking at the update action for memberships the redirect is fixed https://github.com/alphagov/govwifi-admin/blob/master/app/controllers/memberships_controller.rb#L10 to a specific path. I think this should be addressed in subsequent work.

PR #824, opened 2019-07-18
Add admin support link for logged in admins

robinmitra:
## What

The black top navigation bar for GovWifi has a link titled "Support". Currently, it works in a confusing way:

#### Expected behaviour
If you're logged in as an admin, the link is supposed to take you to the **support contact form** (i.e. https://admin.wifi.service.gov.uk/help/new/signed_in). Otherwise, it is supposed to take you to the **product support page** (https://www.wifi.service.gov.uk/support/).

#### Actual behaviour
The link works as per the expected flow, except when you're **not** logged in as an admin but on the admin login page where it still takes you to the **support contact form**.

## Task

Make the "Support" link on the black top navigation bar always take you to the **product support page**, regardless of whether a user is logged in as an admin or not.

Add a new "Admin Support" link specifically for logged in admins, to the end of the current admin portal navigation (see screenshot), which would take you to **support contact form**.

**Before**
GovWifi_Admin

**After**
GovWifi_Admin

### DIFF
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -64,7 +64,7 @@ class HelpController < ApplicationController
 private
 
   def redirect_signed_in_user
-    return redirect_to signed_in_new_help_path if current_user
+    redirect_to signed_in_new_help_path if current_user
   end
 
   def support_form_params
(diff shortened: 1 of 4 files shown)


robinmitra:
[Add Support link to admin portal navigation bar](https://trello.com/c/FRoeoOPj/51-add-support-link-to-admin-portal-navigation-bar)

PR #828, opened 2019-07-23
Rationalise the locations page

quis:
See individual commits for each specific change.

The overall aims of this pull request is to make the locations page clearer by:
- removing redundant information
- changing the prominence of the actions on the page 
- changing the spacing of elements of the page so related things group together more strongly

# Before 

Screen Shot 2019-07-24 at 15 30 11

# After 

Screen Shot 2019-07-24 at 15 19 41

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -17,6 +17,10 @@ class Location < ApplicationRecord
     "#{address}, #{postcode}"
   end
 
+  def sorted_ip_addresses
+    ips.sort_by { |ip| ip.address.split('.').map(&:to_i) }
+  end
+
 private
 
   def validate_postcode
--- a/app/views/application/_radius_details.html.erb
+++ b/app/views/application/_radius_details.html.erb
@@ -18,14 +18,14 @@
     <h3> London: </h3>
     <ul class="plain-list" id="london-radius-ips">
       <% london_ips.each do |ip| %>
-        <li><p class="govuk-body"><span class="ip-address"><% ip.split('.')[0...-1].each do |chunk| -%><%= chunk -%><span class="ip-address-separator">.</span><% end -%><%= ip.split('.').last %></span></li>
+        <li><p class="govuk-body"><%= render partial: 'ips/ip_address', locals: { ip: ip } %></li>
       <% end %>
     </ul>
 
     <h3> Dublin: </h3>
     <ul class="plain-list" id="dublin-radius-ips">
       <% dublin_ips.each do |ip| %>
-        <li><p class="govuk-body"><span class="ip-address"><% ip.split('.')[0...-1].each do |chunk| -%><%= chunk -%><span class="ip-address-separator">.</span><% end -%><%= ip.split('.').last %></span></li>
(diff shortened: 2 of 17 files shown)


app/models/location.rb
```diff
@@ -17,6 +17,10 @@ def full_address
     "#{address}, #{postcode}"
   end
 
+  def sorted_ip_addresses
+    ips.sort_by { |ip| ip.address.split('.').map(&:to_i) }
```
steventux: What's the intention here? Given `ip.address` is a string isn't lexicographical sorting enough?

```ruby
irb(main):001:0> ip1 = '1.1.1.3'
=> "1.1.1.3"
irb(main):002:0> ip2 = '1.1.1.4'
=> "1.1.1.4"
irb(main):003:0> ip3 = '1.1.1.1'
=> "1.1.1.1"
irb(main):004:0> ip4 = '1.1.1.2'
=> "1.1.1.2"
irb(main):005:0> [ip1, ip2, ip3, ip4].sort
=> ["1.1.1.1", "1.1.1.2", "1.1.1.3", "1.1.1.4"]
```

I'm not sure this method is needed.

app/views/application/_radius_details.html.erb
```diff
@@ -18,14 +18,14 @@
     <h3> London: </h3>
     <ul class="plain-list" id="london-radius-ips">
       <% london_ips.each do |ip| %>
-        <li><p class="govuk-body"> <%= ip %></li>
+        <li><p class="govuk-body"><span class="ip-address"><% ip.split('.')[0...-1].each do |chunk| -%><%= chunk -%><span class="ip-address-separator">.</span><% end -%><%= ip.split('.').last %></span></li>
```
steventux: This IP formatting seems like it would be better defined in a ERB partial as it's repeated on line 28.

app/views/ips/_table.html.erb
```diff
@@ -1,53 +1,56 @@
...
       <p class="govuk-body govuk-!-margin-bottom-1">
-        <%= link_to 'Remove this location', location_remove_path(location, remove: true), class: "govuk-link" %>
+        <%= link_to 'Remove this location', location_remove_path(location, remove: true), class: "red-link" %>
       </p>
     <% end %>
   </caption>
 
   <tbody class="govuk-table__body" id="ips-table">
-    <% location.ips.each do |ip| %>
+    <% location.sorted_ip_addresses.each do |ip| %>
```
steventux: See earlier comment, I think lexicographical sorting will do what you want.

app/models/location.rb
quis: Resolving as discussed (lexicographical sorting doesn’t account for `.` characters)

steventux: Great - yes my example is a bit simplisitic ordering when ip begins with `100.` vs. `1.` would be wrong :+1:

app/views/application/_radius_details.html.erb
steventux: As discussed this is being resolved. :+1:

PR #848, opened 2019-08-19
GOV.UK Frontend v3

freesteph:
# Description
We're porting our services to the [new release of GOV.UK Frontend (v3)](https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0). Namely markup changes and some updates to the Gemfile to reduce our usage of the deprecated `sass` gem.

# Testing
1. build the branch;
2. run exploratory testing.


app/assets/stylesheets/utilities/colours.scss
```diff
@@ -1,6 +1,6 @@
-$colour-black: govuk-colour("black");
-$colour-blue: govuk-colour("blue");
-$colour-light-blue: govuk-colour("light-blue");
-$colour-light-grey: govuk-colour("grey-1");
-$colour-turquoise: govuk-colour("turquoise");
-$colour-white: govuk-colour("white");
+$colour-black: govuk-colour('black');
+$colour-blue: govuk-colour('blue');
+$colour-light-blue: govuk-colour('light-blue');
+$colour-light-grey: govuk-colour('dark-grey');
```
phantas: rename this to dark-grey, plz.

app/assets/stylesheets/utilities/colours.scss
```diff
@@ -1,6 +1,6 @@
...
+$colour-turquoise: govuk-colour('turquoise');
```
phantas: is this used for a confirmation panel? If so, we should use green instead according to https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0

app/assets/stylesheets/utilities/colours.scss
```diff
@@ -1,6 +1,6 @@
-$colour-black: govuk-colour("black");
```
phantas: Can you please add a card to tech debt for refactoring and add an item to see if we can get rid of this file?

app/assets/stylesheets/utilities/colours.scss
freesteph: no I believe we use it for our flash messages, we don't sport any confirmation panel as far as I can tell.

app/assets/stylesheets/utilities/colours.scss
freesteph: Sure.

PR #849, opened 2019-08-21
Simplify and rename the ‘setup’ page

quis:
There’s a lot of confusing duplication between the setup page and the documentation, which:
- makes it harder to pick out the relevant information on this page
- makes it harder for the user to know where they should be looking for information

We think that the documentation can’t (and shouldn’t) be eliminated, because it:
- covers stuff that’s not included on these pages
- has all the nice navigation and formatting afforded by the tech docs tool
- doesn’t require a login

Users will have to have both open side by side – as they do now – so this page should focus on the stuff that we can only show to someone’s who’s logged in.

I’ve had a chat with the technical writer (Pippa) about this and she agrees.

***

The second aim of stripping back this page is to emphasise the memorandum of understanding. Currently we have a problem with not enough people signing it, which we hypothesise is because they’re not noticing it or because it’s presented at the wrong point in their journey.

I expect we’ll have to do more on this front, but this should get us to a place where we can get useful research feedback.

# Before

localhost_8080_setup_instructions(iPad)

# After 

localhost_8080_setup_instructions(iPad) (1)


phantas:
@quis this is currently failing to build (https://cd.gds-reliability.engineering/teams/govwifi/pipelines/admin-pr/jobs/lint%20%26%20test/builds/503)

Have you tried running `make lint`  locally? It's failing on lint with 4 errors:
```
spec/features/ips/view_ip_addresses_spec.rb:16:13: C: RSpec/ContextWording: Start context description with 'when', 'with', or 'without'. (http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording)
    context 'visiting the root path' do
            ^^^^^^^^^^^^^^^^^^^^^^^^
spec/features/ips/permissions_spec.rb:28:1: C: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body end. (https://github.com/rubocop-hq/ruby-style-guide#empty-lines-around-bodies)
spec/features/guidance_after_sign_in_spec.rb:17:5: C: RSpec/LetSetup: Do not use let! to setup objects not referenced in tests. (http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/LetSetup)
    let!(:location) { create(:location, organisation: user.organisations.first) }
    ^^^^^^^^^^^^^^^
spec/features/team/permissions_spec.rb:47:13: C: RSpec/ContextWording: Start context description with 'when', 'with', or 'without'. (http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording)
    context 'visiting the new user invitation page' do
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

quis:
What the page looks like when the MOU has been uploaded:
image

PR #850, opened 2019-08-27
Feature/gaap headers

freesteph:
# Description
Update headers to match GaaP headers. See https://github.com/alphagov/govwifi-product-page/pull/212 for a  full explanation.

# Screenshots
## Before (public v3)
localhost_8080_super_admin_organisations

## After (GaaP v3)
localhost_8080_super_admin_organisations (1)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,5 +49,5 @@ group :development, :test do
   gem 'byebug', '~> 11'
   gem 'listen', '~> 3'
   gem 'pry'
-  gem 'rack-mini-profiler', github: "MiniProfiler/rack-mini-profiler", require: false
+  gem 'rack-mini-profiler', require: false
 end
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -2,6 +2,27 @@ $govuk-image-url-function: "image-url";
 $govuk-font-url-function: "font-url";
 
 @import "../../../node_modules/govuk-frontend/govuk/all";
+
+
+.govuk-header__content,
+.govuk-header__logo {
+  @include govuk-media-query($from: desktop) {
+    width: 50%;
+  }
+}
+
+// remove the default 5px vertical padding on the nav as it
+// pushes it down in an uncentered fashion
+.govuk-header__navigation--end {
+  padding: 0;
+}
+
+// let the beta phase banner breathe and align it
+.govuk-phase-banner__content__tag {
+  margin-left: 5px;
+  vertical-align: top;
+}
+
 @import "./utilities/colours";
 @import "./components/*";
 @import "./utilities/*";
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -54,24 +54,25 @@
                 class="govuk-header__logotype-crown"
                 xmlns="http://www.w3.org/2000/svg"
                 viewbox="0 0 132 97"
-                height="31"
-                width="35">
+                height="30"
+                width="36">
                 <path
                   fill="currentColor" fill-rule="evenodd"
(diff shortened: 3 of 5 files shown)


Gemfile
```diff
@@ -49,5 +49,5 @@ group :development, :test do
   gem 'byebug', '~> 11'
   gem 'listen', '~> 3'
   gem 'pry'
-  gem 'rack-mini-profiler', github: "MiniProfiler/rack-mini-profiler", require: false
+  gem 'rack-mini-profiler', require: false
```
phantas: Sounds like you can also remove `git` from https://github.com/alphagov/govwifi-admin/blob/65580fd94da019c22b2fb0b51960bd90f5b27629/Dockerfile#L39, based on https://github.com/alphagov/govwifi-admin/commit/3bb369952064434c6427fb9269e489ed545129bb

app/assets/stylesheets/application.scss
```diff
@@ -2,6 +2,27 @@ $govuk-image-url-function: "image-url";
 $govuk-font-url-function: "font-url";
 
 @import "../../../node_modules/govuk-frontend/govuk/all";
+
+
+.govuk-header__content,
+.govuk-header__logo {
+  @include govuk-media-query($from: desktop) {
+    width: 50%;
```
phantas: out of curiosity, where is this 50% coming from?

app/views/layouts/application.html.erb
```diff
@@ -54,24 +54,25 @@
                 class="govuk-header__logotype-crown"
                 xmlns="http://www.w3.org/2000/svg"
                 viewbox="0 0 132 97"
-                height="31"
-                width="35">
+                height="30"
```
phantas: just checking, isn't the height 32? https://github.com/alphagov/product-page-example/blob/master/source/stylesheets/modules/_govuk-logo.scss

app/views/layouts/application.html.erb
```diff
@@ -81,7 +82,7 @@
             Menu
           </button>
           <nav>
-            <ul id="navigation" class="govuk-header__navigation " aria-label="Top Level Navigation">
+            <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
```
phantas: Admin is not "active" here:
Screen Shot 2019-08-27 at 20 06 29

do you also need something like this? https://github.com/alphagov/govwifi-product-page/blob/555747c93081cd771869fdb0e77e5009ab0300fd/config.rb#L50-L53

phantas: Admin actually seems to disappear and get replaced by Sign out once you log in. Not sure what the accessibility says on that (although it's not on scope to change it now)

app/assets/stylesheets/application.scss
freesteph: this is to make the banner fit on one line, it's adapted from https://github.com/alphagov/product-page-example/blob/transition-to-govuk-frontend/source/stylesheets/modules/_header.scss

app/views/layouts/application.html.erb
freesteph: I aligned it with what we have in the product page repo, but happy to align onto whichever is the correct one. :triangular_ruler:

app/views/layouts/application.html.erb
freesteph: the active menu item functionality was done for the Product Page in the same PR as the new headers. In hindsight it was out of scope, so decided to not attempt it here. Are we happy with a new ticket to track that work?

Gemfile
freesteph: done :heavy_check_mark:

app/views/layouts/application.html.erb
phantas: I defer it to you, you will know best what is the correct one. I am happy either way, if this is what's on the product page also.

PR #861, opened 2019-09-02
Set focus styles for red links

quis:
These were missing, making the page harder to use for keyboard users.

Also equalises the horizontal padding on the ‘rotate secret key’ links, and does so using the GOV.UK Frontend mixins, rather than hard coding the spacing values.

# Before

image

# After

image


freesteph:
@quis have you run this past Steve? I'm happy with it, so if he is too we can merge this.

PR #880, opened 2019-09-20
Fix new IP addresses not being published

quis:
We’ve had a couple of users report that changes users have made to their IP addresses haven’t taken effect.

As far as I can tell this line of code sets off a process of re-publishing some kind of configuration file which contains all the IP addresses (see here: https://github.com/alphagov/govwifi-admin/blob/master/lib/use_cases/radius/generate_radius_ip_whitelist.rb#L15-L20). So I’m fairly sure that the reason for the bug is that the process is not being triggered when someone changes or adds an IP address.

This happened because adding the IP addresses used to happen at the same time as adding the location, and the code to publish the changes still lives there.

I’m not sure why this bug wasn’t present before, because the publishing looks like it wasn’t happening in the old `update` method: https://github.com/alphagov/govwifi-admin/pull/831/files#diff-54b4e4f2b562d74dad3af78e66e74e4fR26

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -13,7 +13,6 @@ class LocationsController < ApplicationController
     )
 
     if @location.save
-      Facades::Ips::Publish.new.execute
       redirect_to(
         @location.ips.any? ? created_location_with_ip_ips_path : created_location_ips_path,
         notice: "Added #{@location.full_address}"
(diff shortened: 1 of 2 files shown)


quis:
The way to prove that this code is the cause of the reported problems would be if someone could:
1. Look at the config file in the S3 bucket, confirm that the affected IPs are missing
2. Add a new location (which should trigger the publishing process)
3. Confirm that the missing IPs are now in the config file

PR #904, opened 2019-11-05
Feature/superadmin dashboard

freesteph:
# Description
This introduces a new, empty dashboard that is accessible to super-admins, separate from the current super-admin panel. The admins are now understood as:

* "classic" super-admins: belong to the super-admin organisation;
* new admins: have the `is_super_admin` boolean toggled in their attributes.

This allows us to progressively port the features from the current super-admin views, and ultimately delete our super-admin organisation as it is *not* an organisation in the scope of GovWifi.

# Testing
Half of the diff's additions are new tests but everyone is encouraged to give it a test drive.

1. Clone the branch;
2. Log in as a normal user, mess around;
3. Log in as a classic admin, mess around;
4. Log in as a new admin,  mess around.

See `db/seeds.rb` for account details.


robinmitra:
@freesteph We don't seem to be using `cancancan` and `Ability` class at the moment. Are these here just for future implementation?

app/views/application/_subnavigation.html.erb
```diff
@@ -4,6 +4,9 @@
       <strong class="govuk-body"><%= current_organisation.name if current_organisation %></strong>
     </div>
     <nav class='govuk-grid-column-one-half'>
+      <% if can? :manage, :all %>
+        <%= link_to 'Manage GovWifi', super_admin_neo_dashboard_path, class: "govuk-link govuk-body" %>
```
robinmitra: Seems a bit tight, this should help:
```suggestion
        <%= link_to 'Manage GovWifi', super_admin_neo_dashboard_path, class: "govuk-link govuk-body govuk-!-margin-right-3" %>
```

freesteph: ✅

db/seeds.rb
```diff
@@ -31,6 +31,12 @@ def create_user_for_organisations(
   name: 'Academy for Social Justice Commissioning', service_email: 'empty@example.net'
 )
 
+create_user_for_organisations(
+  [],
+  email: 'loneadmin@gov.uk',
+  confirmed_at: Time.zone.now,
+).update!(is_super_admin: true)
+
```
phantas: I can't change the rest of the file, but I recommend making the create_user take the `is_super_admin` flag and setting it explicitly both for loneadmin and admin (below)
See https://github.com/alphagov/govwifi-admin/compare/add-empty-superadmin-dashboard#diff-f7a25bc30a409fc77050703fc048ca8c

app/models/ability.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+class Ability
+  include CanCan::Ability
+
+  def initialize(user)
+    if user.present?
+      can :administrate, :all # FIXME: we can probaly scope this to user orgs
```
phantas: do we need this?

app/controllers/home_controller.rb
```diff
@@ -1,11 +1,18 @@
...
-    elsif new_super_admin?
-      redirect_to super_admin_neo_dashboard_path
-    else
-      redirect_to(current_organisation.ips.empty? ? new_organisation_setup_instructions_path : overview_index_path)
-    end
+    destination =
+      if current_user.super_admin?
+        if current_user.new_super_admin?
+          super_admin_neo_dashboard_path
+        else
```
phantas: I am not sure we need this else but let's check afterwards

app/controllers/super_admin_controller.rb
```diff
@@ -2,6 +2,6 @@ class SuperAdminController < ApplicationController
   before_action :authorise_admin
 
   def authorise_admin
-    redirect_to root_path unless super_admin? || new_super_admin?
+    redirect_to root_path unless super_admin?
```
phantas: fantastic stuff, no need to have the code littered with this logic!! 😍

app/controllers/home_controller.rb
```diff
@@ -1,7 +1,20 @@
 class HomeController < ApplicationController
   def index
-    return redirect_to super_admin_organisations_path if super_admin?
+    destination =
+      if current_user.super_admin?
+        if current_organisation.nil?
```
freesteph: this should be `if user.new_super_admin?`

app/models/ability.rb
freesteph: it's not used much but it is used.

app/controllers/home_controller.rb
freesteph: as per our call, this was needed for routing purposes 🔀

PR #888, opened 2019-10-02
Tidy up the create/switch organisation pages 

quis:
This PR removes stuff that is organisation-specific from pages which are not specific to an organisation (ie the pages you use to switch to another organisation, or to create an organisation).

It also fixes some broken things on the create organisation page:
- incorrect column width
- the [accessible autocomplete](https://github.com/alphagov/accessible-autocomplete) which wasn’t getting initialised 

# Before 

image

# After 

image

PR #919, opened 2019-12-11
Use autocomplete for organisations on signup page

freesteph:
# Description 
See 37ebfd. A missing ID was getting the auto-complete confused.

## Before

User has to scroll/guess through ~1000 organisations.

Screenshot 2019-12-11 at 18 55 12

## After

User can find "District" even though it's not the first word.

after

# Ticket
https://trello.com/c/bDJoUWHR/225-when-an-administrator-sign-up-use-an-auto-complete-to-find-their-organisation

### DIFF
--- a/app/views/users/confirmations/show.html.erb
+++ b/app/views/users/confirmations/show.html.erb
@@ -13,7 +13,7 @@
           <div class="govuk-form-group <%= field_error(resource, "organisation.name") %>">
             <%= organisation_form.label :name, "Organisation name", class: "govuk-label" %>
             <div class="govuk-hint">If your organisation name does not appear below <%= link_to "contact us", technical_support_new_help_path, class: "govuk-link" %>.</div>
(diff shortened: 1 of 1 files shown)


quis:
Hey @freesteph would you consider merging https://github.com/alphagov/govwifi-admin/pull/888 instead? It has some extra changes to style the autocomplete and bring in the latest version. I didn’t merge it before because I ran out of time trying to work out how to remove the left navigation from the two factor pages.

freesteph:
@quis yep sorry I'll have a look now.

freesteph:
@quis let's merge this one first if that's OK? I want to have a closer look at your PR (and get it merged, maybe as soon as tomorrow) but that'll take me a bit of time.

freesteph:
Turns out this PR isn't a simple as I thought as the tests are failing rather obscurely. I'll just focus on #888 instead.

PR #928, opened 2020-01-07
Fix/restore accessible autocomplete

freesteph:
# Description
Some of our pages (creating a new org, confirmation page after signup) weren't using the correct `select` identifier, preventing the library from correctly mounting itself on the page. This is now fixed and the pages are progressively enhanced if possible.

This work is a merge/rebase of #888 and #919 which address the two pages separately.

Also disable the sidebar in some pages with the new interface (i.e the `sidebar` and `subnav` controller methods introduced in the Superadmin dashboard PR (#904), as opposed to the work in #888).

# Before/After
See #888 and #919 for screenshots.


freesteph:
@phantas thanks! I was going to refocus on my original PR but I managed to backport the other autocomplete (new org page) and a couple visual tweaks from #888 without too much effort. Made a new PR as it's changed quite a bit.

PR #929, opened 2020-01-07
fix: restrict session lifetime to 30 minutes

freesteph:
# Description
We were allowing session lasting a full hour, which isn't optimal with regards to security. Downgrade back 30 minutes and update the relevant tests.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -225,7 +225,7 @@ Devise.setup do |config|
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this
   # time the user will be asked for credentials again. Default is 30 minutes.
-  config.timeout_in = 1.hour
+  config.timeout_in = 30.minutes
 
   # ==> Configuration for :lockable
   # Defines which strategy will be used to lock an account.
--- a/spec/features/logout_user_after_inactivity_spec.rb
+++ b/spec/features/logout_user_after_inactivity_spec.rb
@@ -3,11 +3,11 @@ require 'timecop'
 describe 'Logout users after period of inactivity', type: :feature do
   let(:user) { create(:user, :with_organisation) }
 
-  context 'when a signed in user has been inactive for 59 minutes' do
+  context 'when a signed in user has been inactive for 29 minutes' do
     before do
       sign_in_user user
       visit root_path
-      Timecop.travel(Time.now + 59.minutes) { visit root_path }
+      Timecop.travel(Time.now + 29.minutes) { visit root_path }
     end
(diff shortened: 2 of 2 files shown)


quis:
To be compatible with WCAG 2.1<sup>1</sup> you need to either:
- warn users when their session will timeout
- have a session timeout of 20 hours or longer

They say:
> The best way to conform to this success criteria is to keep the user data for at least 20 hours.

On Notify we went for 20 hours option:
https://github.com/alphagov/notifications-admin/blob/91d39b3fdf2d95a8f91d1dad1f898b6f365a730e/app/config.py#L59

***

1. https://www.w3.org/WAI/WCAG21/Understanding/timeouts.html

phantas:
> To be compatible with WCAG 2.11 you need to either:
> 
> * warn users when their session will timeout
> * have a session timeout of 20 hours or longer
> 
> They say:
> 
> > The best way to conform to this success criteria is to keep the user data for at least 20 hours.
> 
> On Notify we went for 20 hours option:
> https://github.com/alphagov/notifications-admin/blob/91d39b3fdf2d95a8f91d1dad1f898b6f365a730e/app/config.py#L59
> 
> 1. https://www.w3.org/WAI/WCAG21/Understanding/timeouts.html

Ah!, thanks for the catch!, I'll have a chat with the PM and suggest we do the same.

config/initializers/devise.rb
```diff
@@ -225,7 +225,7 @@
   # ==> Configuration for :timeoutable
   # The time you want to timeout the user session without activity. After this
   # time the user will be asked for credentials again. Default is 30 minutes.
-  config.timeout_in = 1.hour
+  config.timeout_in = 30.minutes
```
phantas:
```suggestion
  config.timeout_in = 20.hours
```
See Chris's comment

phantas: Cancelling this suggestion, let's merge 🚀

quis: Would have been nice to see the commit message updated with the rationale for 30 minutes as the ultimate choice 🙃

phantas: Good call:
As per the discussion on https://docs.google.com/spreadsheets/d/15AiWew6boVL3tGa0Xw130p2UlajT1jRGKTQzYkthYnc/
The 20 hour requirement is AAA; our level of compliance is AA. Since W3C recommendation contradicts the OWASP (security body) recommendation, there is further discussion going on on the sidelines. In the meantime, we are going to meet the EOW deadline for the ITHC issue raised and consider if we want to meet AAA in the future.

PR #931, opened 2020-01-08
Implement the ability for admins to be able to reset 2FA for others

robinmitra:
## What is it?

This PR adds the functionality to allow org admins and super admins to be able to reset two-factor authentication device for other admin users. 

## Why is it needed?

We already require super admins to login with 2FA, and will soon allow normal admins to do the same. Currently, there isn't a way to reset 2FA device for super admins from the web admin portal, and devs have to do the dirty work of doing this directly on the database! This PR adds the necessary options to the web admin portal, so admins can easily do this themselves (as it should be!). This will become much more important once we roll out 2FA for org admins.

## How does it work?

For super admins, a button that resets 2FA has been added next to each member of an organisation, on the organisation detail pages.

GovWifi_Admin

For normal admins, a link that resets 2FA has been added next to each member of their organisation who have 2FA already enabled on their account, on the membership listing page. The link isn't shown next to the current user's name on the listing.

GovWifi_Admin

### DIFF
--- a/app/controllers/users/two_factor_authentication_setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication_setup_controller.rb
@@ -40,6 +40,14 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
     user.totp_timestamp = nil
     user.otp_secret_key = nil
     user.save!
+
+    redirect_path = if current_organisation&.super_admin?
+                      super_admin_organisations_path
+                    else
+                      memberships_path
+                    end
+
+    redirect_to redirect_path, notice: "Two factor authentication has been reset"
   end
 
   def validate_can_manage_team
--- a/app/controllers/users/two_factor_authentication_setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication_setup_controller.rb
@@ -3,6 +3,11 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
   skip_before_action :handle_two_factor_authentication
   # Skips 2FA setup confirmation callback in ApplicationController.
   skip_before_action :confirm_two_factor_setup
+  before_action :validate_can_manage_team, only: %i[edit destroy]
+
+  def edit
+    @user = User.find(params[:id])
+  end
 
   def show
     # Used to populate the QR code used in setup.
@@ -23,6 +28,28 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
     else
       flash[:alert] = 'Six digit code is not valid'
       render 'show'
+    end
+  end
+
+  def destroy
+    user = User.find(params.fetch(:id))
+    user.second_factor_attempts_count = nil
+    user.encrypted_otp_secret_key = nil
+    user.encrypted_otp_secret_key_iv = nil
+    user.encrypted_otp_secret_key_salt = nil
+    user.totp_timestamp = nil
+    user.otp_secret_key = nil
+    user.save!
+  end
+
+  def validate_can_manage_team
+    if super_admin?
+      return true
+    end
+
+    user = User.find(params.fetch(:id))
+    unless current_user.can_manage_team?(current_organisation) && user.membership_for(current_organisation)
+      raise ActionController::RoutingError.new('Not Found')
     end
   end
 
--- a/app/controllers/users/two_factor_authentication_setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication_setup_controller.rb
@@ -51,13 +51,12 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
   end
 
   def validate_can_manage_team
-    if super_admin?
-      return true
-    end
-
     user = User.find(params.fetch(:id))
-    unless current_user.can_manage_team?(current_organisation) && user.membership_for(current_organisation)
-      raise ActionController::RoutingError.new('Not Found')
+
+    if current_user.can_manage_other_user_for_org?(user, current_organisation)
+      @user = user
+    else
+      redirect_to root_path
     end
   end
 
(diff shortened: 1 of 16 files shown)


app/controllers/users/two_factor_authentication_setup_controller.rb
```diff
@@ -26,6 +31,36 @@ def update
...
+                      super_admin_organisations_path
+                    else
+                      memberships_path
+                    end
+
+    redirect_to redirect_path, notice: "Two factor authentication has been reset"
+  end
+
+  def validate_can_manage_team
+    if super_admin?
```
freesteph: guards read better as one-liners `return true if super_admin?`

app/controllers/users/two_factor_authentication_setup_controller.rb
```diff
@@ -26,6 +31,36 @@ def update
...
+      return true
+    end
+
+    user = User.find(params.fetch(:id))
+    unless current_user.can_manage_team?(current_organisation) && user.membership_for(current_organisation)
+      raise ActionController::RoutingError.new('Not Found')
```
freesteph: We should redirect with `redirect_to` and a 403.

app/controllers/users/two_factor_authentication_setup_controller.rb
freesteph: you probably don't even need to return `true` actually.

app/controllers/users/two_factor_authentication_setup_controller.rb
```diff
@@ -26,6 +31,36 @@ def update
...
+
+    redirect_to redirect_path, notice: "Two factor authentication has been reset"
+  end
+
+  def validate_can_manage_team
+    if super_admin?
+      return true
+    end
+
+    user = User.find(params.fetch(:id))
```
freesteph: I would set the instance variable here rather than fetch the user twice.

app/views/memberships/_table.html.erb
```diff
@@ -52,6 +52,11 @@
         <div class='govuk-grid-column-one-third govuk-list govuk-!-padding-right-4 govuk-!-margin-top-0 text-right'>
           <% if current_user.can_manage_team?(current_organisation) && member.id != current_user.id %>
             <%= link_to 'Edit permissions', edit_membership_path(member.membership_for(current_organisation)), class: "govuk-link govuk-link--no-visited-state" %>
+            <% if member.totp_enabled? %>
+            <%= link_to 'Reset 2FA',
```
freesteph: this should have a 2-space indent since it's part of the if block

config/routes.rb
```diff
@@ -13,6 +13,8 @@
     get 'users/confirmations/pending', to: 'users/confirmations#pending'
     put 'users/two_factor_authentication_setup', to: 'users/two_factor_authentication_setup#update'
     get 'users/two_factor_authentication_setup', to: 'users/two_factor_authentication_setup#show'
+    get 'users/:id/two_factor_authentication/edit', to: 'users/two_factor_authentication_setup#edit'
```
freesteph: we can probably tidy this up with a scoping rule

spec/features/team/reset_2fa_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+      click_on "Reset 2FA"
+      click_on "Yes, reset two factor authentication"
+
+      expect(page).to have_current_path(super_admin_organisations_path)
+    end
+
+    it "shows a notice once 2FA has been reset" do
+      visit super_admin_organisation_path(organisation)
+      click_on "Reset 2FA"
+      click_on "Yes, reset two factor authentication"
```
freesteph: you could contextualise the successful 2FA scenario and then just run the assertion within each test.

app/controllers/users/two_factor_authentication_setup_controller.rb
```diff
@@ -26,6 +31,36 @@ def update
...
 
+  def destroy
+    user = User.find(params.fetch(:id))
+    user.second_factor_attempts_count = nil
+    user.encrypted_otp_secret_key = nil
+    user.encrypted_otp_secret_key_iv = nil
+    user.encrypted_otp_secret_key_salt = nil
+    user.totp_timestamp = nil
+    user.otp_secret_key = nil
+    user.save!
```
freesteph: `user.save!(attr1: val, attr2: val, [etc...])`

spec/features/team/reset_2fa_spec.rb
```diff
@@ -0,0 +1,133 @@
+describe "Reset two factor authentication", type: :feature do
+  let(:organisation) { create(:organisation) }
+  let(:org_admin_user_1) { create(:user, organisations: [organisation]) }
+  let(:org_admin_user_2) { create(:user, organisations: [organisation]) }
+  let(:org_admin_user_3) { create(:user, organisations: [organisation]) }
+
+  let(:super_admin_organisation) { create(:organisation, super_admin: true) }
```
freesteph: `super_admin` is a flag on the user, not the org. You can probably remove this `super_admin_organisation` as it's not always used anymore.

spec/features/team/reset_2fa_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+  let(:super_admin_user) { create(:user, organisations: [super_admin_organisation]) }
```
freesteph: you can use `create(:user, :new_admin)` and you're done. Or `create(:user, :superadmin)` to have the org created but in both cases the 2FA is setup straight away I believe (see the user factory).

spec/support/filters/validate_can_manage_team.rb
```diff
@@ -0,0 +1,41 @@
+RSpec.shared_context 'when using validate_can_manage_team' do
+  let(:organisation) { create(:organisation) }
+
+  let(:superadmin) { create(:user, :super_admin) }
+  let(:admin) { create(:user, organisations: [organisation]) }
+  let(:stranger) { create(:user, :with_organisation) }
```
freesteph: this isn't needed anymore

PR #933, opened 2020-01-10
FIX: Move yml anchor declaration above usage

phantas:
Fixes an error caused by a change to the yml parser used on CI. Similar to alphagov/govwifi-acceptance-tests#25


phantas:
> LGTM. I remember doing a similar change with Rich. It's a bit of a 'gotcha'.

yeah, and I think it came about from an update to the concourse box, which got us talking about dependencies outside the system breaking the pipeline

PR #939, opened 2020-01-15
bau: replace govuk-lint

freesteph:
# Description
Following the instructions on the deprecated govuk-lint gem[1], remove
govuk-lint in favour of `rubocop-govuk` and `scss-lint-govuk`.

# Changes
_not panicking when looking at the diff_

## Double-quote strings

Are now the default. The reason the Ruby world was using single-quotes was for performance reasons (AFAIK) but they are frozen by default in recent versions of Ruby and we have the magic `frozen_string_literals` comment anyway. ERB-lint also applied this in our views.

## Rails.root.join
Is better than `File.join Rails.root, "x", "y", "z"`.

## Time.zone.today
`Date.today` is dangerously generic as it doesn't account for timezone, whereas `Time.zone.today` does.

## Others
Ignored all previous migrations (`db/migrate/2019*`). Ignored some of the `before_action` filters we had on overriden Devise controllers since the action they target is defined in the Devise controller (rather than the subclassed one).

[1]: https://github.com/alphagov/govuk-lint#deprecated


app/controllers/users/invitations_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Users::InvitationsController < Devise::InvitationsController
   before_action :set_target_organisation, if: :super_admin?, only: %i(create new)
   before_action :delete_user_record, if: :user_should_be_cleared?, only: :create
-  after_action :confirm_new_user_membership, only: :update
+  after_action :confirm_new_user_membership, only: :update # rubocop:disable Rails/LexicallyScopedActionFilter
```
phantas: Do you know why this comment is needed? If I understand correctly, `Rails/LexicallyScopedActionFilter` ["checks that methods specified in the filter's only or except options are explicitly defined in the controller." ](https://github.com/rubocop-hq/rubocop/pull/5177) and I can see `confirm_new_user_membership` is defined in this controller

freesteph: This inherits from a Devise controller, and the action is defined there I believe.

PR #941, opened 2020-01-15
fix: don't let unauthenticated users download blobs

freesteph:
# Description
We don't want unauthenticated users to download files so override the ActiveStorage controller to prevent this. See commit and and commit message.

### DIFF
--- a/app/controllers/active_storage/blobs_controller.rb
+++ b/app/controllers/active_storage/blobs_controller.rb
@@ -1,0 +1,39 @@
+# NOTE TO READER: this replacement of ActiveStorage's blob controller
+# is in place to enforce authentication when accessing a blob. At the
+# time of writing, the `/rails/active_storage/blobs/XXX` path is not
+# protected which means the Rails app will happily generate a valid
+# link to the underlying service+asset and redirect anyone who asks
+# there.
+#
+# However ActiveStorage will naturally evolve and probably implement a
+# way to do this without redefining the controller itself but in the
+# meantime it means we've started diverging. This should be kept in
+# sight and ultimately removed.
+#
+# PS: a more direct `before_action: authenticate_user!` was causing an
+# error, probably because the filter is meant to be used in the
+# context of a Rails app rather than its internal engines.
+
+# frozen_string_literal: true
+
+# Take a signed permanent reference for a blob and turn it into an expiring service URL for download.
+# Note: These URLs are publicly accessible. If you need to enforce access protection beyond the
+# security-through-obscurity factor of the signed blob references, you'll need to implement your own
+# authenticated redirection controller.
+class ActiveStorage::BlobsController < ActiveStorage::BaseController
+  include ActiveStorage::SetBlob
+  include Rails.application.routes.url_helpers
+
+  before_action :check_user
+
+  def check_user
+    if current_user.nil?
+      redirect_to new_user_session_path
+    end
+  end
+
+  def show
+    expires_in ActiveStorage::Blob.service.url_expires_in
+    redirect_to @blob.service_url(disposition: params[:disposition])
+  end
+end
--- a/app/controllers/active_storage/blobs_controller.rb
+++ b/app/controllers/active_storage/blobs_controller.rb
@@ -1,18 +1,25 @@
 # NOTE TO READER: this replacement of ActiveStorage's blob controller
-# is in place to enforce authentication when accessing a blob. At the
-# time of writing, the `/rails/active_storage/blobs/XXX` path is not
-# protected which means the Rails app will happily generate a valid
-# link to the underlying service+asset and redirect anyone who asks
-# there.
+# is in place to enforce authentication with devise when accessing a
+# blob. At the time of writing, the `/rails/active_storage/blobs/XXX`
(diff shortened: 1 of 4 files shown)


app/controllers/active_storage/blobs_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  before_action :check_user
+
+  def check_user
+    if current_user.nil?
+      redirect_to new_user_session_path
+    end
+  end
+
+  def show
```
phantas: this does not solve the issue that anyone can see an organisation MOU, as long as they have an admin account. I can see 3 ways to solve this:

1. add a, say, `/mou/signed_mou/` route path to the mou controller that bypasses and does the redirect itself (bit hacky)
2. on show, test `current_organisation.signed_mou.blob`equals `@blob` similar to what you are doing in https://github.com/alphagov/govwifi-admin/commit/e131f7073cb38d604421dd7416e96f84da0fc229
3. or you can do 2. but on `check_user` instead

I think I have a preference for 2.

app/controllers/active_storage/blobs_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+# PS: a more direct `before_action: authenticate_user!` was causing an
+# error, probably because the filter is meant to be used in the
+# context of a Rails app rather than its internal engines.
+
+# frozen_string_literal: true
+
+# Take a signed permanent reference for a blob and turn it into an expiring service URL for download.
+# Note: These URLs are publicly accessible. If you need to enforce access protection beyond the
+# security-through-obscurity factor of the signed blob references, you'll need to implement your own
+# authenticated redirection controller.
```
phantas: This paragraph is a bit confusing as it's stated here, since we are implementing our own authenticated redirection controller, maybe make it explicit this is a reference from the original?

app/controllers/active_storage/blobs_controller.rb
freesteph: Yeah that makes sense. I thought we could get away with just making sure the MOUs are only accessed from within the app but I'll scope it to the correct organisation as well.

freesteph: I don't think 1 is is feasible but I might be confused.

phantas: yeah, my initial impression was that 1. would be too hacky and moves the concern outside the context. I prefer 2. (or 3.)

app/controllers/active_storage/blobs_controller.rb
```diff
@@ -29,6 +36,17 @@ class ActiveStorage::BlobsController < ActiveStorage::BaseController
   def check_user
     if current_user.nil?
       redirect_to new_user_session_path
+    else
+      # current_organisation is not available here
+      org = Organisation.find(
+        ActiveStorage::Attachment
+          .find_by(blob: @blob)
+          .record_id
+      )
```
phantas: How performant is this, do you know?
Personally I am not _too_ concerned if it's a bit slow, if we think we're going to replace this anyway...

freesteph: It is an extra query, you're right, but from testing it it has very little impact on user experience. ActiveStorage already makes a request to the database in the first place, and since this is fetching a single record based on an SQL-indexed-column I don't think it would incur too much delay.

phantas: cool, just making sure it is searching for id or hash over an index

spec/features/upload_download_mous_spec.rb
```diff
@@ -71,7 +71,8 @@
           it "lets them download the MoU" do
             visit @link
 
-            expect(page).to have_current_path(@link)
+            # as we're running locally, the actual asset lives under rails/active_storage/disk
+            expect(page.current_path).to start_with("/rails/active_storage/disk")
```
phantas: Minor thing: I'd prefer the use of a variable instead of the hardcoded path, if one existed, but this is already a bit hacky as is anyway so I am not too bothered under the circumstances.

app/controllers/active_storage/blobs_controller.rb
```diff
@@ -37,15 +37,14 @@ def check_user
...
-        ActiveStorage::Attachment
-          .find_by(blob: @blob)
-          .record_id
-      )
-
-      unless can? :read_mou, org
-        redirect_to overview_index_path, alert: "You are not allowed to see this MoU."
+      attachment = ActiveStorage::Attachment.find_by(blob: @blob)
+
+      if attachment.name == "signed_mou"
```
phantas: Another comment: we don't actually have a test that covers normal admins downloading the template. But we have one for the super admin which covers this enough so, again, I am not bothered if you don't add one.

spec/features/upload_download_mous_spec.rb
robinmitra: I agree but since this probably won't be needed soon, I'm okay with this for now.

app/controllers/active_storage/blobs_controller.rb
freesteph: I've added that now.

PR #943, opened 2020-01-15
Guard against upload of non pdf files

phantas:
- Given a normal admin, reject upload or update of a signed MOU if file is not a PDF
- Given super-admin, reject upload or update of an MOU template if file is not a PDF
- Ensure mou.pdf fixture is a valid pdf and use former plain text fixture as an example of a non valid pdf

Signed-off-by: Paula Valenca <paula.valenca@digital.cabinet-office.gov.uk>


app/controllers/mou_controller.rb
```diff
@@ -7,8 +7,14 @@ def index
   def create
     redirect_path = replacing? ? replaced_mou_index_path : created_mou_index_path
     if params[:signed_mou]
-      current_organisation.signed_mou.attach(params[:signed_mou])
-      flash[:notice] = "MOU uploaded successfully."
+      mime_type = Marcel::MimeType.for(params[:signed_mou])
```
freesteph: the mime type is already available in `params[:signed_mou].content_type`.

app/controllers/super_admin/mou_controller.rb
```diff
@@ -17,17 +17,27 @@ def create
 
   def attach_to_organisation(organisation)
     if params[:signed_mou]
-      organisation.signed_mou.attach(params[:signed_mou])
-      flash[:notice] = "MOU uploaded successfully."
+      mime_type = Marcel::MimeType.for(params[:signed_mou])
```
freesteph: ibid

spec/features/super_admin/upload_download_mou_spec.rb
```diff
@@ -27,14 +40,15 @@
...
-  context "when dowloading the MOU template" do
+  context "when downloading the MOU template" do
     before do
       attach_file("unsigned_document", Rails.root + "spec/fixtures/mou.pdf")
       click_on "Upload"
       click_on "Download current template"
     end
 
     it "downloads the MOU" do
+      convert_pdf_to_page
```
freesteph: I would just assert that the `page.current_path` has changed to `rails/active_storage/disk` as this will effectively point to the backing local URL for that asset. You can also check that it contains the file's name, but anything past that seems like overkill.

app/controllers/mou_controller.rb
phantas: done

app/controllers/super_admin/mou_controller.rb
phantas: done

PR #947, opened 2020-01-16
Make it optional for org admins to be able to setup 2FA

robinmitra:
## What is it?

This will basically present the two factor authentication (2FA) setup screen, as currently shown to super admin users only, to all org admins. At the moment, this is an optional thing, and users can skip setting it up, until next time.

## Why is it needed?

We already require two factory authentication for super admins. This PR extends it to org admins, for better security.

## How will it work?

Once org admins log into the admin portal, they will be presented with a screen to setup 2FA, by scanning a QR code - this is how it works for super admins currently and hasn't changed.

However, there is a new option added for skipping the 2FA setup, until next time, if the user isn't ready for this yet.

---

### Screenshot 1: Option to skip setting up 2FA until next time

GovWifi Admin - Option to skip setting up 2FA until next time

---

### Screenshot 2: Flash message to confirm that 2FA setup has been skipped

GovWifi Admin - Flash message to confirm that 2FA setup has been skipped

### DIFF
--- a/app/controllers/users/two_factor_authentication_setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication_setup_controller.rb
@@ -15,9 +15,9 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
   end
 
   def update
-    if params[:commit] == 'Remind me next time'
+    if params[:commit] == "Remind me next time"
       # Ensures the user doesn't go through 2FA check again.
-      request.env['warden'].session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = false
+      request.env["warden"].session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = false
       flash[:notice] = "Two factor authentication setup skipped until next time"
       redirect_to stored_location_for(:user) || root_path
       return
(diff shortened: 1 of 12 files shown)


app/controllers/users/two_factor_authentication_setup_controller.rb
```diff
@@ -15,6 +15,14 @@ def show
   end
 
   def update
+    if params[:commit] == "Remind me next time"
+      # Ensures the user doesn't go through 2FA check again.
+      request.env["warden"].session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION] = false
```
freesteph: could you factor this and line 31 into one method instead of having the same 2 lines twice?

app/models/user.rb
```diff
@@ -78,7 +78,10 @@ def super_admin?
   end
 
   def need_two_factor_authentication?(request)
-    !ENV.key?("BYPASS_2FA") && request.env["warden"].user.super_admin?
+    return false if ENV.key?("BYPASS_2FA")
+
+    needs_auth = request.env["warden"].session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION]
+    needs_auth.nil? || needs_auth || request.env["warden"].user.super_admin?
```
freesteph: I think it works the same with the simpler:

```ruby
  def need_two_factor_authentication?(request)
    return false if ENV.key?("BYPASS_2FA")

    needs_auth = request
                   .env["warden"]
                   .session(:user)[TwoFactorAuthentication::NEED_AUTHENTICATION]

    (needs_auth != false) || super_admin?
  end
```

PR #959, opened 2020-02-03
Fix/allow password edit

freesteph:
# Description
Allow people to update their password in the settings panel. Also fix the sidebar highlighting (see the Settings item).

# Before
admin staging wifi service gov uk_setup_instructions_initial(iPad) (2)

# After
localhost_8080_setup_instructions(iPad)


app/views/setup_instructions/_servers.html.erb
```diff
@@ -0,0 +1,55 @@
+<h3 class="govuk-heading-m" id="setup-header">GovWifi servers</h3>
+
+<table class="govuk-table govuk-!-margin-bottom-3">
+  <caption class="govuk-visually-hidden" id="setup-header">Settings</caption>
```
robinmitra: Same as above - the final HTML ends up having the same ID multiple times.

PR #961, opened 2020-02-06
Fix/insecure direct object references

freesteph:
# Description
Ensure users can only modify objects (namely: IPs) when they are allowed to. Previously we would just parse the params and retrieve these objects which was a security concern.

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -30,7 +30,15 @@ class LocationsController < ApplicationController
 
   def add_ips
     @location = Location.find(params[:location_id])
-    @location.add_blank_ips(5)
+
+    if !current_organisation.locations.include? @location
+      redirect_to(
+        ips_path,
+        notice: "You are not authorised to edit this location",
+      )
+    else
+      @location.add_blank_ips(5)
+    end
   end
 
   def update_ips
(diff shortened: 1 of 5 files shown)


app/controllers/locations_controller.rb
```diff
@@ -1,5 +1,6 @@
 class LocationsController < ApplicationController
   before_action :authorise_manage_locations
+  before_action :authorise_manage_current_location, only: [:add_ips, :update_ips]
```
robinmitra: Any reason why this doesn't apply to `update` method as well?

spec/controllers/locations_controller_spec.rb
```diff
@@ -1,26 +1,82 @@
...
 
+  before do
+    sign_in user
+  end
+
   describe "GET /add_ips" do
-    before do
-      sign_in user
+    context "when the user does not belong to the organisation" do
+      it "redirects to user locations page" do
```
robinmitra:
```suggestion
      it "redirects to root path" do
```

spec/controllers/locations_controller_spec.rb
```diff
@@ -1,26 +1,82 @@
...
+      let(:other_params) {
+        params[:location_id] = other_location.id
+        params
+      }
+
+      before do
+        @response = post :update_ips, params: other_params
+      end
+
+      it "redirects them to the form with an error" do
```
robinmitra:
```suggestion
      it "redirects them to the root path" do
```

PR #981, opened 2020-03-03
Feature/cookies

freesteph:
# Description
See https://github.com/alphagov/govwifi-product-page/pull/240. 

The difference in approach here is that admin is an actual server (as opposed to a statically generated website) so we want to include the shared frontend assets server-side but do the checking and potential disabling of analytics client-side.

### DIFF
--- a/app/views/layouts/_google_analytics.html.erb
+++ b/app/views/layouts/_google_analytics.html.erb
@@ -3,17 +3,13 @@
 
 <script>
  window.dataLayer = [{
-       'signedIn': '<%= user_signed_in?.to_json %>',
+     'signedIn': '<%= user_signed_in?.to_json %>',
      'hasIPs': '<%= (!!current_user && current_organisation&.ips&.any?).to_json %>'
  }];
 </script>
 
 <!-- Google Tag Manager -->
 <script>
- if (!GovWifi.cookies.isCategoryAllowed("tracking")) {
-     window['ga-disable-<%= SITE_CONFIG["google_analytics_script_id"] %>'] = true;
- }
-
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
--- a/app/views/layouts/_google_analytics.html.erb
+++ b/app/views/layouts/_google_analytics.html.erb
@@ -1,31 +1,35 @@
 <!-- Global site tag (gtag.js) - Google Analytics -->
 <script async src="https://www.googletagmanager.com/gtag/js?id=<%= SITE_CONFIG["google_analytics_script_id"] %>"></script>
 
-  <script>
-    window.dataLayer = [{
-      'signedIn': '<%= user_signed_in?.to_json %>',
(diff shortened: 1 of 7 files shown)


app/views/layouts/_google_analytics.html.erb
```diff
@@ -1,31 +1,31 @@
 <!-- Global site tag (gtag.js) - Google Analytics -->
 <script async src="https://www.googletagmanager.com/gtag/js?id=<%= SITE_CONFIG["google_analytics_script_id"] %>"></script>
 
-  <script>
-    window.dataLayer = [{
-      'signedIn': '<%= user_signed_in?.to_json %>',
-      'hasIPs': '<%= (!!current_user && current_organisation&.ips&.any?).to_json %>'
-    }];
-  </script>
+<script>
```
freesteph: this is juste whitespace.

robinmitra:
Authentication seems to be broken when running locally. The previously merged PR (dependabot) seems to be the culprit.

freesteph:
> Authentication seems to be broken when running locally. The previously merged PR (dependabot) seems to be the culprit.

Ah cheers, I couldn't figure out why master or this branch wouldn't log anymore, thought it was something on my machine.

robinmitra:
The relative link "How GovWifi uses cookies" is broken here - I suggest we link to the absolute URL.

package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "dependencies": {
     "govuk-frontend": "^3.5.0",
+    "govwifi-shared-frontend": "https://github.com/alphagov/govwifi-shared-frontend/releases/download/0.6.2/govwifi-shared-frontend-0.6.2.tgz",
```
robinmitra: We should update the version here as well (and the package lock file).

yarn.lock
```diff
@@ -7,6 +7,17 @@ govuk-frontend@^3.5.0:
...
+  version "0.6.3"
+  resolved "https://github.com/alphagov/govwifi-shared-frontend/releases/download/0.6.3/govwifi-shared-frontend-0.6.3.tgz#6513c83caffbea007c6a7fef8fb89c7e48959166"
+  dependencies:
+    js-cookie "^2.2.1"
+
 html5shiv@^3.7.3:
   version "3.7.3"
   resolved "https://registry.yarnpkg.com/html5shiv/-/html5shiv-3.7.3.tgz#d78a84a367bcb9a710100d57802c387b084631d2"
+
+js-cookie@^2.2.1:
```
robinmitra: Are we using this?

freesteph: in the shared-frontend: https://github.com/alphagov/govwifi-shared-frontend/tree/0.6.1

freesteph: sorry wrong link: https://github.com/alphagov/govwifi-shared-frontend/releases/tag/0.6.1

robinmitra: So we don't need this again in this repo then?

freesteph: The lock file will include all of them.

package.json
freesteph: Yeah for some reason I'd only committed the `.lock` file change, which is strange....

PR #996, opened 2020-03-20
Fix broken styling on various links

alan-gds:
Styling of (for example) the 'memorandum of understanding' reminder link was broken during implementation of the 2FA reset task (PR 931).

https://trello.com/c/eGfyV8mO/69-fix-broken-styling-of-links-on-admin-portal-estimate-small-1-2-days

### DIFF
--- a/app/assets/stylesheets/components/team_members.scss
+++ b/app/assets/stylesheets/components/team_members.scss
@@ -1,9 +1,11 @@
-.list-item-padding {
-  margin-bottom: -5px;
-}
+.membership-list {
+  .list-item-padding {
+    margin-bottom: -5px;
+  }
 
-.govuk-link {
-  display: block;
+  .govuk-link {
+    display: block;
+  }
 }
 
 .govuk-table__row {
--- a/app/views/memberships/_table.html.erb
+++ b/app/views/memberships/_table.html.erb
@@ -1,6 +1,6 @@
 <div class='govuk-section-break--l govuk-section-break--visible govuk-!-margin-top-0 govuk-!-margin-bottom-3'></div>
 
-<ul class='govuk-list govuk-width-container govuk-!-margin-left-0'>
+<ul class='govuk-list membership-list govuk-width-container govuk-!-margin-left-0'>
   <% team_members.each do |member| %>
     <li class='govuk-!-margin-bottom-1'>
       <div class='govuk-grid-row'>


freesteph:
Sorry ignore me I just read the Trello ticket. Just a sec.

app/views/memberships/_table.html.erb
```diff
@@ -51,11 +51,11 @@
 
         <div class='govuk-grid-column-one-third govuk-list govuk-!-padding-right-4 govuk-!-margin-top-0 text-right'>
           <% if current_user.can_manage_team?(current_organisation) && member.id != current_user.id %>
-            <%= link_to "Edit permissions", edit_membership_path(member.membership_for(current_organisation)), class: "govuk-link govuk-link--no-visited-state" %>
+            <%= link_to "Edit permissions", edit_membership_path(member.membership_for(current_organisation)), class: "govuk-single-link govuk-link--no-visited-state" %>
             <% if member.totp_enabled? %>
               <%= link_to "Reset 2FA",
                           { controller: "users/two_factor_authentication_setup", action: "edit", id: member },
-                          class: "govuk-link govuk-link--no-visited-state" %>
+                          class: "govuk-single-link govuk-link--no-visited-state" %>
```
robinmitra: Not using govuk-link will break goodies such as accessibility improvements we get from the library. I suggest the following approach:

* revert back to `govuk-link` on line 54 and 58
* add a new class `membership-list` on line 3 (i.e. `ul` element)
* scope the relevant CSS to this class - see comments on `team_members.scss`.

app/assets/stylesheets/components/team_members.scss
```diff
@@ -2,7 +2,7 @@
   margin-bottom: -5px;
 }
 
-.govuk-link {
+.govuk-single-link {
   display: block;
 }
```
robinmitra: *Please read comments on `_table.html.erb` first and then come back here.*

Revert this back to `govuk-link` and scope `.list-item-padding` and `.govuk-link` to be within `membership-list` class.

E.g.
```
.membership-list {
  .list-item-padding { ... }
  .govuk-link { ... }
}
```

app/views/memberships/_table.html.erb
alan-gds: Good idea, that makes more sense.

alan-gds:
I've followed Robin's suggestion.  I noticed though the the style wasn't actually getting applied on the reset 2FA links (with or without the changes from this task, I believe), as those have `govuk-link--no-visited-state` rather than plain `govuk-link`.  I've amended the CSS accordingly.

app/assets/stylesheets/components/team_members.scss
```diff
@@ -1,8 +1,8 @@
-.list-item-padding {
+.membership-list.list-item-padding {
   margin-bottom: -5px;
 }
 
-.govuk-link {
+.membership-list.govuk-link--no-visited-state {
```
freesteph: since this is SCSS, you could just scope the top class and do your things in there:

```scss
.membership-list {
  .list-item-padding {
    margin-bottom: -5px;
  }

  .govuk-link {
    display: block;
  }
}
```

I've reverted back to styling the `govuk-link` as `--no-visisted-state` will just inherit from it. I believe.

app/views/memberships/_table.html.erb
```diff
@@ -51,11 +51,11 @@
 
         <div class='govuk-grid-column-one-third govuk-list govuk-!-padding-right-4 govuk-!-margin-top-0 text-right'>
           <% if current_user.can_manage_team?(current_organisation) && member.id != current_user.id %>
-            <%= link_to "Edit permissions", edit_membership_path(member.membership_for(current_organisation)), class: "govuk-link govuk-link--no-visited-state" %>
+            <%= link_to "Edit permissions", edit_membership_path(member.membership_for(current_organisation)), class: "membership-list govuk-link--no-visited-state" %>
```
freesteph: I think I would agree with Robin's previous comment here: style the parent `ul` element and then you can just do the styling within the scope of that (as opposed to tagging both elements with it). And then you can group the CSS styling mentioned above.

alan-gds: Good plan - I've now done this.

app/assets/stylesheets/components/team_members.scss
alan-gds: From my testing, it doesn't seem to work as `govuk-link`; I have to declare it as `govuk-link--no-visited-state` in the scss.  If you can get it to inherit, please let me know how!

alan-gds: As per Slack discussion: we've agreed that we need styling on both `govuk-link` and `govuk-link--no-visited-state`.

PR #997, opened 2020-03-20
BAU: stop pinning sprockets dependency

alan-gds:
We pinned sprockets to 4.0.0beta10 during an abortive attempt to
switch to sassc-rails, and did not unpin it when we switched back
to sass-rails.  Dependabot subsequently updated it to 4.0.0 final.
Dependanbot has been misbehaving since then, refusing to update
its own PRs because of an alleged incompaibility between
sass-rails 5.1 (which we are stuck on) and sprockets 4.0.0.

Nothing else depends on sprockets, and we're hoping that unpinning
the dependency will allow a compatible version to be used,
pacifying Dependabot.


alan-gds:
> You need to remove the pinning of `sass-rails` and run `bundle clean --force && bundle install` and commit the Gemfile.lock as well.

As per Slack discussion: I think changing our `sass-rails` dependency is outside the scope of this PR, but you're absolutely right about the `Gemfile.lock`.  I've updated it and force-pushed the changes.

PR #1037, opened 2020-05-12
Upgrade to Rails 6

freesteph:
# Description
We want to inherit all the goodness of Rails 6 so just do that. It seemed like an awfully complicated adventure the last time I've tried because I'd loaded the Rails 6 default config (i.e `config.load_defaults 6.0` and that sets Zeitwerk as the default autoloader for the app, which we don't want because we use loads of things in the `lib/` folder that don't respect the naming conventions Zeitwerk relies on to load files.

As per the rest of the changes, they're minor API updates.

# Testing
- Build the container with the changes;
- Try and break it :sunglasses:


app/models/ip.rb
```diff
@@ -1,7 +1,7 @@
 class Ip < ApplicationRecord
   belongs_to :location
 
-  validates :address, presence: true, uniqueness: true
+  validates :address, presence: true, uniqueness: { case_sensitive: true }
```
robinmitra, line 4: Is this needed as part of the upgrade? Also, the default is `true` anyway.

config/application.rb
```diff
@@ -19,8 +19,12 @@
 module GovwifiAdmin
   class Application < Rails::Application
     config.exceptions_app = self.routes
-    # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.1
+
+    config.load_defaults 6.0
+
+    # our lib/ folder doesn't follow the Zeitwerk conventions for
+    # autloading, so rely the classic loader instead.
```
robinmitra, line 26:
```suggestion
    # autoloading, so rely on the classic loader instead.
```

app/models/ip.rb
freesteph:
```
.DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in
 Rails 6.1. To continue case sensitive comparison on the :address attribute in Ip model, pass 
`case_sensitive: true` option explicitly to the uniqueness validator. (called from block (4 le
vels) in <top (required)> at /usr/src/app/spec/models/ip_spec.rb:66)
```

config/application.rb
freesteph: thanks, amended

PR #1058, opened 2020-06-04
Client-side password validation

alan-gds:
* User is given option to toggle visibility of the password
* User is given live feedback on password strength as they enter it
* User is given suggestions for how to improve the password strength if it is too weak when the password box is unfocused

https://trello.com/c/Pi2rCjuv/182-improve-password-validation-feedback-for-admin-users


app/assets/javascripts/application.js
```diff
@@ -1,5 +1,60 @@
...
+};
+
+if ($('input[name="user[password]"]').length) {
+  $('body').on('keyup', 'input[name="user[password]"]', function(e) {
+    const { score, scoreText } = getPasswordStrength(this.value)
+    const feedback = `Password strength: ${scoreText}`;
+    if (this.value.length) {
+      $('.password-strength')
+        .css('display', 'inline-block')
+        .css('background-position', score + '% 0%')
```
robinmitra, line 26: Don't think this is needed anymore - it's probably for the fancy CSS transition thing I did earlier.

app/assets/javascripts/application.js.erb
```diff
@@ -11,3 +13,55 @@ window.addEventListener("DOMContentLoaded", () => {
...
+      $('.password-strength')
+        .css('display', 'inline-block')
+        .text(feedback);
+    } else {
+      $('.password-strength').css('display', 'none');
+    }
+  });
+
+  $('body').on('blur', 'input[name="user[password]"]', function(e) {
+    const { score, suggestion } = getPasswordStrength(this.value)
```
freesteph, line 41: sorry I'm late to the party, but this should really just send an AJAX request to the server, which has all the necessary artillery to compute password strength without having the `zxcvbn` on both sides of the wire.

```
on blur do
  send value to server
    get score and suggestion back
    offer feedback
```

app/views/users/registrations/edit.html.erb
```diff
@@ -10,15 +10,35 @@
...
   </div>
   <div class="govuk-form-group">
     <%= f.label :password_confirmation, class: "govuk-label" %>
-    <%= f.password_field :password_confirmation, class: "govuk-input" %>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-three-quarters">
+        <%= f.password_field :password_confirmation, class: "govuk-input" %>
+      </div>
+      <div class="govuk-grid-column-one-quarter">
+        <button id="user_password_confirmation_visbility_toggle" class="govuk-button govuk-button--secondary" type="button" onclick="toggleVisibility('user_password_confirmation_visbility_toggle', 'user_password_confirmation')">Show</button>
```
freesteph, line 39: can we add this handler via JS? i.e `EventTarget.addEventListener()`?

app/assets/javascripts/application.js.erb
```diff
@@ -11,3 +13,55 @@ window.addEventListener("DOMContentLoaded", () => {
...
+  const score = 100 * result.score / 4;
+  const scoreText = score < 50 ? 'weak' : (score < 100 ? 'medium' : 'good');
+  const rawSuggestion = result.feedback.suggestions.join(', ');
+  // zxcvbn-js sometimes fails to return a suggestion even when the password is too weak, so we need to perform some coercion.
+  const suggestion = (!rawSuggestion) ? 'Choose a password that is harder to guess' : rawSuggestion;
+  return { score, scoreText, suggestion };
+};
+
+if ($('input[name="user[password]"]').length) {
+  $('body').on('keyup', 'input[name="user[password]"]', function(e) {
```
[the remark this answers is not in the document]
robinmitra: The idea is to show the password strength *while* a user types their password, rather than typing and then clicking outside the password field just to update its strength, which is not a very intuitive experience. This pattern is particularly useful in our specific use-case, since we require quite strong passwords which would probably take a few tries to get right and we ought to make it as easy as possible.

This is also how it's been implemented on most other website (e.g. Google and Dropbox), but happy to continue the discuss if you'd like.

app/assets/javascripts/application.js.erb
robinmitra: That's a good option, but wonder how the user experience would be due to the delay. May be worth giving it a try @alan-gds ?

alan-gds:
I've intregrated Steph's branch but we've noticed there's a problem with the password confirmation not being enforced.  I'm working on a fix; do not merge this yet.

app/views/users/registrations/edit.html.erb
alan-gds: Done.

alan-gds:
The password confirmation issue turns out to be pre-existing, so doesn't need to be fixed under the scope of this task (I'll raise a new Trello card for it).  I've made a change in response to Steph's PR feedback; hopefully this will be mergeable now after a quick re-review.

app/assets/javascripts/application.js.erb
freesteph: fixed now.

PR #1061, opened 2020-06-08
gem: attempt an update to SASSC

freesteph:
This usually fails at the Concourse level but maybe having upgraded to
Rails 6 has eased the path.

Note that this removes the SASS linting tool as both
gems (`scss_lint-govuk` and `scss_lint`) are deprecated and are the
only gems dragging `sass` and `sass-rails`.

We should eventually replace it (stylelint seems to be the recommended
solution) but another PR will help keep this one small (also stylelint
might think differently to scss_lint causing a trivial-but-huge
diff).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ gem "notifications-ruby-client", "~> 5.1.2"
 gem "puma", "~> 4.3"
 gem "rails", "~> 6.0.3"
 gem "rqrcode"
-gem "sass-rails", "~> 5.1"
+gem "sassc-rails"
 gem "sentry-raven"
 gem "two_factor_authentication"
 gem "tzinfo-data"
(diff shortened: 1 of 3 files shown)


freesteph:
Already done: https://trello.com/c/BopI18Az/155-move-away-from-the-deprecated-sass-library
Robin Mitra writes:

> @robinmitra approved this pull request.
>
> I'm happy with this, but let's put a task in Trello to track the SASS lint work.

robinmitra:
> Already done: [trello.com/c/BopI18Az/155-move-away-from-the-deprecated-sass-library](https://trello.com/c/BopI18Az/155-move-away-from-the-deprecated-sass-library) Robin Mitra writes:
> […](#)
> Robin Mitra approved this pull request. I'm happy with this, but let's put a task in Trello to track the SASS lint work.

Thanks, I've added this task to the existing ticket.

PR #1063, opened 2020-06-09
Feature/async client side password validation

freesteph:
# Description
Alan and I have had a look together at this change and have agreed to get it merged in the base PR to carry on work.


freesteph:
For the record, I've just had a go on master and the bug is already
present so I'm afraid the regression isn't new.

alan-gds writes:

> @alan-gds approved this pull request.
>
> Happy for this to be merged into the branch.  As you spotted, it does cause a regression with allowing a mismatched password confirmation, but that can be fixed back on the original branch.

PR #1064, opened 2020-06-09
gem: attempt an update to SASSC

freesteph:
This usually fails at the Concourse level but maybe having upgraded to
Rails 6 has eased the path.

Note that this removes the SASS linting tool as both
gems (`scss_lint-govuk` and `scss_lint`) are deprecated and are the
only gems dragging `sass` and `sass-rails`.

We should eventually replace it (stylelint seems to be the recommended
solution) but another PR will help keep this one small (also stylelint
might think differently to scss_lint causing a trivial-but-huge
diff).


freesteph:
Re-opened PR because Concourse might be having issues with force pushes.

PR #1070, opened 2020-06-12
Fix: Enforce password confirmation validation

alan-gds:
Password confirmation was not being enforced on the password change screen.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,7 +24,8 @@ class User < ApplicationRecord
             presence: true,
             length: { within: 6..80 },
             on: :update,
-            if: :password_present?
+            if: :password_present?,
+            confirmation: true
 
   validate :strong_password, on: :update, if: :password_present?
 
--- a/spec/features/authentication/edit_password_spec.rb
+++ b/spec/features/authentication/edit_password_spec.rb
@@ -41,4 +41,25 @@ describe "Editing password", type: :feature do
 
     expect(page).to have_content "Sign out"
   end
+
+  it "stays on the same page and does not change the password, if the confirmation does not match" do
+    visit edit_user_registration_path
+
+    fill_in "Current password", with: user.password
+    fill_in "Password", with: pw
+    fill_in "Password confirmation", with: "different password"
+
+    click_on "Submit"
+
+    expect(page).to have_content "Password confirmation doesn't match Password"
+
+    sign_out
+
+    fill_in "Email", with: user.email
+    fill_in "Password", with: pw
+
+    click_on "Continue"
+
+    expect(page).not_to have_content "Sign out"
+  end
 end
--- a/spec/features/authentication/edit_password_spec.rb
+++ b/spec/features/authentication/edit_password_spec.rb
@@ -52,14 +52,5 @@ describe "Editing password", type: :feature do
     click_on "Submit"
 
     expect(page).to have_content "Password confirmation doesn't match Password"
-
-    sign_out
-
-    fill_in "Email", with: user.email
-    fill_in "Password", with: pw
-
-    click_on "Continue"
-
-    expect(page).not_to have_content "Sign out"
   end
 end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -3,6 +3,7 @@ describe User do
 
   describe "validation" do
     let(:user) { create(:user) }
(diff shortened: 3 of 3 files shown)


spec/features/authentication/edit_password_spec.rb
```diff
@@ -41,4 +41,25 @@
...
+  it "stays on the same page and does not change the password, if the confirmation does not match" do
+    visit edit_user_registration_path
+
+    fill_in "Current password", with: user.password
+    fill_in "Password", with: pw
+    fill_in "Password confirmation", with: "different password"
+
+    click_on "Submit"
+
+    expect(page).to have_content "Password confirmation doesn't match Password"
```
freesteph, line 54: do we expect this capitalisation?

spec/features/authentication/edit_password_spec.rb
```diff
@@ -41,4 +41,25 @@
...
+
+    sign_out
+
+    fill_in "Email", with: user.email
+    fill_in "Password", with: user.password
+
+    click_on "Continue"
+
+    expect(page).to have_content "Sign out"
```
freesteph, line 63: this implies the user successfully logs in?

alan-gds: Yes - they are still using the original password, not the one they tried to change to (pw).

spec/features/authentication/edit_password_spec.rb
alan-gds: I don't like it but it's what Rails gave me; let me know if you know a way to override it.

spec/features/authentication/edit_password_spec.rb
freesteph: OK I see. The logic is valid but inverted. Can you make the test hold the new password and make sure the new password does _not_ work?

alan-gds: Fair enough - now changed

freesteph:
Rebase to avoid AWS dependency issue.

spec/features/authentication/edit_password_spec.rb
```diff
@@ -41,4 +41,25 @@
...
+
+    fill_in "Current password", with: user.password
+    fill_in "Password", with: pw
+    fill_in "Password confirmation", with: "different password"
+
+    click_on "Submit"
+
+    expect(page).to have_content "Password confirmation doesn't match Password"
+
+    sign_out
```
freesteph, line 56: let's just discard this whole bit, the whole signing out then logging in then checking for the _absence_ of the Sign out button is still confusing...

app/models/user.rb
```diff
@@ -24,7 +24,8 @@ class User < ApplicationRecord
             presence: true,
             length: { within: 6..80 },
             on: :update,
-            if: :password_present?
+            if: :password_present?,
+            confirmation: true
```
freesteph, line 28: this could probably do with a model test:

```patch
From 50634d918f0715a5357eefce00800170c112eca0 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?St=C3=A9phane=20Maniaci?= <stephane.maniaci@gmail.com>
Date: Wed, 17 Jun 2020 16:28:39 +0100
Subject: [PATCH] fix: enforce password confirmation at the user model level

---
 spec/models/user_spec.rb | 14 +++++++++++++-
 1 file changed, 13 insertions(+), 1 deletion(-)

diff --git a/spec/models/user_spec.rb b/spec/models/user_spec.rb
index 2ba1aebe..2332f6fa 100644
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -3,7 +3,7 @@ describe User do
 
   describe "validation" do
     let(:user) { create(:user) }
-
+    let(:pw) { "this is 333 a strong !!! password" }
     it "has a valid factory user" do
       expect(user).to be_valid
     end
@@ -15,6 +15,18 @@ describe User do
         expect(user).not_to be_valid
       end
     end
+
+    it 'accepts a strong password' do
+      user.update(password: pw)
+
+      expect(user).to be_valid
+    end
+
+    it 'checks for a password confirmation field' do
+      user.update(password: pw, password_confirmation: pw.reverse)
+
+      expect(user).not_to be_valid
+    end
   end
 
   it { is_expected.to have_many(:organisations).through(:memberships) }
-- 
2.26.2
```

robinmitra:
🎉 ✌️

PR #1094, opened 2020-06-30
Relax the password strength requirement to a more sensible level

robinmitra:
## Description

We used to require a password score of `4`, which was very high and not really necessary. This is because we already have account locking, password hashing and 2FA in place. With these measures, it's unrealistic for someone with reasonable skills to be able to hack passwords with scores of even `2` and above easily.

The [zxcvbn's](https://github.com/dropbox/zxcvbn) docs provide a rough estimation on how easy it is to guess passwords with various scores. This PR is an attempt to lower the required score while erring on the side of caution (by opting for a score of `2` instead of `1`), because some passwords with a score of `1` were found to be quite short, and passwords with scores of `2` were found to be reasonably complex but not onerous.

---

## Scoring examples

**Easy word (score: `0`):**

GovWifi_Admin

**Common password (score `1`):**

GovWifi_Admin

**Common letter substitutions i.e. s=5, o=0, s=$ (score `1`):**

GovWifi_Admin

**Common numerical sequence (score `1`):**

GovWifi_Admin

**Random numbers (score `2`):**

GovWifi_Admin

**Random characters (score `2`):**

GovWifi_Admin

**Three random words, as recommended by NCSC (score `2`):**

GovWifi_Admin

## Indicative comparison of acceptable passwords, before and after this PR

**Before:**

GovWifi_Admin

**After:**

GovWifi_Admin

### DIFF
--- a/app/assets/javascripts/password_validation.js
+++ b/app/assets/javascripts/password_validation.js
@@ -13,7 +13,7 @@ const getSuggestionHtml = suggestions => {
 };
 
 const getScoreHtml = score => {
-  const scoreClassification = score < 2 ? 'weak' : (score < 4 ? 'medium' : 'good');
+  const scoreClassification = score < 2 ? 'weak' : (score < 3 ? 'good' : 'excellent');
   return `
     Password strength:
     <span class="password-strength-score score-${scoreClassification}">${scoreClassification}</span>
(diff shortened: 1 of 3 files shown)


ihateusernames:
Thanks Robin - looks good. (So much better than the one we had a few months back)

PR #1103, opened 2020-07-23
Feature: Save chosen 2FA method

alan-gds:
Store the chosen 2FA method in the user model, and present the chosen 2FA screen when the user logs in again.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -61,15 +61,28 @@ protected
 
   def confirm_two_factor_setup
     user = request.env["warden"].user
+
+    if Rails.configuration.enable_enhanced_2fa_experience
+      return if user.nil?
+
+      if user.second_factor_method.present?
+        # Handled by Devise's controller.
+        return if user.second_factor_method == "app"
+
+        # Handled by us.
+        redirect_to users_two_factor_authentication_email_path
+        return
+      end
+
+      # TODO: Replace with route helper once fully migrated to enhanced 2FA.
+      redirect_to "/users/two_factor_authentication/setup"
+      return
+    end
+
     return unless user &&
       user.need_two_factor_authentication?(request) &&
       !user.totp_enabled?
 
-    if Rails.configuration.enable_enhanced_2fa_experience
-      # TODO: Replace with route helper once fully migrated to enhanced 2FA.
-      redirect_to "/users/two_factor_authentication/setup"
-    else
-      redirect_to users_two_factor_authentication_setup_path
-    end
+    redirect_to users_two_factor_authentication_setup_path
   end
 end
--- a/app/controllers/users/two_factor_authentication/setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication/setup_controller.rb
@@ -51,7 +51,7 @@ class Users::TwoFactorAuthentication::SetupController < ApplicationController
 private
 
   def save_two_factor_method(method)
-    # TODO: This will be implemented in future.
+    current_user.update(second_factor_method: method)
   end
 
   def send_email
--- a/spec/features/registration/sign_up_as_an_organisation_with_enhanced_2fa_spec.rb
+++ b/spec/features/registration/sign_up_as_an_organisation_with_enhanced_2fa_spec.rb
@@ -1,0 +1,243 @@
+require "support/notifications_service"
+require "support/confirmation_use_case"
+
+describe "Sign up as an organisation", type: :feature do
+  let(:name) { "Sally" }
+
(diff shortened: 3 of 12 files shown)


app/controllers/application_controller.rb
```diff
@@ -61,15 +61,28 @@ def valid_help_actions
 
   def confirm_two_factor_setup
     user = request.env["warden"].user
-    return unless user &&
-      user.need_two_factor_authentication?(request) &&
-      !user.totp_enabled?
 
     if Rails.configuration.enable_enhanced_2fa_experience
+      return unless user
```
freesteph, line 66: `return unless user.present?` is easier to read

app/controllers/application_controller.rb
```diff
@@ -61,15 +61,28 @@ def valid_help_actions
...
+
+      unless user.second_factor_method.nil?
```
freesteph, line 68: I would also favour `if user.second_factor_method.present?` as I find the positive case easier to read.

app/controllers/application_controller.rb
```diff
@@ -61,15 +61,28 @@ def valid_help_actions
...
       # TODO: Replace with route helper once fully migrated to enhanced 2FA.
       redirect_to "/users/two_factor_authentication/setup"
-    else
-      redirect_to users_two_factor_authentication_setup_path
+      return
     end
+
+    return unless user &&
+      user.need_two_factor_authentication?(request) &&
+      !user.totp_enabled?
```
freesteph, line 84: all this logic should live in the user model (`user.rb`) in a method that indicates what this is checking for (user exists, and needs 2FA but doesn't have app 2FA?)

app/controllers/users/two_factor_authentication/setup_controller.rb
```diff
@@ -51,7 +51,8 @@ def qr_code_uri
 private
 
   def save_two_factor_method(method)
-    # TODO: This will be implemented in future.
+    current_user.second_factor_method = method
+    current_user.save
```
freesteph, line 55: use `current_user.update_attributes(second_factor_method: method)` to save yourself the save (sic).

spec/features/registration/sign_up_as_an_organisation_with_enhanced_2fa_spec.rb
```diff
@@ -0,0 +1,243 @@
...
+      let(:email) { "" }
+
+      it "tells me my email is not valid" do
+        expect(page).to have_content("Email address must be from a government or a public sector domain. If you're having trouble signing up, contact us.")
+      end
+    end
+
+    context "without a name" do
+      let(:name) { "" }
+      let(:email) { "someone@other.gov.uk" }
```
freesteph, line 86: all of this looks like unit tests for the User model to me. Given how expensive feature tests are, we definitely don't want to enforce all of our validation at the system level.

app/controllers/application_controller.rb
koetsier: or ``` if user.nil?```

app/controllers/application_controller.rb
```diff
@@ -61,15 +61,28 @@ def valid_help_actions
...
 
     if Rails.configuration.enable_enhanced_2fa_experience
+      return unless user
+
+      unless user.second_factor_method.nil?
+        # Handled by Devise's controller.
+        return if user.second_factor_method == "app"
+
+        # Handled by us.
+        redirect_to users_two_factor_authentication_email_path
```
koetsier, line 73: I'm not sure about this redirect here.  I believe this bit of code is supposed to check if the user has chosen email/app authentication. I did a bit of refactoring here which makes this a bit easier to read I think: https://github.com/alphagov/govwifi-admin/pull/1104/commits/7ed57ea221e7e41e5bbae3370a87164892bb2b3a

robinmitra: Thanks Jos, my preference would be to land this as a separate PR. Let me know if you have any concerns with this.

app/controllers/application_controller.rb
robinmitra: @koetsier was looking at refactoring this bit anyway, so happy to take your comment onboard again on that PR.

spec/features/registration/sign_up_as_an_organisation_with_enhanced_2fa_spec.rb
robinmitra: That's a good point, but this was taken from an existing test in order to make it work specifically with the new 2FA journeys. Happy to consider this outside this PR.

PR #1107, opened 2020-08-03
Fixed linting issues.

koetsier:
Fixing these issues allows us to upgrade to the latest GOV.UK Rubocop version

- Added two migrations to add indices to models with uniqueness constraints
- Added a ```dependent: :destroy``` option to the user model to make sure that memberships get destroyed when a user is deleted.
- The indices to the name column of the authorised_email_domain table means that the order of ips returned in the test in authorised_email_domains_spec file is reversed. The order does not seem to matter so that is now reflected in the test

### DIFF
--- a/spec/models/organisation_spec.rb
+++ b/spec/models/organisation_spec.rb
@@ -213,7 +213,13 @@ describe Organisation do
     let(:ip4) { create(:ip, location: location2) }
     let(:ip5) { create(:ip, location: location3) }
 
-    before { ip1; ip2; ip3; ip4; ip5 }
+    before do
+      ip1
+      ip2
+      ip3
+      ip4
+      ip5
+    end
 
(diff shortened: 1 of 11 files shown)


spec/models/organisation_spec.rb
```diff
@@ -213,7 +213,13 @@
     let(:ip4) { create(:ip, location: location2) }
     let(:ip5) { create(:ip, location: location3) }
 
-    before { ip1; ip2; ip3; ip4; ip5 }
+    before do
+      ip1
+      ip2
+      ip3
+      ip4
+      ip5
```
freesteph, line 221: You can save yourself the trouble by immediately invoking the `let` blocks with the special [`let!` form](https://relishapp.com/rspec/rspec-core/v/3-9/docs/helper-methods/let-and-let).

PR #1108, opened 2020-08-03
Send emails when selecting the email 2FA method

koetsier:
This PR also contains various refactorings so that all authentication now goes through the Devise 2fa gem as much as possible.

Trello: https://trello.com/c/1xXXw0Qu/164-implement-sending-email-with-totp-link-that-can-be-clicked-to-log-in


app/controllers/application_controller.rb
```diff
@@ -59,30 +61,26 @@ def valid_help_actions
...
+  def tfa_feature_flag?
+    Rails.configuration.enable_enhanced_2fa_experience
+  end
 
-        # Handled by us.
-        redirect_to users_two_factor_authentication_email_path
-        return
-      end
+  def choose_two_factor_method
+    return if user.nil? || !user.second_factor_method.nil?
```
freesteph, line 69: for all these `check X first then check X.Y` scenarios I think it might be easier to use the safe operator (`user&.second_factor_method`) or any of the [techniques described here](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/). I quite like `.try` as well which is ActiveRecord specific but is nice and clear.

freesteph: Rubocop might concur with `return unless user.try(:second_factor_method).present?` (the `.present` isn't necessary but it makes it easier to read)

app/controllers/application_controller.rb
```diff
@@ -59,30 +61,26 @@ def valid_help_actions
...
 
+  def confirm_two_factor_setup
     return unless user &&
       user.need_two_factor_authentication?(request) &&
       !user.totp_enabled?
 
     redirect_to users_two_factor_authentication_setup_path
   end
+
+  def user
```
freesteph, line 83: shouldn't this be `current_user` throughout?

freesteph: which is provided out of the box by Devise

app/controllers/users/two_factor_authentication_controller.rb
```diff
@@ -1,5 +1,32 @@
...
+  end
+
+  def show
+    @resent = !!params[:resent]
+  end
+
+  def handle_missing_2fa_method
+    return if current_user.second_factor_method.present?
+
+    redirect_to("/") && return if params[:method].nil?
```
freesteph, line 19: `root_path` might be a simpler choice

config/locales/devise.en.yml
```diff
@@ -2,6 +2,12 @@
 
 en:
   devise:
+    two_factor_authentication:
+      success: "Two factor authentication setup successful."
+      attempt_failed: "Six digit code is not valid"
+      max_login_attempts_reached: "Access completely denied as you have reached your attempts limit"
```
freesteph, line 8: I feel like we should get this copy reviewed before we tell users their access has been **COMPLETELY DENIED** :smile:

config/locales/devise.en.yml
```diff
@@ -2,6 +2,12 @@
...
+      contact_administrator: "Please contact your system administrator."
```
freesteph, line 9: This doesn't seem to be used elsewhere

spec/features/authentication/logging_in_after_signing_up_with_enhanced_2fa_spec.rb
```diff
@@ -23,20 +27,34 @@
...
       expect(page).to have_content "Invalid Email or password"
     end
   end
+
+  context "email 2fa has been setup" do
+    before :each do
+      @user.update(second_factor_method: "email")
+      @user.create_direct_otp
+    end
+    it "sends an email" do
```
freesteph, line 55: newline

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
```diff
@@ -62,17 +94,55 @@
...
+        expect(page).to have_content("If you do not receive the email, we can send you a new one.")
+      end
+      it "does not send further emails" do
+        expect(notification_instance).to have_received(:send_email).once
+      end
+      context "the user confirms" do
+        before :each do
+          click_on "Resend email"
+        end
+        it "sends another email" do
```
freesteph, line 116: let's space all the `it` block in newlines

app/controllers/application_controller.rb
koetsier: This one is not quite the same. ```User.nil?``` signifies the case that the user hasn't signed in - the second bit of the expression is a logged in user not having a ```second_factor_method``` set. Say the user hasn't logged in then the second bit is making this expression always true if I use the safe operator. I think the ```unless``` case in your second comment is harder to read

config/locales/devise.en.yml
koetsier: This isn't actually used at the moment I think.

app/controllers/application_controller.rb
freesteph: > Say the user hasn't logged in then the second bit is making this expression always true if I use the safe operator. 

Is always false. `nil.try(:anything)` -> `nil`.

Happy to stick to your `if` version.

app/controllers/application_controller.rb
koetsier: Probably. Copy-pasted from what was there. I'll change it

app/controllers/users/two_factor_authentication_controller.rb
koetsier: 👍

config/locales/devise.en.yml
koetsier: I've remove the unused lines

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
koetsier: 👍

spec/features/authentication/logging_in_after_signing_up_with_enhanced_2fa_spec.rb
koetsier: 👍

config/routes.rb
```diff
@@ -23,11 +25,15 @@
     # Newer implementation: will supersede the above.
     get "users/two_factor_authentication/setup", to: "users/two_factor_authentication/setup#show"
     put "users/two_factor_authentication/setup", to: "users/two_factor_authentication/setup#update"
-    get "users/two_factor_authentication/email", to: "users/two_factor_authentication/email#show"
-    post "users/two_factor_authentication/email", to: "users/two_factor_authentication/email#update"
+
+    get "users/two_factor_authentication/email/resend", to: "users/two_factor_authentication/email#resend"
```
robinmitra, line 29: I would have preferred to keep the URLs restful as much as possible, and annotate actions as parameters rather than entities on their own. So, since `resend` is an action on the `email` entity, my preference would have been to specify it as a parameter on the `email` resource (e.g. `/email?action=resend` or `/email?resend=true`, etc.).

I feel the same about the other routes below as well, with paths ending with an action or which are there just for convenience rather than leading up to an entity.

config/gov_notify.yml
```diff
@@ -30,6 +34,8 @@ development:
 
 test:
   support_email: 'test@localhost'
+  otp_email:
+    template_id: '776fff48-3e94-4417-83b0-bff027c16247'
```
robinmitra, line 38: Should be create one for production too? Might be good to set it up now rather than a separate PR later but let me know if you feel otherwise.

config/routes.rb
koetsier: Maybe I can rename the routes to 'email/confirmation' and 'email/resend_form' or something so they are nouns? That would make it Restful again. I'd rather not add routing complexity in the controller. What do you think?

config/gov_notify.yml
koetsier: ok!

config/routes.rb
```diff
@@ -23,11 +25,15 @@
...
+    get "users/two_factor_authentication/email/confirm", to: "users/two_factor_authentication/email#confirm"
```
robinmitra, line 30: IMHO this path should ideally be nested within `/setup` as it's part of the setup process (e.g. `/setup/email?action=confirm`), and would have preferred to have the action (i.e. `confirm`) not specified as an entity on the path.

config/routes.rb
```diff
@@ -23,11 +25,15 @@
...
+    get "users/two_factor_authentication/app/totp_with_qr", to: "users/two_factor_authentication/app#totp_with_qr"
```
robinmitra, line 31: Following the same reasoning as above, this path should ideally be nested within `/setup` as it's part of the setup process (e.g. `/setup/app`). 

Why is `totp_with_qr` needed? Can we not do without it? I'm not entirely clear.

config/routes.rb
```diff
@@ -23,11 +25,15 @@
...
+    put "users/two_factor_authentication/email", to: "users/two_factor_authentication/email#update"
 
     get "users/:id/two_factor_authentication/edit", to: "users/two_factor_authentication#edit"
     delete "users/:id/two_factor_authentication", to: "users/two_factor_authentication#destroy"
+    get "users/two_factor_authentication/auth/:code", to: "users/two_factor_authentication#update", as: "users_two_factor_authentication_direct_otp"
```
robinmitra, line 36: Since this is for email OTP, IMHO the path would be clearer if it had the email resource in it (e.g. `..two_factor_authentication/email`), in order to avoid any confusion and since we already have the `..two_factor_authentication/email` resource.

Also, my preference would be to have the code provided as a query string, rather than being a separate entity (e.g. `two_factor_authentication/email?action=auth&code=1234`).

robinmitra:
Looks good overall. I noticed a strange bug when I tried to login with `loneadmin@gov.uk` user - Chrome gave me the error that I got redirected too many times. See if you can reproduce this.

config/routes.rb
robinmitra: Actually, having looked at the behaviour on the browser, I'm happy with `..two_factor_authentication/email/resend` – apologies for the confusion.

config/routes.rb
robinmitra: I'm happy with your suggestion to replace `confirm` with `confirmation`. But would it be possible to have it nested within `setup`?

config/routes.rb
koetsier: totp_with_qr is needed because it displays the QR of the secret key, which should only be shown at setup.

koetsier:
> Looks good overall. I noticed a strange bug when I tried to login with `loneadmin@gov.uk` user – Chrome gave me the error that I got redirected too many times. See if you can reproduce this.

Good catch! This is a redirect that happens when a user does not have an organisation. This causes it to redirect to a help page, which gets redirected to the 2fa page and back again after. I'll add a commit to fix this

config/routes.rb
koetsier: renamed it to the default show

config/routes.rb
koetsier: I moved both to setup as requested

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
```diff
@@ -1,161 +1,311 @@
...
-  end
+  include_context "when using the notifications service"
 
-  it "explains the setup step" do
-    expect(page).to have_content("Set up two-factor authentication")
+  before :each do
+    allow_any_instance_of(ROTP::TOTP).to receive(:verify).with(correct_totp_code, anything).and_return(true)
+    allow_any_instance_of(ROTP::TOTP).to receive(:verify).with(incorrect_totp_code, anything).and_return(false)
+    allow(Rails.application.config).to receive(:enable_enhanced_2fa_experience).and_return true
   end
```
robinmitra, line 14: Newline please

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
```diff
@@ -1,161 +1,311 @@
...
+          click_on "Sign out"
+          visit root_path
+
+          fill_in "Email", with: "tom@gov.uk"
+          fill_in "Password", with: correct_password
+          click_on "Continue"
+        end
+
+        it "sends an email again" do
+          expect(notification_instance).to have_received(:send_email).twice
```
robinmitra, line 143: I'm probably missing something obvious, but why is email being sent twice in this case?

spec/models/user_spec.rb
```diff
@@ -224,4 +228,38 @@
       expect(user).not_to be_totp_enabled
     end
   end
+
+  describe "#send_new_otp_after_login?" do
+    it "does not send an email if the direct otp has not been set" do
+      user = create(:user, :with_organisation, second_factor_method: "email")
+      expect(user.send_new_otp_after_login?).to be false
+    end
```
robinmitra, line 236: Missing line break.

spec/models/user_spec.rb
```diff
@@ -224,4 +228,38 @@
...
+    it "it sends an email if the direct otp has been set" do
+      user = create(:user, :with_organisation, second_factor_method: "email")
+      user.create_direct_otp
+      expect(user.send_new_otp_after_login?).to be true
+    end
+    it "does not send an email if the user has chosen to 2fa through the app" do
```
robinmitra, line 242: Same.

spec/models/user_spec.rb
```diff
@@ -224,4 +228,38 @@
...
+    end
+  end
+
+  describe "#send_new_otp" do
+    let(:user) { create(:user, :with_organisation) }
+    context "the user has a direct otp set" do
+      before :each do
+        user.create_direct_otp
+        user.send_new_otp
+      end
```
robinmitra, line 254: Same.

spec/models/user_spec.rb
```diff
@@ -224,4 +228,38 @@
...
+      expect(user.send_new_otp_after_login?).to be true
+    end
+    it "does not send an email if the user has chosen to 2fa through the app" do
+      user = create(:user, :with_organisation, second_factor_method: "app", otp_secret_key: "12345678")
+      expect(user.send_new_otp_after_login?).to be false
+    end
+  end
+
+  describe "#send_new_otp" do
+    let(:user) { create(:user, :with_organisation) }
```
robinmitra, line 249: Line break would be good here.

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
```diff
@@ -1,161 +1,311 @@
...
       click_on "Continue"
 
-      fill_in :code, with: "123456"
-      click_on "Complete setup"
-    end
-
-    it "returns an error" do
-      expect(page).to have_content("Six digit code is not valid")
+      visit root_path
     end
```
robinmitra, line 294: Line break.

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
```diff
@@ -1,161 +1,311 @@
...
-
-    it "returns to the 2FA screen" do
-      expect(page).to have_content("Scan the QR code")
+    it "presents the setup page" do
+      expect(page).to have_current_path("/users/two_factor_authentication/setup")
     end
```
robinmitra, line 297: Same.

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
```diff
@@ -1,161 +1,311 @@
...
-    it "doesn't store a totp for the user" do
-      expect(user.otp_secret_key).to be nil
+    context "continues with the setup process" do
+      before :each do
+        choose "app"
+        click_on "Continue"
+
+        fill_in :code, with: correct_totp_code
+        click_on "Complete setup"
+      end
```
robinmitra, line 305: Same.

robinmitra:
@koetsier Changes look good apart from minor cosmetic things 👍🏼

spec/features/authentication/set_up_enhanced_two_factor_authentication_spec.rb
koetsier: The email has already been sent in the previous ```before``` blocks and now the user clicks 'resend' which means it should be sent again.

koetsier:
> @koetsier Changes look good apart from minor cosmetic things 👍🏼

Added all the newlines where necessary

app/views/users/two_factor_authentication/show.html.erb
```diff
@@ -1,22 +1,10 @@
+<% method = tfa_feature_flag? ? current_user&.second_factor_method : "app" %>
```
freesteph, line 1: Can this logic be moved back into the controller rather than in the view?

koetsier: 👍

PR #1132, opened 2020-09-02
BAU: Update govuk-rubocop to 3.6.0

alan-gds:
I've just noticed that we can now get rubocop-govuk to a higher version than we were able to back when I looked at linting over firebreak; this is probably a result of the linting fixes @koetsier committed recently.  This PR bumps it up to 3.6.0, requiring only one small fix in a .erb file along the way.  3.7.0 is blocked by an issue in another .erb, which I'm not sure how best to fix.  Let's get this update in, and then we can try and sort that out.


alan-gds:
Damn, the build is failing, probably because of my change to the .erb file.

alan-gds:
The .erb issue is now fixed and the PR is ready for review.

Gemfile
```diff
@@ -37,7 +37,7 @@ group :test do
   gem "rack_session_access"
   gem "rails-controller-testing"
   gem "rspec-rails", "~> 4"
-  gem "rubocop-govuk"
+  gem "rubocop-govuk", "=3.6.0"
```
freesteph, line 40: can  we use `~> 3.6` here?

app/views/layouts/_flash_notices.html.erb
```diff
@@ -4,7 +4,7 @@
 
   <% if name == 'notice' %>
     <div class='flash-holder'>
-      <%= content_tag :p, msg, class: "govuk-heading-m flash-message-notice" %>
+      <%= tag.p(msg, class: "govuk-heading-m flash-message-notice") %>
```
robinmitra, line 7: @alan-gds Can you provide more context please as to why `tag.p` is preferred here?

Gemfile
alan-gds: Eek, this was a mistake.  Well spotted.

app/views/layouts/_flash_notices.html.erb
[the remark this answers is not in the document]
alan-gds: Against 3.6.0?  Is the linter definitely getting run against the .erb files on your machine?

alan-gds: Apparently content_tag is legacy syntax.
https://github.com/rubocop-hq/rubocop-rails/pull/242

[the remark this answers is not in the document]
freesteph: Yep, no offence here.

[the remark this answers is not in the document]
freesteph: My apologies, it is an offence, I ran `rubocop` instead of `erb-lint`.

alan-gds:
I've fixed the problem Steph spotted, where rubocop-govuk was being pinned to a specific version.  It shouldn't be pinned at all; I committed that in error.

Hopefully this is all good to go now!

Gemfile
freesteph: Do we just enforce in the lock file then?

freesteph: just saw the comment, all good :+1:

freesteph (approved):

PR #1137, opened 2020-09-14
Fix/ally sept 2020

freesteph:
# Description
Attempt to address all the manually reported issues from our Admin portal.

# Review
- Read commit log;
- Try branch locally.

### DIFF
--- a/app/assets/stylesheets/components/overview.scss
+++ b/app/assets/stylesheets/components/overview.scss
@@ -14,6 +14,10 @@
   color: $colour-white;
 }
 
+.summary-link-colour:focus {
+  color: $colour-black;
+}
+
 .overview-section-alignment {
   padding-left: 0;
   padding-right: 0;
--- a/app/assets/stylesheets/components/overview.scss
+++ b/app/assets/stylesheets/components/overview.scss
@@ -20,6 +20,8 @@
 }
 
 .overview-stats-panel {
+  @include govuk-responsive-margin(6, "bottom");
+
   padding-left: 20px;
   padding-top: 28px;
   height: 140px;
--- a/app/views/overview/index.html.erb
+++ b/app/views/overview/index.html.erb
@@ -1,13 +1,15 @@
 <% unless @current_org_signed_mou %>
-<div class="govuk-grid-row">
-  <div class="govuk-warning-text" id="mou-warning">
-    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
-    <strong class="govuk-warning-text__text">
-      <span class="govuk-warning-text__assistive">Warning</span>
-      You must sign the GovWifi <%= link_to "memorandum of understanding", mou_index_path, class: "govuk-link govuk-link--no-visited-state" %> to use this service.
-    </strong>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-warning-text" id="mou-warning">
+        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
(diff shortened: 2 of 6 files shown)


app/views/overview/index.html.erb
```diff
@@ -1,13 +1,15 @@
 <% unless @current_org_signed_mou %>
-<div class="govuk-grid-row">
-  <div class="govuk-warning-text" id="mou-warning">
-    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
-    <strong class="govuk-warning-text__text">
-      <span class="govuk-warning-text__assistive">Warning</span>
-      You must sign the GovWifi <%= link_to "memorandum of understanding", mou_index_path, class: "govuk-link govuk-link--no-visited-state" %> to use this service.
-    </strong>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
```
robinmitra, line 3: @freesteph Which guideline does this relate to?

freesteph: I'm not sure, I just fixed it as a I was tidying up the page and it looked bad on mobile.

app/views/overview/index.html.erb
```diff
@@ -1,43 +1,45 @@
...
```
robinmitra, line 3:
```suggestion
    <div class="govuk-grid-column-full">
```
You're right. But on desktop, it unnecessary wraps (and I remember us fixing this a while back). The suggestion will fix this.

app/assets/stylesheets/components/overview.scss
```diff
@@ -1,17 +1,5 @@
...
-  color: $colour-white;
-}
-
-.summary-link-colour:visited {
-  color: $colour-white;
-}
-
-.summary-link-colour:hover {
-  color: $colour-white;
+  border: 3px solid $colour-blue;
```
robinmitra, line 2: This seems like a significant change visually. Do you think we can make it work like this?
Kapture 2020-09-17 at 9 55 46

app/views/overview/index.html.erb
freesteph: > For most types of page, we recommend using either a ‘two-thirds’ or a ‘two-thirds / one third layout’. That stops lines of text getting so long that the page becomes difficult to read on desktop devices.

https://design-system.service.gov.uk/styles/layout/

app/assets/stylesheets/components/overview.scss
freesteph: think you might be caching old CSS, the new version is this: 

file:///home/steph/Pictures/Screenshot%20from%202020-09-17%2010-25-31.png

freesteph: Screenshot from 2020-09-17 10-25-31

robinmitra: Yes that's what I meant when I said the change looks significantly different visually. The gif I attached keeps the look how it used to be while still keeping the contrast acceptable.

Having said that, I just went back to the audit doc and the issue was actually having white text over yellow background, when links in those boxes are focused. So the only change we need to make is to change the link text to black when focused. I've checked the contrast against the blue background and it seems acceptable for WCAG 2.1A.

app/views/overview/index.html.erb
robinmitra: It's already in a three-quarters column, and restricting it further to two-thirds effectively makes it one-half, hence I found it looking a bit odd. This is non-blocking so feel free to disregard if you disagree.

robinmitra (approved):
🎉

PR #1156, opened 2020-10-05
fix: add missing page titles

freesteph:
Beth and I have gone through most of the pages exposed in the GovWifi
admin and have amended the page titles to reflect their content.

Some pages were intentionally left out because of their complexity,
namely the superadmin "allow orgs/users" flow, and the help form which
can be figured out in the coming days.

### DIFF
--- a/app/views/mou/index.html.erb
+++ b/app/views/mou/index.html.erb
@@ -1,3 +1,5 @@
+<% content_for :page_title, "Memorandum of understanding - GovWifi" %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-three-quarters">
     <%= link_to "Back to settings", setup_instructions_path, class: "govuk-back-link" %>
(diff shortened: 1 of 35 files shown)


app/views/mou/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :page_title, "Memorum of understanding - GovWifi" %>
```
robinmitra, line 1:
```suggestion
<% content_for :page_title, "Memorandum of understanding - GovWifi" %>
```

app/views/super_admin/whitelists/email_domains/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :page_title, "Manage access - GovWifi" %>
```
robinmitra, line 1: This doesn't quite sound right to me. Something like "Manage allowed domains - GovWifi" is more accurate.

robinmitra (approved):
🚀

PR #1160, opened 2020-10-12
fix: don't enforce Bundler 1

freesteph:
Bundler 2 has been out for a while so we should really stop enforcing
1.17 here.


alan-gds (approved):
👍

PR #1168, opened 2020-10-21
Fix null second factor attempts count bug

alan-gds:
The database column now defaults to 0 instead of null.

https://trello.com/c/sEnMZaaW/451-second-factor-attempt-is-null-by-default-on-admin-database

### DIFF
--- a/db/migrate/20201020154112_change_default_value_of_second_factor_attempt_count.rb
+++ b/db/migrate/20201020154112_change_default_value_of_second_factor_attempt_count.rb
@@ -1,0 +1,27 @@
+class ChangeDefaultValueOfSecondFactorAttemptCount < ActiveRecord::Migration[6.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        change_column_default(:users, :second_factor_attempts_count, from: nil, to: 0)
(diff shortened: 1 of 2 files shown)


alan-gds:
Added a second commit to fix some issues spotted by Steph.  @freesteph, please have another go at reviewing.

db/migrate/20201020154112_change_default_value_of_second_factor_attempt_count.rb
```diff
@@ -0,0 +1,15 @@
+class ChangeDefaultValueOfSecondFactorAttemptCount < ActiveRecord::Migration[6.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        change_column_default(:users, :second_factor_attempts_count, from: nil, to: 0)
+        User.update_all(second_factor_attempts_count: 0)
+      end
+
+      dir.down do
+        User.update_all(second_factor_attempts_count: nil)
```
freesteph, line 10: I'd remove this line. Restoring a database default I can get behind but voluntarily restoring buggy data, not so much.

freesteph (approved):
Just one last edit and we're good to go

PR #1177, opened 2020-10-30
Fix/remaining page titles

freesteph:


app/views/users/two_factor_authentication/max_login_attempts_reached.html.erb
```diff
@@ -0,0 +1,12 @@
+<% content_for :page_title, "You've reached the maximum amount of sign in attempts" %>
```
alan-gds, line 1: This new page looks fine, but it might be worth mentioning in the commit message that it's been added.

alan-gds (commented):
What's been done all looks good, but I'm not seeing page titles on app-based 2FA setup stuff (the pages where you enter the 2FA code once it's already been set up are fine).  Does that need to be changed?

app/views/users/two_factor_authentication/max_login_attempts_reached.html.erb
freesteph: Beth suggested we don't focus on that new journey because it's undergoing substantial changes, hence why it's been left out. You can turn off the new journey by going in `config/development.rb` and editing for this: 

```
  config.enable_enhanced_2fa_experience = false
```

Note that production is not running the new journey either.

alan-gds (approved):

PR #1178, opened 2020-11-05
Add `has_2fa?` method to `user`, and use in memberships table

colinbm:
We've previously used `totp_enabled?` to decide if we should display
2fa reset link, however this only checks app 2fa. This method checks both
app and email.

Note, the `reset_2fa!` method on `user` already resets the
`second_factor_method` column, which covers email, so no changes are
required there.


freesteph (approved):
awesome :+1:

PR #1188, opened 2020-11-12
Bau/add secrets baseline

alan-gds:
https://github.com/alphagov/gds-pre-commit

The majority of detected secrets are clearly false passwords used in tests or in developer config.  There are a handful that might conceivably be real ones:

* `Dockerfile` contains a NOTIFY_API_KEY which looks to me like it could be genuine.  Is this a problem?
* It also has a few DB_PASS keys with a value of `root`; I presume those aren't for anything that affects a live environment.
* `config/initializers/devise.rb` had a commented out config item called `config.pepper` with a lengthy hex value.  This had been unchanged (i.e. always commented out) since the file was created.  I'm assuming it was just there as a template.  To avoid confusion and reduce the noise in .secrets.baseline, I've replaced the hex value with a note.  I'm assuming the original value was not actually relevant to our system - but if it was, we will need to do something about the git history.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-11-12T16:28:50Z",
+  "generated_at": "2020-11-12T16:47:39Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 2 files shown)


freesteph (approved):
Happy with it.

PR #1203, opened 2020-12-02
BAU: Upgrade httparty from 0.18.0 to 0.18.1

alan-gds:
Dependabot PR #1076 has been sitting around for a while because it breaks the build.  I just took a closer look and the failures are because the new version of httparty encodes keys in query parameters; there are some places where we had literal [] in key names, and the brackets now get encoded.  This caused a few tests to fail because they assert that the body of a request will contain a specific URL including the literal [].

I'm assuming that the right solution here is to update the test expectations, but I don't have much understanding of how this stuff is being used, so please be sceptical about this assumption.

More detail about the relevant change in httparty can be found here: https://github.com/jnunemaker/httparty/pull/698


freesteph (approved):

freesteph:
I think it's fine; let's deploy to staging and see what happens.

PR #1230, opened 2021-02-01
Remove enhanced 2fa

koetsier:
Enhanced 2FA was supposed to enable email otp based 2fa as well
as app-based TOTP 2fa. It was decided that email based 2fa was not
secure enough as users can reset their password using email as well.

This PR removes the email based 2FA option

Trello: https://trello.com/c/9QhzrTg5/543-tech-work-to-roll-back-2fa-email


app/models/user.rb
```diff
@@ -81,13 +79,8 @@ def super_admin?
   end
 
   def need_two_factor_authentication?(request)
```
colinbm, line 81: This is flagged by the linter, however it relates to comment coming in 3... 2... 1...

app/controllers/application_controller.rb
```diff
@@ -61,17 +60,6 @@ def valid_help_actions
...
-  def choose_two_factor_method
-    return if current_user.nil? || !current_user.second_factor_method.nil?
-
-    # TODO: Replace with route helper once fully migrated to enhanced 2FA.
-    redirect_to "/users/two_factor_authentication/setup"
-  end
-
   def confirm_two_factor_setup
     return unless current_user &&
       current_user.need_two_factor_authentication?(request) &&
```
colinbm, line 65: This will no longer need request passed once the linter issue is fixed.

colinbm (changes requested):

app/models/user.rb
```diff
@@ -98,25 +91,10 @@ def reset_2fa!
       encrypted_otp_secret_key_salt: nil,
       totp_timestamp: nil,
       otp_secret_key: nil,
-      second_factor_method: nil,
     )
   end
 
   def send_new_otp_after_login?
```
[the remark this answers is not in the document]
koetsier: `send_new_otp_after_login` is called from the 2fa Gem and is used to send an email to the user in case of email based 2fa. Since we don't do that it has to be set to false.

https://github.com/Houdini/two_factor_authentication/blob/7e0faa6e1f3ab497767c59e1e97fb01e340b59ce/lib/two_factor_authentication/hooks/two_factor_authenticatable.rb#L10

app/models/user.rb
koetsier: fixed that

app/controllers/application_controller.rb
koetsier: It does need something passed in because it overloads https://github.com/Houdini/two_factor_authentication/blob/7e0faa6e1f3ab497767c59e1e97fb01e340b59ce/lib/two_factor_authentication/models/two_factor_authenticatable.rb#L59

colinbm (approved):

colinbm (approved):

PR #1239, opened 2021-02-19
Rescue Zendesk exceptions when submitting tickets

koetsier:
This can occur when an email address passes the local form validation
 but is rejected by Zendesk

Trello: https://trello.com/c/dpv3GGFg/526-handle-malformed-emails-in-our-support-form

### DIFF
--- a/spec/features/contact_us_when_not_signed_in_spec.rb
+++ b/spec/features/contact_us_when_not_signed_in_spec.rb
@@ -160,6 +160,18 @@ describe "Contact us when not signed in", type: :feature do
         expect(page).to have_content "Email is not a valid email address"
       end
     end
+
+    context "correct but rejected by Zendesk" do
+      let(:email) { "blah_blah%\#@\#$" }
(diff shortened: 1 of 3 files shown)


spec/features/contact_us_when_not_signed_in_spec.rb
```diff
@@ -119,6 +119,8 @@
   end
 
   context "with incorrect email formats" do
+    let(:zendesk_error) { ZendeskAPI::Error::RecordInvalid }
```
colinbm, line 122: Is this doing anything? Below the class is referenced directly.

colinbm (changes requested):

spec/features/contact_us_when_not_signed_in_spec.rb
koetsier: uh good point. I'll remove

colinbm (approved):

PR #1240, opened 2021-02-22
Update dependencies.

colinbm:
The goal is to update Rails to fix the mime type bug, but all dependencies have bee updated to the latest allowed by Gemfile with a `bundle update`.

Fixes new linter errors created by upgrading `rubocop-govuk`


colinbm:
This is a bit of a mess. The linter updates break the tests (for good reason, so going to reassess this.

PR #1241, opened 2021-02-23
Upgrade to Rails 6.1.3 and update other dependencies.

colinbm:
I'd initially thought that a minor Rails bump would bring in the change needed to resolve the `Mime::Type::InvalidMimeType` Sentry alert. In fact, it needed an upgrade to 6.1, so that has been done.

A notable change required due to the Rails update is renaming `ActiveStorage::BlobsController` to `ActiveStorage::SecureBlobsController` and referencing it explicitly in the routes. This is due to it otherwise being ignored after the upgrade.

Other changes include:
* General dependency update within bounds of `Gemfile`
* Resolve new linter errors caused by `govuk-rubocop` update.
* Fix MOU forms that were set to use UJS, as that has been disabled.


Ian-Nicholls (approved):
All look good - one tiny change to min version number in Gemfile for clarity done after consult with Colin

PR #1244, opened 2021-02-26
Switch to active_record_store for sessions

colinbm:
There is a 4kb limit on cookie sessions, which occasionally gets passed causing CookieOverflow exception.
Switching to database storage solves this.

```
ActionDispatch::Cookies::CookieOverflow: ActionDispatch::Cookies::CookieOverflow
  actionpack (6.0.3.4) lib/action_dispatch/middleware/cookies.rb:637:in `commit'
  actionpack (6.0.3.4) lib/action_dispatch/middleware/cookies.rb:472:in `[]='
  actionpack (6.0.3.4) lib/action_dispatch/middleware/session/cookie_store.rb:110:in `set_cookie'
  rack (2.2.3) lib/rack/session/abstract/id.rb:403:in `commit_session'
  rack (2.2.3) lib/rack/session/abstract/id.rb:268:in `context'
...
(32 additional frame(s) were not displayed)
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
+gem "activerecord-session_store"
 gem "aws-sdk-route53", "~> 1.44.0"
 gem "aws-sdk-s3", "~> 1.86.0"
 gem "cancancan"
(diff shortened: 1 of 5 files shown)


koetsier:
Any idea why the cookie overflow is triggered? Ideally we shouldn't use the session to store a whole lot of stuff

koetsier (changes requested):
Can we find out the root cause of why the cookie overflow happens?

colinbm:
Cause unclear. We talked at standup yesterday and the consensus was to get this fix in and add a card for further investigation to the board. We can more easily monitor typical session size too once they're in the database.

koetsier (approved):
lgtm 🎉

PR #1245, opened 2021-02-26
Don't attempt to `destroy!` a `nil` record

colinbm:
An error occurred where `invited_user` was `nil`. 

I was able to reproduce this by appending `resend=true` to the post data, however I don't think this was the likely cause in 'the field'. In any case, we only want to delete the user account at this point _if it already exists_.

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -63,7 +63,7 @@ private
   end
 
   def delete_user_record
-    invited_user.destroy!
+    invited_user.destroy! unless invited_user.nil?
   end
 
   def set_target_organisation


app/controllers/users/invitations_controller.rb
```diff
@@ -63,7 +63,7 @@ def authenticate_inviter!
   end
 
   def delete_user_record
-    invited_user.destroy!
+    invited_user.destroy! unless invited_user.nil?
```
koetsier, line 66: could shorten to '''invited_user&.destroy!```

koetsier (approved):
could address the comment if you want, but 👍

app/controllers/users/invitations_controller.rb
colinbm: I love a `&.` but feel the existing is more readable so will leave as is if you're ok with that... not merging yet until the deploy issue is resolved.

PR #1249, opened 2021-03-02
Don't prevent load of ActiveStorage when precompiling assets.

colinbm:
I suspect this was disabled as a optimisation as it wasn't relevant to
asset compilation, but now we've added the ActiveStorage route it's required.


sarahseewhy (approved):
LGTM

PR #1250, opened 2021-03-02
Resolve deploy build issue by applying correct fix.

colinbm:
On updating to Rails/ActiveStorage 6.1, our `ActiveStorage::BlobsController` stopped overriding the default.

I fixed this by renaming it to `ActiveStorage::SecureBlobsController` and referencing it directly in the routes.

As ActiveStorage is disabled when compiling assets, this caused a build failure when deploying to Staging.

I attempted to work around this, but there were further issues.

On further investigation, I identified that `ActiveStorage::BlobsController` had been renamed to `ActiveStorage::Blobs::RedirectController` which is why our custom controller was no longer automatically overriding it.

So, this PR reverts the previous changes to fix it, and implements the new fix, which is to rename our custom file accordingly,.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,6 +51,6 @@ COPY . .
 
 ARG RUN_PRECOMPILATION=true
 RUN if [ ${RUN_PRECOMPILATION} = 'true' ]; then \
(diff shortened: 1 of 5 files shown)


koetsier (approved):
lgtm

PR #1254, opened 2021-03-08
Update links to product pages

bethwalton:
We restructured the product pages when we added the text number to them. This is to use the new URLs.


sarahseewhy (approved):
Looks good to me. You may need to update the branch before it can be merged.

PR #1267, opened 2021-03-22
Revert session to cookies and handle overflow

colinbm:
As we've now established the cause of the large cookies leading to the cookie overflow error, we can now revert the switch to ActiveRecord storage.

We now check the length of the requested URL, and if it is more than 1/2 of the cookie storage available (4096) then the URL is not stored for redirection.


koetsier (approved):
nice work! 👍

PR #1275, opened 2021-04-07
Update rails to 6.1.3.1 to fix mimemagic issue.

colinbm:
See https://dev.to/cseeman/what-s-up-with-mimemagic-breaking-everything-he1

TLDR: `mimemagic` was pulled due to licensing issue, breaking any/all Rails 6.1+ builds.


Ian-Nicholls (approved):
LGTM so long as the version gods align and agree

PR #1281, opened 2021-04-27
Hide navigation from 2fa setup page

joelanman:
the 2fa setup page currently shows navigation, which is not right - you are not yet signed in. Clicking the navigation just takes you back to the 2fa setup page


koetsier:
this has been fixed

PR #1304, opened 2021-05-21
Add a warning page that 2fa is going to be mandatory soon.

koetsier:
Currently non-superadmins can skip setting up 2fa but that is
going to disappear soon. This commit adds a warning page that this
is happening and a link where users can get help.

image

image

Trello: https://trello.com/c/jCZwXfk6/1149-change-admin-site-to-warn-of-two-factor-authentication-changes

### DIFF
--- a/app/controllers/users/two_factor_authentication_setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication_setup_controller.rb
@@ -9,12 +9,6 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
   end
 
   def update
-    if params[:commit] == "Remind me next time"
-      flash[:notice] = "Two factor authentication setup skipped until next time"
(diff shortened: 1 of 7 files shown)


colinbm (approved):

joelanman:
thanks @koetsier can you explain the controller logic a bit so we're all clear on it?

koetsier:
> thanks @koetsier can you explain the controller logic a bit so we're all clear on it?

There's a new link on the 2FA page that only shows up if you're not a super admin (otherwise 2fa would already be mandatory) that takes you to a new controller `two_factor_authentication_unable_controller.rb`, which handles showing the 2FA warning screen (2nd screenshot). If the user decides to skip 2fa, the `update` method on that controller is called and 2FA is ignored for the session`. If not, one link takes the user back and the other sends the email

PR #1305, opened 2021-05-25
Remove `autofocus` attribute from form fields

colinbm:
## What

Remove `autofocus` attribute from form fields

## Why

Autofocus is frowned upon for accessibility purposes, as "users would expect to navigate the page in reading order, from top to bottom and left to right."


Ian-Nicholls (approved):
LGTM

PR #1306, opened 2021-05-25
Wrap the "no results" messages in `aria-role="status"`

colinbm:
## What

Implement `aria-role="status"` on "no results" messages.

## Why

Screen readers don't automatically know to read the no results message when searching. This adds the hint that it's relevant content.

## Note

Based on [this](https://www.davidmacd.com/blog/test-aria-live-display-none.html) the role needs to be on a containing element, as it relates to the inner content, so setting it on the `p` that's being shown/hidden wouldn't have an effect.

### DIFF
--- a/app/views/ips/index.html.erb
+++ b/app/views/ips/index.html.erb
@@ -27,9 +27,11 @@
       You need to add at least one location to offer GovWifi
     </p>
   <% else %>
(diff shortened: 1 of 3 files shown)


Ian-Nicholls (approved):
LGTM

Ian-Nicholls (approved):
LGTM (again)

PR #1308, opened 2021-05-27
Remove `hr` under main heading.

colinbm:
## What

The line under the heading was implemented with an `hr` tag. Remove it.

## Why

`hr` now has semantic meaning - "paragraph-level thematic break" - this doesn't make sense under the heading and the same appearance can be easily achieved with CSS.


Ian-Nicholls (approved):
LGTM

Ian-Nicholls (approved):
LGTM

PR #1309, opened 2021-06-01
Accessibility:  information and relationships fixes

colinbm:
## What

1. Move table caption content outside of table
2. Add headers to table rows

## Why 

1. Caption should be used as just that - our captions had heading, radius key, option to rotate, delete location button.
2. Some tables had rows with `label: value` - `label` in that context should be a `th`


Ian-Nicholls (approved):
LGTM

PR #1312, opened 2021-06-08
DAC Validation Pattern fixes

colinbm:
## What

* Update form errors template to repeat errors and link to fields
* Use form errors template where previously not used
* Set "Error: " prefix on page title within form errors template
* Repeat on Add IPs as this has custom errors

## Why

Because the DAC Doc says so.

## Note

The implementation of the links in the top errors is a bit of a hack - Devise just isn't designed to do that, but this works pretty well. 

Additionally adding the "Error: " prefix to the page titles is a similar hack. [In the top error template](https://github.com/alphagov/govwifi-admin/pull/1312/files#diff-66b69dc0610365d039ef8e12646a3e263511ec7e3deffec7914747225c6318b4R2) If we're displaying the errors, then we also take the previously set page title and add the prefix there.

### DIFF
--- a/app/views/ips/_confirm_remove_ip.html.erb
+++ b/app/views/ips/_confirm_remove_ip.html.erb
@@ -1,4 +1,4 @@
-<div class="govuk-error-summary">
+<div class="govuk-error-summary" role="alert">
   <h2 class="govuk-error-summary__title">
     Remove this IP address?
   </h2>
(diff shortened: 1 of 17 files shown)


Ian-Nicholls (dismissed):
LGTM

Ian-Nicholls (dismissed):
LGTM

Ian-Nicholls (approved):
LGTM

PR #1313, opened 2021-06-08
Improve permissions list accessibility

joelanman:
Permissions list was confusing for screen reader users

This PR:
 - fixes heading hierarchy (used to be H1, H3)
 - adds a heading for permissions
 - uses the Design System Summary List component - the recommended component for name value pairs, instead of ul (without role="list" nothing is announced for Safari/Voiceover)
 - uses text instead of images

Before:
screenshot of permission list before PR

After:

screenshot of permission list after PR


joelanman:
closed in favour of #1363

PR #1314, opened 2021-06-10
Add action prefixes to page titles.

colinbm:
## What

Add action titles... e.g. when you click "Remove Location" and you get a confirmation dialog, now the title changes to indicate the action

## Why

So that the page is more accessible.


sarahseewhy (approved):
LGTM

PR #1316, opened 2021-06-10
Update heading/hint, and add help for resetting 2FA

colinbm:
## What

Updates the copy for the heading & hint, and links them to the input field with `label` and `aria-describedby`

## Why 

The page does not say where the 6 digit code will come from and some people are thinking it will be via SMS
Accessibility fix


Ian-Nicholls (approved):
LGTM

Ian-Nicholls (approved):
LGTM

PR #1319, opened 2021-06-14
Correct the button text. Add class to link, to style it as link.

colinbm:
## What

* Change button "Authenticate" to "Continue"
* add `govuk-link` class to the email link

## Why

Because these were missed on earlier PR


Ian-Nicholls (approved):
LGTM

PR #1321, opened 2021-06-17
Disable the IP spoofing check that's being problematic with proxies.

colinbm:
## What

Set the config to skip the IP spoofing check

## Why

There are lots of false positives for this, due to proxies we have no control over. As it's not a big risk for us, the solution is to just disable the check.


Ian-Nicholls (approved):
LGTM

koetsier (approved):
👍

PR #1337, opened 2021-07-22
Remove option to skip 2FA

colinbm:
### What

* Remove the skip 2FA link
* Remove the linked page
* Remove the skip logic
* Update page design

### Why

We now require 2FA

Link to Trello card (if applicable): https://trello.com/c/yAmMqzjT


sarahseewhy (approved):

koetsier (changes requested):
Could you split this up into 2 PRs? One for removing 2FA and the other for displaying the key?

I think for the removing 2FA you might have to alter User#need_two_factor_authentication (https://github.com/alphagov/govwifi-admin/blob/364c146c7fea0ecce7469b3b07f5ceba3963c8f2/app/models/user.rb#L81) as well.

Could you also add a feature test demonstrating the change works as expected?

colinbm:
> Could you split this up into 2 PRs? One for removing 2FA and the other for displaying the key?

@koetsier How much do you want this? The key is a minor addition (bundled with design change), but I'll have to go back and redo the commits rather than cherry-pick to split them up.

spec/features/authentication/set_up_two_factor_authentication_spec.rb
```diff
@@ -12,134 +12,92 @@
...
 
-      it "shows a success message" do
-        expect(page).to have_content("Two factor authentication setup successful")
-      end
+    it "shows a success message" do
+      expect(page).to have_content("Two factor authentication setup successful")
+    end
 
-      it "redirects the user to the admin app" do
+    context "with a super admin user" do
```
koetsier, line 70: maybe, now, we can make the default in line 6 a regular user and then set the user here to be a super user? That might make more sense now we don't have any special business logic for the super user anymore.

koetsier (approved):
I had one more comment but otherwise 👍 🥳

PR #1346, opened 2021-07-29
Correct LineLength rule (should be within Layout)

colinbm:
### What

Use the correct namespace for LineLength Rubocop rule

### Why

It was wrong

Link to Trello card (if applicable): https://trello.com/c/qyOF6wLc


szd55gds (approved):
LGTM

PR #1358, opened 2021-08-05
add hidden text for screen readers

colinbm:
### What

Add visually hidden text to make links with identical text clearer for screen readers. e.g. Add IP addresses [for {location}]

### Why

Screen readers can list all links on a page... if the text of the link is repeated, it can be unclear what it will do.

Link to Trello card (if applicable): https://trello.com/c/isClWtzI


cbaines (approved):
Looks good to me :+1:

PR #1361, opened 2021-08-10
Display the success rate of requests per IP in the last day

koetsier:
### What
We want to display the rate of successful vs unsuccessful requests for an IP address

### Why
As an admin, I want to be able to see quickly if an IP address has a high error rate so I can take
action if something is misconfigured.

https://trello.com/c/4SN3SpTB/1488-show-successful-failed-requests-in-the-last-hour-in-admin-portal2

Added the "71% success rate" column in the screenshot below
image


cbaines (approved):
Looks OK to me :+1:

PR #1363, opened 2021-08-11
Make permissions more accessible

joelanman:
### What
- add a Permissions heading to explain the list
- add role=list so Voiceover treats it as a list
- make image alt blank and use hidden text instead on recommendation in the audit
- Changes 'allowed' to 'can' and 'not allowed' to 'cannot' so for example 'Can view team members'

### Why

From the audit:

> Users did not understand the purpose of the allowed/ not allowed images and assumed them to be decorative. This may be because the fact that these are presented to users as ‘images’ obscures the primary purpose of the information they are delivering.

Screen reader user comments:
>“When navigating in context, I located a multitude of images on the page described as ‘Allowed’. It seems that these images are present for decorative purposes, therefore providing no benefit to a screen reader user.”

When I tested the original page, the fact that the permissions do not have a heading, and are not read out as a list, seemed to contribute to the confusion so I have fixed those too.

Link to Trello card:

https://trello.com/c/JPF4EX90/1205-a11y-issue-non-text-content-missing-alts-page-14-dependency-on-testing-with-real-users-in-trello-card-no-1200 

Screenshot before:
image

Screenshot after:
image

### DIFF
--- a/app/views/memberships/_table.html.erb
+++ b/app/views/memberships/_table.html.erb
@@ -6,43 +6,52 @@
       <div class='govuk-grid-row'>
         <div class='govuk-grid-column-two-thirds'>
           <% if member.invitation_pending? || member.pending_membership_for?(organisation: current_organisation) %>
-            <h2 class='govuk-heading-s govuk-!-margin-bottom-0'><%= member.name %>
+            <h2 class='govuk-heading-m'><%= member.name %>
               <span class='govuk-!-padding-left-1 govuk-body text-dark-grey'><%= member.email %> (invited)</span>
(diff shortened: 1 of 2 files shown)


cbaines:
> * add a Permissions heading to explain the list
> * add role=list so Voiceover treats it as a list
> * make image alt blank and hidden text instead

It would be useful to add this information, and any more about why the changes are being made to the commit message.

joelanman:
@cbaines thanks! Done

bethwalton (approved):

colinbm (approved):

PR #1366, opened 2021-08-13
Make link buttons behave like buttons

colinbm:
### What

Add role, draggable, and data-module attributes to links styled as buttons.

### Why

As they're styled as buttons they should also behave like buttons, so as not to confuse the user.

Link to Trello card (if applicable): https://trello.com/c/SLDfWeAe


joelanman:
can't review at the mo as it wont run locally and theres no deploy

joelanman (approved):
tested and works well, thanks!

PR #1369, opened 2021-08-16
Update to current Sentry implementation as previous is deprecated

colinbm:
### What
Update to the current Sentry implementation

### Why
The old gem is deprecated. The new one also has the advantage of automatically filtering out the Mime Type error!

Link to Trello card (if applicable): https://trello.com/c/FWwuns8t


sarahseewhy (approved):
🚀 THANK YOU 🚀

PR #1385, opened 2021-08-31
Set a higher max in flight threshold for the Concourse pipeline

cbaines:
### What
Set a higher max in flight threshold for the Concourse pipeline

### Why
So that Concourse can process multiple Pull Requests at once. This is
helpful when there are lots of Dependabot Pull Requests open.


szd55gds (approved):
LGTM

PR #1387, opened 2021-09-01
Write CSV of service emails to Google Drive

colinbm:
### What
rake task to backup the org name and service email to a CSV on Google Drive.

### Why
In the event of a total AWS failure, we want an up to date list of email addresses, so we can maintain contact.

Link to Trello card (if applicable): https://trello.com/c/cofhRdNA


lib/gateways/google_drive.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def write(file_name:, folder_name:, upload_source:, mime_type: "text/plain")
+      file = Google::Apis::DriveV3::File.new
+      file.name = file_name
+      file.mime_type = mime_type
+
+      drive_file = @service.create_file(file, upload_source: upload_source)
+
+      if folder_name
```
cbaines, line 24: In what circumstances would `folder_name` be not truthy?

colinbm: If it were not provided... the default is to write the file to the root of the account, if you don't then assign in to a folder, it stays there.

In this specific use, a folder will always be provided, but the gateway could feasibly be reused and not.

cbaines: OK, thanks :+1:

cbaines (approved):
:+1: Looks good to me!

(just poke me on Slack if I need to re-approve)

PR #1393, opened 2021-09-03
Force the use of the sessions_start_username index in session calc

koetsier:
### What
Force the use of the sessions_start_username index in session calc

### Why
Mysql uses the SiteIP index instead of the much faster sessions_start_username
index when calculating the number of sessions for an IP address.
This change forces the use of this index.

In future we may add a new index which could speed things up
even more if necessary

Link to Trello card (if applicable): https://trello.com/c/QMqQGD6i/1545-handle-504-when-accessing-hmctss-page-in-the-super-admin-site

### DIFF
--- a/app/models/ip.rb
+++ b/app/models/ip.rb
@@ -34,7 +34,7 @@ class Ip < ApplicationRecord
 private
 
   def sessions(within: 1.day)
-    Session.where(siteIp: address).where("start > ?", Time.zone.today - within)
+    Session.use_index(:sessions_start_username).where(siteIp: address).where("start > ?", Time.zone.today - within)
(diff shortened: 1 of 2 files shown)


colinbm:
Why aren't we just creating the `siteIP,start` index now? I guess that may take a while to apply to 50m records!

koetsier:
I'm not sure how much adding start, siteIp (and even success) would increase speed. I tried testing a bit locally and using the  'start' index by itself already seems to deliver a huge performance boost. The other fields less so, so I wanted to try and see how this small change performs first.

colinbm (approved):

PR #1398, opened 2021-09-08
Use different header/label when on non-production

colinbm:
### What

On non-production systems, the header border, and label tag are yellow, and the tag is named for the environment rather than "beta"

Note, I've made the border the "full" yellow, rather than the pastel to match the tag, as it reduces its visibility against the white.

Screenshot 2021-09-08 at 05 48 43

### Why

We want to be able to see at a glance which environment we're in, and specifically know when we're on production, to avoid making any breaking changes to live data.

Link to Trello card (if applicable): https://trello.com/c/oXE4L3F0


szd55gds (approved):
This looks ok to me

PR #1399, opened 2021-09-09
Add paging to location view

koetsier:
### What
Add paging to location view
### Why
When there are a large number of locations for an organisation, the view
that displays the location will end up being very long and slow to load, sometimes
even timing out.

Paging the results will help performance

The paginator is similar to the one used by the Land Registry
(http://land-registry-elements.herokuapp.com/incubation-area/pagination-complex/demo)

Link to Trello card (if applicable): https://trello.com/c/QMqQGD6i/1545-handle-504-when-accessing-hmctss-page-in-the-super-admin-site

image
image
image


szd55gds (approved):
this looks solid to my eyes

app/views/super_admin/organisations/show.html.erb
```diff
@@ -44,7 +45,7 @@
         <%= render partial: "shared/manage_ips", locals: { location: location, show_ip_controls: false } %>
       <% end %>
     <% end %>
-
+    <%== pagy_nav_govuk(@pagy, link_extra: 'class: "govuk-link"') %>
```
colinbm, line 48: the `link_extra` appears to want raw html, so this should be `class="govuk-link"` - all instances.

app/views/ips/index.html.erb
```diff
@@ -23,6 +23,7 @@
     </div>
   <% end %>
 
+  <%== pagy_nav_govuk(@pagy) %>
```
colinbm, line 26: Think this should also have the class?

colinbm (changes requested):

app/views/super_admin/organisations/show.html.erb
koetsier: good catch!

app/helpers/pagy_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+    pagy.series.each do |item|  # series example: [1, :gap, 7, 8, "9", 10, 11, :gap, 36]
+      html << case item
+              when Integer then %(<li>#{link.call item}</li>) # page link
+              when String  then %(<li>#{item}</li>) # current page
+              when :gap    then %(<li">#{pagy_t('pagy.nav.gap')}</li>)   # page gap
+              end
+    end
+    html << %(</ul>)
+    if p_next
+      html << (link_to "Next", pagy_url_for(pagy, p_next), class: "pager__next", rel: "next")
```
colinbm, line 26: This needs the `govuk-link` class too.

koetsier: I changed the helper so it doesn't take the link-extras argument anymore, which simplifies the code quite a bit. I added the govuk-link class to the Prev and Next links

colinbm (approved):

PR #1403, opened 2021-09-10
Only validate postcode on pattern if not blank

colinbm:
### What
Only display one validation message for postcode (blank or valid)

### Why
When you don't enter a postcode, both validations fire, but it's only really relevant that you haven't entered it.

### DIFF
--- a/spec/models/location_spec.rb
+++ b/spec/models/location_spec.rb
@@ -32,24 +32,32 @@ describe Location do
     end
   end
 
+  # rubocop:disable Rails/SaveBang
   context "when validating postcode" do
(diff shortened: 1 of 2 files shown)


spec/models/location_spec.rb
```diff
@@ -32,24 +32,32 @@
     end
   end
 
+  # rubocop:disable Rails/SaveBang
```
colinbm, line 35: Robocop complains that we should be using `update!` below as we're not checking the result... however we are checking it, by inspecting the errors.

jimnarey (approved):

PR #1404, opened 2021-09-13
Unpin Sprockets gem and upgrade to version 4.

koetsier:
We had this pinned to version 3.7.2 because of a segmentation
fault that occurred when upgrading to 4.0+

After investigation it seems that Sprockets now uses the manifest.js
file to compile assets instead of the assets.rb file.

See: https://github.com/sass/sassc-rails/issues/122 and
https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs

This commit upgrades Sprockets and now uses the manifest.js file
to precompile assets which ensures the segmentation fault no
longer occurs.

Link to Trello card (if applicable):  https://trello.com/c/8Datu3zf/121-unpin-sprockets-gem-in-govwifi-admin


sarahseewhy (approved):
LGTM

PR #1406, opened 2021-09-15
Drop the permissions table as it's unused

colinbm:
### What
Drop the `permissions` table

### Why
It's no longer used, and is messy.

Link to Trello card (if applicable): https://trello.com/c/qkwhsASY


koetsier (approved):
👍

PR #1407, opened 2021-09-15
Remove the poster from the admin

colinbm:
### What
Remove the poster from within `/settings`

### Why
This was added in 2019 as a way to link to the poster externally, whilst still requiring authentication.

On the docs, we are hosting the poster, and making it available freely - there is no restriction to govwifi admins.

Checking the access logs, this has not been accessed in the last 3 months.


cbaines (approved):
Sounds good to me :+1:

PR #1446, opened 2021-11-02
Update default staging url

sarahseewhy:
### What

Update default staging URL to point to `staging-temp`.

### Why

* The staging components in the primary AWS account have been decommissioned therefore there is no more URL for `admin.staging.wifi.service.gov.uk`.
* We need to point to the URL for our secondary AWS account where staging components now run.
* This fixes broken email links generated when inviting a new user to Staging Admin site, they were pointing to the decommissioned `admin.staging.wifi.service.gov.uk` URL.


koetsier (approved):
👍

PR #1467, opened 2021-11-25
Remove google tag manager

koetsier:
### What
Remove google tag manager

### Why
Because we are not using it

https://trello.com/c/i7uOzQLd/1856-remove-google-tag-manager


cbaines (approved):
Sounds good to me :+1:

PR #1471, opened 2021-11-29
Enforce 2fa to download MOU

koetsier:
### What
When logged into the admin portal, but not completing 2FA the user is able to download the MOU through the download link if known. This PR enforces 2FA for download links as well.

### Why
It is more secure and consistent to enforce 2FA throughout the whole app

https://trello.com/c/4c5JBS65/1869-enforce-2fa-when-downloading-mou


cbaines (approved):
Don't have a lot of context on this, but it seems fine to me :+1:

PR #1474, opened 2021-11-30
Remove ASSET_PRECOMPILATION_ONLY references.

koetsier:
This was introduced to get around the fact that activestorage
requires the secret key base to be set

This no longer works because in config/initializers/activestorage.rb
explicit references are made to the ActiveStorage module.

An easier fix is to set the SECRET_KEY_BASE to a random string
when compiling assets

This also ensures production and staging are more alike

(see https://github.com/alphagov/govwifi-admin/pull/186#discussion_r224438529)


cbaines (approved):
Sounds like it's worth trying :+1:

PR #1478, opened 2021-12-01
Fix the CSRF protection

koetsier:
The protect_from_forgery action needs to be the first in the sequence
of before_action handlers otherwise the exception
'ActionController::InvalidAuthenticityToken' is thrown,


szd55gds (approved):
LGTM

PR #1479, opened 2021-12-01
Removes the password validation feature using Javascript

koetsier:
### What
Removes the password validation feature using Javascript

## Why
- This feature only applies to changing passwords and not to
creating them resulting in an inconsistent experience

- It is the only piece of Javascript that requires JQuery and
removing this allows us to remove the dependency on JQuery

- The code is untested

https://trello.com/c/CycP2x3T/1194-update-password-validation


cbaines (approved):
Sounds good to me :+1:

szd55gds (approved):
This looks reasonable (untested).

PR #1482, opened 2021-12-02
Switch to the Zeitwerk class loader

koetsier:
### What
We are using the classic rails autoloader and this PR makes the switch to the newer Zeitwerk one

### Why
Zeitwerk is the new Classloader for Rails. The classic version will soon be deprecated when Rails 7 is released. Furthermore it is good to follow the new class/module naming convention

https://trello.com/c/gQXR8Usc/1241-switch-to-zeitwerk-autoloader

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -21,13 +21,7 @@ module GovwifiAdmin
     config.exceptions_app = routes
 
     config.load_defaults 6.0
-
-    # our lib/ folder doesn't follow the Zeitwerk conventions for
-    # autoloading, so rely the classic loader instead.
-    config.autoloader = :classic
-
-    config.autoload_paths += %W[#{config.root}/lib]
-    config.autoload_paths += Dir["#{config.root}/lib/**/"]
+    config.autoloader = :zeitwerk
 
     # Force HTTPS for all requests except healthcheck endpoint
     config.force_ssl = true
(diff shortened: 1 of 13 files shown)


config/application.rb
```diff
@@ -24,10 +24,7 @@ class Application < Rails::Application
 
     # our lib/ folder doesn't follow the Zeitwerk conventions for
     # autoloading, so rely the classic loader instead.
-    config.autoloader = :classic
-
-    config.autoload_paths += %W[#{config.root}/lib]
-    config.autoload_paths += Dir["#{config.root}/lib/**/"]
+    config.autoloader = :zeitwerk
```
cbaines, line 27: Would be good to say what's motivated this change in the commit message.

config/application.rb
```diff
@@ -46,7 +43,7 @@ class Application < Rails::Application
 
     # Don't generate system test files.
     config.generators.system_tests = nil
-    config.eager_load_paths << Rails.root.join("lib")
+    config.eager_load_paths << Rails.root.join("lib").to_s
```
cbaines, line 46: Weird, searching the internet suggests this is because `Rails.root` is a `Pathname`, I guess it wasn't working before adding `.to_s`.

cbaines (approved):
I made one comment about a commit message, but apart from that, the changes look good to me :+1:

config/application.rb
koetsier: done

config/application.rb
koetsier: I thought it wasn't but after reading your comment I tried again and it works without the .to_s as well so I removed it again

PR #1483, opened 2021-12-03
Remove remnants of the now unused Registers.

koetsier:
### What
Remove remnants of the now unused Registers.

### Why
The Registers service was deprecated a while ago and in its place
we hardcoded a list of organisations.

There was still a lot of code left behind that dealt with the
Registers service that is now unused.

https://trello.com/c/3EKMiMJg/1243-remove-old-registers-code


sarahseewhy (approved):
LGTM

PR #1500, opened 2021-12-29
Upgrade to ruby 3

koetsier:
### What
Upgrade to Ruby 3.0.3

### Why
It is good to upgrade regularly so we don't fall behind


cbaines (approved):
Looks good to me :+1:

PR #1513, opened 2022-01-07
The docker-compose file now uses the healthcheck feature

koetsier:
We now use the docker-compose healthcheck feature to make sure
that MySql is running before the app starts

The mysql databases are now started by defining dependencies
within the docker-compose file, instead of starting them up
separately in the makefile.


cbaines (approved):
Sounds good to me :+1:

PR #1530, opened 2022-01-20
Make side bar menu display consistently

jimnarey:
### What
Prevents the side-bar menu from displaying when an email field error occurs, changes instances of 'register organisation' and 'create organisation' to 'add organisation' and adds a 'Cancel' (back) link.

### Why
To make the page display consistently in the case of all error types, use language consistently and match the design prototype

https://trello.com/c/KvrKdbSM/1946-make-display-of-navigation-sidebar-consistent-between-the-default-register-organisation-page-and-the-page-displayed-in-case-of-i

### DIFF
--- a/app/controllers/organisations_controller.rb
+++ b/app/controllers/organisations_controller.rb
@@ -21,7 +21,7 @@ class OrganisationsController < ApplicationController
       redirect_to root_path, notice: "#{@organisation.name} created"
     else
       @register_organisations = Organisation.fetch_organisations_from_register
-      
+
       @hide_sidebar = true
       render :new
     end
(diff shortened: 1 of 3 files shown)


app/controllers/organisations_controller.rb
```diff
@@ -21,7 +21,8 @@ def create
       redirect_to root_path, notice: "#{@organisation.name} created"
     else
       @register_organisations = Organisation.fetch_organisations_from_register
-
+      
+      @hide_sidebar = true
```
koetsier, line 25: Could you add a feature test for this?

koetsier (changes requested):
Looks good! One comment

app/controllers/organisations_controller.rb
jimnarey: Test added. Have also added a test to ensure the sidebar doesn't appear when the form is first displayed as the same logic was untested there.

koetsier (approved):
looks good to me

PR #1539, opened 2022-01-27
Make email error messages consistent

jimnarey:
### What

Adds the correct, design-system compliant error message to organisation e-mail fields in a number of forms in place of the former 'invalid email' message.

### Why

To be compliant with the design system.

https://trello.com/c/1XUyvh8D/1949-make-error-messages-in-case-of-a-blank-email-field-consistent-across-admin-site


koetsier (approved):
lgtm

PR #1551, opened 2022-02-07
Error messages on the 'register new org' pages are design system compliant

koetsier:
Ensure error messages display as intended on 'register new org' admin page

When leaving the organisation empty, a design-system compliant error message
is now shown. Value are now also retained if the input is not valid

https://trello.com/c/coP1mjaU/1951-ensure-error-messages-display-as-intended-on-register-new-org-admin-page

image


cbaines (approved):
Sounds good to me

PR #1565, opened 2022-02-15
Start using the form builder gem for a couple more forms

cbaines:
### What
Start using the form builder gem for a couple more forms

### Why
This is mostly aimed at adjusting the error messages shown by the
invite team member form. Since there's some overlap with validation
for the sign up form, I've ended up changing that as well.

Using the form builder gem means that the error messages don't all
start with the name of the field, which makes the format less
constrained.

I haven't changed the invite team member form checkboxes to be
generated through the gem provided helpers since I believe this part
of the form is changing shortly.

Link to Trello card: https://trello.com/c/CwZ50yCx/1950-make-error-messages-produced-by-devise-gem-design-system-compliant

### Before
image
### After
image

### Before
image
### After
image

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -27,7 +27,7 @@ protected
 
   def set_user_object_with_errors
     @user = User.new(sign_up_params)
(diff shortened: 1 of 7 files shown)


sarahseewhy (approved):
This looks fine to me.

PR #1583, opened 2022-03-03
Run tests with eager loading

cbaines:
### What
Run tests with eager loading

### Why
As this makes sure the tests pick up issues that will occur with eager
loading in the production Rails environment, like missing gems.


cbaines:
> I believe the addition of a couple of Gems here will fix that #1582

#1582 is merged, so the tests should pass here once this is rebased.

I'll leave it as is until it's reviewed though, as I think that's helpful.

koetsier (approved):
lgtm

PR #1584, opened 2022-03-03
Switch to using puma directly in the Dockerfile

cbaines:
### What
Switch to using puma directly in the Dockerfile

### Why
As this allows passing options like --threads, to increase the number
of threads being run.

Currently there are only a maximum of 5 threads being run, which leads
to poor utilisation of the tasks in ECS.


koetsier (approved):
👍

PR #1593, opened 2022-03-10
Remove the super_admin field from organisations

cbaines:
### What
Remove the super_admin field from organisations

### Why
This is now managed at the user level, rather than through association
with a super_admin organisation.

Link to Trello card: https://trello.com/c/08QFeITR/2093-admin-site-migrate-from-super-admin-organisations-to-individual-super-admins-5


koetsier (approved):
👍

PR #1596, opened 2022-03-15
Replace 'back to list' breadcrumb with 'Cancel' link

jimnarey:
### What
Replace 'back to list' breadcrumb with 'Cancel' link

### Why
More clearly represents the navigation which takes place
when clicking on the link, and generally easier to see.

https://trello.com/c/1DbWYmzf/2085-improve-the-ui-of-the-add-location-page-through-rewording-buttons-1

### DIFF
--- a/app/views/locations/new.erb
+++ b/app/views/locations/new.erb
@@ -6,7 +6,6 @@
   <%= f.govuk_error_summary %>
   <div class='govuk-grid-row'>
     <div class='govuk-grid-column-full'>
-      <%= link_to 'Back to list', ips_path, class: 'govuk-back-link' %>
       <h1 class='govuk-heading-l'>Add a location</h1>
     </div>
     <div class='govuk-grid-column-full'>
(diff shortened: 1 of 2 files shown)


app/views/locations/new.erb
```diff
@@ -6,14 +6,14 @@
...
     <div class='govuk-grid-column-full'>
-      <%= link_to 'Back to list', ips_path, class: 'govuk-back-link' %>
       <h1 class='govuk-heading-l'>Add a location</h1>
     </div>
     <div class='govuk-grid-column-full'>
       <%= f.govuk_text_field :address,
                               hint: { text: "For example, 10 High Street, London" } %>
       <%= f.govuk_text_field :postcode %>
-      <%= f.govuk_submit "Add location" %>
+        <%= f.govuk_submit "Add location" %>
```
cbaines, line 15: I think the indentation is slightly off here.

cbaines:
Merge commits are generally something to avoid, so it would be good if you could rebase on master prior to merging to get rid of the one here.

app/views/locations/new.erb
jimnarey: Thanks. Odd the linter doesn't catch that.

koetsier (approved):

PR #1597, opened 2022-03-16
Improve membership permissions

cbaines:
### What
Remove the NULLs from the membership permissions and set `can_manage_locations` to true where `can_manage_team` is true. 

### Why
So that the NULLs aren't confusing, and the values for these two fields match up with the 3 permission levels.

Link to Trello card: https://trello.com/c/qGBxPnU0/2087-changes-to-invite-team-member-and-edit-team-member-page-2

### DIFF
--- a/db/migrate/20220316112023_change_membership_can_manage_locations.rb
+++ b/db/migrate/20220316112023_change_membership_can_manage_locations.rb
@@ -1,0 +1,7 @@
+class ChangeMembershipCanManageLocations < ActiveRecord::Migration[6.1]
+  def up
+    execute "UPDATE memberships SET can_manage_locations = 1 WHERE can_manage_team = 1 AND can_manage_locations = 0"
+  end
+
+  def down; end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20220316112023_change_membership_can_manage_locations.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeMembershipCanManageLocations < ActiveRecord::Migration[6.1]
```
koetsier, line 1: I know that this was done in the past, but this is probably better done using SQL - reason is that if we ever rename the Membership model or one of the attributes this will stop working. See: https://guides.rubyonrails.org/v3.2/migrations.html#using-models-in-your-migrations.

koetsier (changes requested):
looking good. One comment

db/migrate/20220316112023_change_membership_can_manage_locations.rb
cbaines: Indeed, I've written the migration just using SQL now.

koetsier (approved):

PR #1598, opened 2022-03-16
Tweak the back link on the invite a team member page

cbaines:
### What
Tweak the back link on the invite a team member page

### Why
So that it's by the button at the bottom. This should be easier for
users to see.

The behaviour here is still not great, since the button does different
things for super admins, but that should be fixable in the future.

Link to Trello card: https://trello.com/c/qGBxPnU0/2087-changes-to-invite-team-member-and-edit-team-member-page-2


koetsier (approved):
👍

PR #1599, opened 2022-03-16
Update no organisation redirect

koetsier:
### What
The logic redirecting users when they are not part of an organisation needs to be improved. Currently, some useful pages, such as logging, cannot be accessed by a super admin that is not part of an organisations. 

### Why
We want to remove the requirement that a super admin has to be a member of an organisation. Only those pages that are specifically scoped to an org should disallow users without organisations.

To be created


cbaines (approved):
This seems good to me, I haven't tried it out locally but the changes look good :+1:

PR #1600, opened 2022-03-17
Comment out success rate table cell

jimnarey:
### What
Comment out success rate table cell

### Why
So that 504 errors do not occur on pages with large numbers of
IP addresses

https://trello.com/c/Xz191Qjn/2126-quick-fix-for-the-504-errors-on-the-admin-site


cbaines (approved):
Sounds good to me.

This does change the table layout a bit, since the cells end up taking thirds of the width, rather than a quarter, but this is fine.

koetsier (commented):
maybe just remove the code? If you want to re-instate it you could simply revert the PR

PR #1602, opened 2022-03-18
Fix side bar

koetsier:
### What
The side bar is now split in a 'super admin' set of links and
a 'regular user' set of links. If the user is a super admin, the 'super admin'
links are shown and if the user is a member of an organisation, the 'regular'
user links are shown. If both are shown at the same time, a line is shown to
distinguish the two sets of link

The link highlighting code was updated.  This logic worked by comparing the 
current path with a group of strings. It is very difficult to show that this
logic ends up being correct. Adding a new route also runs the risk that the logic 
breaks and confusing behaviour may result.

The new logic simply remembers the last side nav clicked and
 stores it in the session.

### Why
The Super User UI was a bit confusing. Accessing logs required switching to an organisation, which changed the side navigation links, allowing different options, including logging.  Super Admins should not need to join an organisation or switch to it to access important information

https://trello.com/c/RhvwTi1c/2129-clean-up-super-user-interface

image


cbaines (approved):
This seems OK to me.

I'm not familiar with using the session for breadcrumb/sidebar things, but it seems to work alright.

I'm also not that sure about having the organisation specific links (like Team members) show on the super admin pages, since the super admin pages (like the Super admins page) aren't associated with any particular organisation. This isn't a big issue though.

PR #1605, opened 2022-03-28
Change page contents to match updated design

jimnarey:
### What
Change page contents to match updated design

### Why
So that it is more consistent with the design system.

https://trello.com/c/zOmwfje7/2119-changes-to-support-page


cbaines (approved):
Looks good to me :+1:

PR #1608, opened 2022-03-31
Improve log page

koetsier:
### What
Add conditionals to the radio buttons of the log selection screen

Selecting which log to show, used to display a set of radio buttons, which took
the user to a new screen which allowed the user input further selection
criteria.

This has been replaced by one screen with conditionals to the radio button
actions, so that the admin user can search directly from the log page rather
then progressing to a 'holding' search page.

the user details screen is now accessed through a link in the side bar.

The user details screen for super users used to be a radio button option
in the log selection screen. However, it is not really showing any logging
and so it is moved to a link in the sidebar.

### Why
Makes for an easier user journey

https://trello.com/c/GAOjorAp/2088-changes-to-log-page-3

image
image


jimnarey (approved):
Works perfectly. There is slightly odd-looking drop-down for orgs without locations. See: https://trello.com/c/GAOjorAp/2088-changes-to-log-page-3

Suggest not worrying about it.

koetsier:
> Works perfectly. There is slightly odd-looking drop-down for orgs without locations. See: https://trello.com/c/GAOjorAp/2088-changes-to-log-page-3
> 
> Suggest not worrying about it.

Fixed it by not showing the locations dropdown when no locations are present

PR #1609, opened 2022-03-31
Make settings page consistent with new design

jimnarey:
### What
Make settings page consistent with new design

### Why
Change the layout of the page. Retains the basic page structure (tables)
rather than using divs/dls as per the prototype.

https://trello.com/c/dQVU1o17/2118-changes-to-the-setting-page-1

### DIFF
--- a/app/views/settings/_account.html.erb
+++ b/app/views/settings/_account.html.erb
@@ -1,22 +1,33 @@
 <h2 class="govuk-heading-m">Your account</h2>
 
-<table class="govuk-table">
+<table class="govuk-table govuk-!-margin-bottom-9">
   <caption class="govuk-visually-hidden">Settings</caption>
   <thead class="govuk-table__head govuk-visually-hidden">
     <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header govuk-!-width-one-half">Setting</th>
+      <th scope="col" class="govuk-table__header govuk-!-width-one-third">Setting</th>
       <th scope="col" class="govuk-table__header">Value</th>
       <th scope="col" class="govuk-table__header">Action</th>
     </tr>
   </thead>
   <tbody class="govuk-table__body">
     <tr class="govuk-table__row">
-      <th scope="row" class="govuk-table__cell govuk-!-width-one-half govuk-!-font-weight-regular">
+      <th scope="row" class="govuk-table__cell govuk-!-width-one-third">
+        Email Address
+      </th>
+      <td class="govuk-table__cell">
+        <%= current_user.email %>
+      </td>
+      <td class="govuk-table__cell govuk-table__cell--numeric"></td>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__cell govuk-!-width-one-third">
         Password
       </th>
       <td class="govuk-table__cell"></td>
       <td class="govuk-table__cell govuk-table__cell--numeric">
-        <%= link_to "Change your password", edit_user_registration_path, class: "govuk-link govuk-link--no-visited-state" %>
+        <%= link_to edit_user_registration_path, class: "govuk-link govuk-link--no-visited-state" do %>
+          Change <span class='govuk-visually-hidden'>your password</span>
+        <% end %>
       </td>
     </tr>
   </tbody>
--- a/app/views/settings/_organisation.html.erb
+++ b/app/views/settings/_organisation.html.erb
@@ -1,17 +1,17 @@
 <h2 class="govuk-heading-m">Organisation</h2>
 
-<table class="govuk-table">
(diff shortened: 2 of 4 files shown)


app/views/settings/_account.html.erb
```diff
@@ -4,19 +4,28 @@
...
+      <td class="govuk-table__cell govuk-table__cell--numeric"></td>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__cell govuk-!-width-one-third">
         Password
       </th>
       <td class="govuk-table__cell"></td>
       <td class="govuk-table__cell govuk-table__cell--numeric">
-        <%= link_to "Change your password", edit_user_registration_path, class: "govuk-link govuk-link--no-visited-state" %>
+        <%= link_to "Change", edit_user_registration_path, class: "govuk-link govuk-link--no-visited-state" %>
```
cbaines, line 28: I'd consider keeping the "your password" text, and just visually hiding it, since this'll provide extra context in situations like when a screen reader is used. See existing uses of `govuk-visually-hidden`.

app/views/settings/_organisation.html.erb
```diff
@@ -4,28 +4,26 @@
...
         Service email
       </th>
       <td class="govuk-table__cell">
         <%= current_organisation.service_email %>
       </td>
       <td class="govuk-table__cell govuk-table__cell--numeric">
-        <%= link_to edit_organisation_path(current_organisation.id), class: "govuk-link govuk-link--no-visited-state" do %>
-          Change <span class='govuk-visually-hidden'>service email</span>
-        <% end %>
+        <%= link_to "Change", edit_organisation_path(current_organisation.id), class: "govuk-link govuk-link--no-visited-state" %>
```
cbaines, line 21: Similar to my previous comment on visually hidden text, I'd want this to be kept here, rather than removed.

app/views/settings/_servers.html.erb
```diff
@@ -1,17 +1,22 @@
 <h2 class="govuk-heading-m">GovWifi servers</h2>
 
+<p class="govuk-hint govuk-!-margin-top-0 govuk-!-margin-bottom-0">
+  The
+  <%= link_to "offer GovWifi documentation", "https://docs.wifi.service.gov.uk/set_up/#step-3-create-a-new-wifi-installation", class: "govuk-link govuk-link--no-visited-state" %> explains how to connect your authenticators to these servers
```
cbaines, line 5: This line is really long, it would be easier to read in the code if it was broken across multiple lines.

app/views/settings/index.html.erb
```diff
@@ -3,9 +3,9 @@
 <h1 class="govuk-heading-l">Settings</h1>
 
 <%= render "account" %>
-
+<br>
 <%= render "organisation" %>
-
+<br>
```
cbaines, line 8: The prototype uses CSS to control the spacing (I see `govuk-!-margin-bottom-9"` used on the `dl` elements, which should be possible to use on the tables as well).

app/views/settings/_account.html.erb
jimnarey: Yes, much better, thank you.

app/views/settings/_organisation.html.erb
jimnarey: Likewise.

app/views/settings/_servers.html.erb
jimnarey: Odd that the linter doesn't pick this up. Should we change that?

app/views/settings/index.html.erb
jimnarey: Thank you!

app/views/settings/_servers.html.erb
cbaines: Potentially, although the linters are crude tools. People's judgement is often better.

app/views/settings/_account.html.erb
```diff
@@ -1,22 +1,33 @@
...
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__cell govuk-!-width-one-third">
         Password
       </th>
       <td class="govuk-table__cell"></td>
       <td class="govuk-table__cell govuk-table__cell--numeric">
-        <%= link_to "Change your password", edit_user_registration_path, class: "govuk-link govuk-link--no-visited-state" %>
+        <%= link_to edit_user_registration_path, class: "govuk-link govuk-link--no-visited-state" do %>
+          Change <span class='govuk-visually-hidden'>your password</span>
```
cbaines, line 29: Looks like most of the quotes (in this file at least) are double quotes, so I'd use double quotes here for consistency.

cbaines (approved):

PR #1610, opened 2022-04-07
Remove superfluous ips#index routes

koetsier:
### What
There were a lot of routes that all point to the same ips#index action. This
is confusing because the associated paths imply different actions exist whereas
in fact they are all synonyms.

### Why
Removing these makes the flow within the application clearer and shortens
the route table


cbaines (approved):
Sounds good to me :+1:

PR #1611, opened 2022-04-08
Remove superfluous memberships#index and mou#index routes

koetsier:
There were a lot of routes that all point to the the same action. This
is confusing because the associated paths imply different actions exist whereas
in fact they are all synonyms.

Removing these makes the flow within the application clearer and shortens
the route table


cbaines (approved):
Looks good to me :+1:

PR #1612, opened 2022-04-08
Implement several small fixes for the recent logging changes.

koetsier:
1. There are a number of superfluous routes that can now be removed
2. The username link in the userdetails view now point to the right page
3. A link in the help page now points correctly to the logging page


cbaines (approved):
Separate commits would make it a little easier to follow, but overall this looks good :+1:

PR #1613, opened 2022-04-08
Use GOV.UK form builder for the UI to add IPs to a Location

koetsier:
### What
Use GOV.UK form builder for the UI to add IPs to a Location

### Why
The GOV.UK form builder provides simpler view code. The previous approach
of using 'field_for' has been replaced by a form model which is more
straight-forward, and requires less code in the controller, view and model
 and is compatible with the form builder.


cbaines (approved):
Sounds good to me :+1:

PR #1636, opened 2022-04-28
Update whitelists forms to use GOV.UK form builder.

koetsier:
### What
Update whitelists forms to use GOV.UK form builder.
### Why
It simplifies code, ensures the frontend and error messages are consistent
and in line with the design system.

Use the whitelist form model to process all parameters instead of just
two. This simplifies controller logic.

https://trello.com/c/geuWyFBU/614-improve-error-messages-in-govwifi-admin


app/models/whitelist.rb
```diff
@@ -1,7 +1,37 @@
...
+  validate :validate_email_domain, unless: proc { |whitelist| whitelist.email_domain.blank? }
+
+  def next_step
+    { "first" => "first",
+      "second" => "second",
+      "third" => "third",
+      "fourth" => "fourth",
+      "fifth" => "fifth",
+      "sixth" => "sixth" }.fetch(step, "first")
+  end
```
cbaines, line 18: This looks odd to me, could it just be `step || "first"` or 

```
%w[first second third fourth fifth sixth].include?(step) ? step : "first"
```

koetsier: It is a bit odd I agree. I changed it to a set (which is backed by a hash), used a search similar to yours and used constants instead of strings

cbaines (approved):
Sounds good to me :+1:

PR #1648, opened 2022-05-06
Remove send_email use cases.

koetsier:
### What
Remove send_email use case classes

### Why
This introduced a lot of extra unnecessary complexity and allows
19 files to be deleted and only 3 added


cbaines (approved):
Looks good to me :+1:

PR #1655, opened 2022-05-16
Add former flexibility in JS execution defaulted-out by new CSP

jimnarey:
### What

The addition of a Content Security Policy prevents, by default,
the execution of inline scripts from other domains and the
evalution of strings as JS. The intention of the CSP (for now)
was simply to restrict which domains scripts could be downloaded
and executed from.

This commit re-introduces the ability to do both.

### Why
To enable execution of Google Analytics scripts.

https://trello.com/c/wy54K9rM/2263-analytics-on-admin-user-journeys


jimnarey:
Closed in order to separate CSP work from Google Analytics implementation.

PR #1664, opened 2022-05-25
Add model validation and database constraint

jimnarey:
### What
Add model validation and database constraint

### Why
So that the same location address cannot be entered twice for any given organisation.
A single organisation may still have two different addresses with the same postcode.

https://trello.com/c/YDAI28UQ/2217-upload-a-list-of-locations-and-associated-ip-addresses-to-connect-to-govwifi

### DIFF
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -9,6 +9,7 @@ class Location < ApplicationRecord
   validates_associated :ips
 
   validates :address, :postcode, presence: true
+  validates :address, uniqueness: { scope: :organisation_id }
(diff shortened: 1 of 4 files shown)


cbaines:
> So that the same location address cannot be entered twice for any given organisation.
> A single organisation may still have two different addresses with the same postcode.

If this is worth saying, it's worth having in the commit message.

cbaines (approved):
Would be good to update the commit message, but apart from that this looks good to me :+1:

jimnarey:
Thank you. Commit message updated.

PR #1690, opened 2022-07-18
Six digit code error message

iram-shehzadi:
### What
Change wording and styling for 2FA error message

### Why
Current wording for 2FA error message is inaccurate

Link to Trello card (if applicable): https://trello.com/c/Cb2s7VqC/2444-wording-for-2fa-error-message-and-help-text

Screenshot error message

### DIFF
--- a/app/views/layouts/_flash_notices.html.erb
+++ b/app/views/layouts/_flash_notices.html.erb
@@ -13,7 +13,11 @@
         There is a problem
       </h2>
       <div class="govuk-error-summary__body">
-        <p><%= msg %></p>
+        <ul class="govuk-list govuk-error-summary__list">
+          <li>
+            <%= msg %>
+          </li>
+        </ul>
       </div>
     </div>
   <% end %>
--- a/app/controllers/users/two_factor_authentication_setup_controller.rb
+++ b/app/controllers/users/two_factor_authentication_setup_controller.rb
@@ -17,7 +17,7 @@ class Users::TwoFactorAuthenticationSetupController < ApplicationController
       flash[:notice] = "Two factor authentication setup successful"
       disable_2fa_checks_for_session
     else
-      flash[:alert] = "Six digit code is not valid"
+      flash[:alert] = "The 6 digit code entered is not valid.<br />  Check the code sent in the email or request a new email.".html_safe
       render "show"
     end
   end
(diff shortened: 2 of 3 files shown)


app/views/layouts/_flash_notices.html.erb
```diff
@@ -13,7 +13,11 @@
         There is a problem
       </h2>
       <div class="govuk-error-summary__body">
-        <p><%= msg %></p>
+        <ul class="govuk-list govuk-error-summary__list">
+          <li>
+            <a href="#6-digit-error-summary"><%= msg.html_safe %></a>
```
koetsier, line 18: The linter complains about this line because 'html_safe' is considered a dangerous operation here because you don't quite know what the value of 'msg' is and if it has been validated properly. To fix, append 'html_safe' to  https://github.com/alphagov/govwifi-admin/blob/a512d2b1722dd853f81c56a96b3e0de9d3530e77/app/controllers/users/two_factor_authentication_setup_controller.rb#L20

The '<a>' tag shouldn't be here I think. This layout is shared between all views and so we can't put something specific to the 6 digit error summary in here. I know the design system says we should have a link here, but the previous code didn't and to change it would mean changing `Users::TwoFactorAuthenticationSetupController' and views and introducing a form model so we can use `form#govuk_error_summary' instead of a flash message which would probably be better but a larger change that may be outside the scope of this PR.

koetsier (changes requested):
Looking good, one comment 👍

app/views/layouts/_flash_notices.html.erb
iram-shehzadi: I have appended html_safe to flash alert message. I may have resolved this conversation before the change was pushed but the change is present on the latest commit

app/views/layouts/_flash_notices.html.erb
```diff
@@ -13,7 +13,11 @@
         There is a problem
       </h2>
       <div class="govuk-error-summary__body">
-        <p><%= msg %></p>
+        <ul class="govuk-list govuk-error-summary__list">
+          <li>
+            <a href="#6-digit-error-summary"><%= msg %></a>
```
koetsier, line 18: I think this `<a>` tag should't be here. I know it is in the design system, but it is specific to the 6 digit error and this layout is shared between a lot of views

koetsier (changes requested):
maybe squash some of the commits as well

koetsier (approved):
looks good! 👍

PR #1692, opened 2022-07-20
Remove superfluous classes

koetsier:
### What
Improve code
### Why
Because good code is better that bad


cbaines (approved):
Sounds good to me :+1:

PR #1708, opened 2022-08-10
Remove more use cases

koetsier:
### What
There were a few use cases left that did nothing. This PR removes them.
### Why
Simplifying code and removing superfluous code is good (-200 LOC)


cbaines (approved):
Sounds good to me :+1:

PR #1709, opened 2022-08-10
add hyperlinks to the logs for easy access to related logs

iram-shehzadi:
### What
Describe the change
Add hyperlinks to logs

### Why
If you spot something in the logs that needs investigating you often need to copy a piece of information and start a new log search to investigate. Improve navigation by introducing hyperlinks to the log tables which would make it simpler and faster for admins and super admins to look into issues

https://technologyprogramme.atlassian.net/browse/GW-271
Screenshot 2022-08-10 at 23 22 59
Screenshot 2022-08-10 at 23 33 37


koetsier (changes requested):
Good change, but could you add a feature test as well?

koetsier (approved):
Looks good to me! 🎉

PR #1712, opened 2022-08-11
Search for an ip address on the ips screen and address address to logs

iram-shehzadi:
### What
Describe the change
Extend search functionality to allow searches using the IP address and to add address information to the IP logs

### Why
The first page of the admin tool (/IPS) allows users to search by location and postcode but not by IP address.

If admins search for a user who is having problems, they will see the IP address they are connecting to but won't know what location the IP address is attributed to.

https://technologyprogramme.atlassian.net/browse/GW-270
Screenshot 2022-08-10 at 10 03 19
Screenshot 2022-08-10 at 10 01 41

### DIFF
--- a/app/controllers/logs_controller.rb
+++ b/app/controllers/logs_controller.rb
@@ -18,6 +18,7 @@ class LogsController < ApplicationController
               current_organisation.ip_addresses.intersection([log_search_form.ip])
             end
       logs = Gateways::Sessions.search(ips:, success:)
+      @current_location = Ip.find_by(address: ips).location if logs.present?
       render locals: { log_search_form:, logs: }
     when LogSearchForm::USERNAME_FILTER_OPTION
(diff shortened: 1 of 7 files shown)


app/controllers/logs_controller.rb
```diff
@@ -18,6 +18,7 @@ def index
               current_organisation.ip_addresses.intersection([log_search_form.ip])
             end
       logs = Gateways::Sessions.search(ips:, success:)
+      @current_location = Ip.find_by(address: log_search_form.ip).location if logs.present?
```
koetsier, line 21: You could user the `ips` variable here instead of `log_search_form.ip` Just so that it is 100% impossible to get information on a location a regular admin user does not control.

cbaines:
I think the commits could be cleaned up a bit, the merges can be removed and a couple look like they could be squashed in to others.

PR #1735, opened 2022-09-14
Change valid CSV delimiter from semi-colon to tab

jimnarey:
### What

Changes the delimiter expected by the bulk upload function from semi-colons to tabs.

Makes the same change for the RADIUS keys download for consitency.

### Why

MS Excel only permits saving CSVs with either commas or tabs as the delimiter. We can't
use commas as they typically occur in street addresses.

Link to Trello card (if applicable): https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-290

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -89,7 +89,7 @@ class LocationsController < ApplicationController
   end
 
   def download_keys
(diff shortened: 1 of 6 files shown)


koetsier (approved):
lgtm

PR #1736, opened 2022-09-14
fix bug for location logs filtering by status type and add tests to c…

iram-shehzadi:
…atch bug

### What
Fix bug for location logs filtering by status type and add tests to catch bug

### Why
When viewing logs via the logs page and selecting ‘location’, attempting to filter the logs by success/failure, ‘Page not found’ is displayed.

https://technologyprogramme.atlassian.net/browse/GW-448


koetsier (approved):
good fix 👍

PR #1744, opened 2022-09-28
add help when choosing a password

iram-shehzadi:
### What
Add help text on choosing a password

### Why
Not possible to offer clear criteria to users on creating on a password unless the technical criteria are refined

https://technologyprogramme.atlassian.net/browse/GW-285


spec/features/registration/change_password.spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      click_on "Change your password"
+      expect(page).to have_content("Change your password")
+    end
+
+    it "displays accordion with text 'help with choosing a strong password'" do
+      expect(page).to have_content("Help with choosing a strong password")
+    end
+
+    it "displays clear instructions on help with choosing a strong password" do
+      click_on "Help with choosing a strong password"
```
koetsier, line 26: the design system element that expands the help section is based on Javascript, which is not enabled in these tests. Seeing how it is from a library maybe we don't need to test that functionality?

koetsier (changes requested):
the spec file should be renamed so that the tests run (I don't think they did) Locally for me at least they don't pass.

koetsier (approved):

PR #1762, opened 2022-10-19
Create production and staging ECR Images as part of one codebuild project.

camdesgov:
### What
Create production and staging ECR Images as part of one Codebuild project.

### Why
This is a necessary work around for the limitations of Codebuild.


szd55gds (approved):
LGTM

PR #1770, opened 2022-10-26
Gw 484 change content to stop end users signing up as admin

iram-shehzadi:
### What
Inset text has been added to make end users aware these pages are only used for administrator sign up.

### Why
End users getting confused and signing up to be admins.

https://technologyprogramme.atlassian.net/browse/GW-484
Screenshot 2022-10-26 at 13 40 00
Screenshot 2022-10-26 at 14 37 11


szd55gds (approved):
A lot of repetition but I guess we need to give them as much notice as possible. LGTM.

PR #1773, opened 2022-11-03
update pagination for the admin site

iram-shehzadi:
### What
Implement Design System Pagination component from the GovWifi admin tool.

### Why
To ensure that our admin tool follows the best practice with the Government Design System.

https://technologyprogramme.atlassian.net/browse/GW-279
Screenshot 2022-11-02 at 23 48 05
Screenshot 2022-11-02 at 23 48 08
Screenshot 2022-11-02 at 23 48 40
Screenshot 2022-11-02 at 23 48 46

### DIFF
--- a/app/helpers/pagy_helper.rb
+++ b/app/helpers/pagy_helper.rb
@@ -4,28 +4,40 @@ module PagyHelper
   def pagy_nav_govuk(pagy)
     return "" if pagy.pages == 1
 
-    link   = pagy_link_proc(pagy, link_extra: "class='govuk-link'")
+    link   = pagy_link_proc(pagy, link_extra: "class='govuk-link govuk-pagination__link'")
     p_prev = pagy.prev
     p_next = pagy.next
 
     html = +%(<nav class="pager__govwifi govuk-body" role="navigation" aria-label="Pagination">)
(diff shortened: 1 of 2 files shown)


app/helpers/pagy_helper.rb
```diff
@@ -4,28 +4,40 @@ module PagyHelper
   def pagy_nav_govuk(pagy)
     return "" if pagy.pages == 1
 
-    link   = pagy_link_proc(pagy, link_extra: "class='govuk-link'")
+    link   = pagy_link_proc(pagy, link_extra: "class='govuk-link govuk-pagination__link'")
```
koetsier, line 7: This method now contains a lot of HTML. Wondering if it is possible to use a partial?

koetsier (commented):
Thinking using a partial may be possible?

app/helpers/pagy_helper.rb
iram-shehzadi: I have looked in to this and using a partial in the helper method is not conventional, taking the html out and using it in the views looks like a mess and doesn't really work. However, if we want to take the helper out completely and create a shared partial within the application views it'll be changing a lot of the application code as it is used in many places. I can do it but I'll create a separate branch for that. Can we approve this for now?

PR #1782, opened 2022-11-18
Gw 290 blank error location uploads

abdullaha1985:
### What
Describe the change
I changed few things in Location upload feature.
* I added a downloadable Template
* The upload now processes using Commas instead of Tabs.
* I have added extra columns in the upload to handle Address.
* Fix issue with Blank values for IP addresses.

### Why
Blank error messages showing for IP addresses.

https://technologyprogramme.atlassian.net/browse/GW-290

### DIFF
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -86,8 +86,10 @@ class LocationsController < ApplicationController
 
     @parent_organisation = current_organisation
     BulkUpload::BulkUpload.add_location_data(data, @parent_organisation)
+    @locations = @parent_organisation.locations.select(&:new_record?)
+    @ips_count = @locations.flat_map(&:ips).count
     @parent_organisation.save!
-    redirect_to(ips_path, notice: "Successfully uploaded locations")
+    redirect_to(ips_path, notice: "Upload complete. You have saved #{@locations.count} new locations and #{@ips_count} new IP addresses")
   rescue ActiveRecord::RecordInvalid
     redirect_to(ips_path, flash: { error: "Uploading data failed. Please try again." })
   rescue StandardError => e
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -59,6 +59,10 @@ class LocationsController < ApplicationController
     end
   end
 
+  def download_locations_upload_template
+    send_file "public/locations_upload_template.csv", disposition: "inline"
+  end
+
   def bulk_upload
     @upload_form = UploadForm.new
   end
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -74,6 +74,8 @@ class LocationsController < ApplicationController
     else
       @parent_organisation = current_organisation
       BulkUpload::BulkUpload.add_location_data(@upload_form.data, @parent_organisation)
+      @new_locations = @parent_organisation.locations.select(&:new_record?)
+      @new_ips_total = @new_locations.flat_map(&:ips).count
       @parent_organisation.validate
     end
   end
--- a/app/models/upload_form.rb
+++ b/app/models/upload_form.rb
@@ -1,29 +1,59 @@
 class UploadForm
   include ActiveModel::Model
   include ActiveModel::Validations
+  include BulkUpload::AddressHeaders
 
   validate :validate_reading_csv
   validate :validate_csv
+  validate :validate_data_present
+  validate :validate_header
 
   def validate_reading_csv
-    @data ||= CSV.read(upload_file, col_sep: "\t")
+    result = CSV.read(upload_file, headers: true)
+    @data ||= result.to_a[1..]
+    @headers ||= result.headers
   rescue TypeError
     errors.add(:upload_file, :csv_error, message: "Choose a file before uploading")
   rescue CSV::MalformedCSVError, CSV::Parser::InvalidEncoding
-    errors.add(:upload_file, :csv_error, message: "Invalid file type")
+    errors.add(:upload_file, :csv_error, message: "File must be csv format")
   end
 
   def validate_csv
     errors.add(:upload_file, :csv_error, message: "Invalid data structure") unless csv_valid?
   end
 
+  def validate_data_present
+    errors.add(:upload_file, :csv_error, message: "File has no data") if first_row_empty?
+  end
+
+  def validate_header
+    errors.add(:upload_file, :csv_error, message: "File must use correct header names") if headers_match?
+  end
+
   attr_accessor :upload_file, :data
 
 private
 
+  def headers_match?
+    return false if @data.nil?
+
+    headers_in_csv = @headers[0..HEADER_COLS.length - 1].map(&:strip).map(&:downcase)
+
+    headers_in_csv != HEADER_COLS.map(&:downcase)
+  end
+
   def csv_valid?
+    return false if @data.nil?
+
     @data.is_a?(Array) &&
       @data.all? { |row| row.is_a? Array } &&
       @data.all? { |row| row.length >= 2 }
+  end
+
+  def first_row_empty?
+    return false if @data.nil?
+    return true if @data[1].nil?
+
+    @data[1].all?(&:blank?)
   end
 end
--- a/app/models/upload_form.rb
+++ b/app/models/upload_form.rb
@@ -30,6 +30,14 @@ class UploadForm
     errors.add(:upload_file, :csv_error, message: "File must use correct header names") if headers_match?
   end
 
+  def self.build_address(row)
+    "#{row[ADDRESS_LINE_1]}
+    #{row[ADDRESS_LINE_2]}
+    #{row[ADDRESS_LINE_3]}
+    #{row[CITY]}
+    #{row[COUNTY]}".squish
+  end
+
   attr_accessor :upload_file, :data
 
 private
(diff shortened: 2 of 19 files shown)


koetsier:
Could you clean up the commit history?

lib/bulk_upload/bulk_upload.rb
```diff
@@ -3,12 +3,26 @@
...
     def self.add_location_data(data, organisation)
-      data.each do |row|
-        location = organisation.locations.new(address: row[0], postcode: row[1])
-        row.slice(2..).each do |ip_address|
+      data_without_header = data.drop(1)
+
+      data_without_header.each do |row|
+        location = organisation.locations.new(address: build_address(row), postcode: row[5]&.strip)
+        row.slice(6..).each do |ip_address|
+          next if ip_address.blank?
```
koetsier, line 11: you could also use `row.slice(6..).reject(&:blank).each`

spec/fixtures/csv/locations_upload_invalid_headers.csv
```diff
@@ -0,0 +1,3 @@
+Invalid,Address line 2,Address line 3,City,County,Postcode,IP address,IP address,IP address,IP address,IP address
+122,London road,,London,Greater London,LE2 0EN
+101,Avenue  street,,Leicester,Leicestershire,SE2 9TT
```
koetsier, line 3: Git complains about there not being a newline.Maybe just add one to make it stop 😁

koetsier (changes requested):

app/controllers/locations_controller.rb
```diff
@@ -70,6 +74,8 @@ def upload_locations_csv
     else
       @parent_organisation = current_organisation
       BulkUpload::BulkUpload.add_location_data(@upload_form.data, @parent_organisation)
+      @locations = @parent_organisation.locations.select(&:new_record?)
```
koetsier, line 77: Could this variable be a bit more descriptive? Like `@new_locations` or `@added_locations` or something so we know it is not just all locations - same with the `@ip_count`

lib/bulk_upload/bulk_upload.rb
```diff
@@ -3,12 +3,24 @@
...
+      data_without_header.each do |row|
+        location = organisation.locations.new(address: build_address(row), postcode: row[5]&.strip)
+        row.slice(6..).reject(&:blank?).each do |ip_address|
           location.ips.new(address: ip_address)
         end
       end
     end
+
+    def self.build_address(row)
+      addres_1 = row[0]
```
koetsier, line 17: there's a lot of magic numbers here that correspond to upload_form.rb - line 10. Could we do something about that?

koetsier (changes requested):
Other comments:
The link to the Jira card in the description is pointing to the wrong ticket I think.
Could you add something to the commit message to say what this commit does?
If possible, you could split this PR up into different commits so it tells a more easily understood story

lib/bulk_upload/bulk_upload.rb
abdullaha1985: @koetsier Im not sure what you mean? Can you please give me an example. Thanks

app/assets/stylesheets/components/flash_messages.scss
```diff
@@ -2,6 +2,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  margin-left: -15px;
```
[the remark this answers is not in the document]
abdullaha1985: This aligns Flash notice with content below.
It was requested by Roger.

Shall I still remove it?

success Screenshot

abdullaha1985:
> Other comments: The link to the Jira card in the description is pointing to the wrong ticket I think. Could you add something to the commit message to say what this commit does? If possible, you could split this PR up into different commits so it tells a more easily understood story

Regarding commits, sorry I misunderstood when you said commits. I squashed all commits to one. I will remember in future to leave few descriptive commits.

app/assets/stylesheets/components/flash_messages.scss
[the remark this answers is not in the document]
koetsier: I think something is different in this view. When replacing the MOU for instance, the flash message now looks like this:
image

lib/bulk_upload/bulk_upload.rb
koetsier: The fact that the first address line is in column 0 and the postcode in column 5 is encoded in https://github.com/alphagov/govwifi-admin/blob/de7bd0180f06b541ac78d6d21c21c1c3ced81765/app/models/upload_form.rb#L10 - I think it would be a good idea to have one source of truth of what the structure of the file looks like and then refer to that instead of to column numbers. That was it is easier to change and harder to make mistakes

app/models/upload_form.rb
```diff
@@ -4,26 +4,52 @@ class UploadForm
...
   def validate_csv
     errors.add(:upload_file, :csv_error, message: "Invalid data structure") unless csv_valid?
   end
 
+  def validate_data_present
+    errors.add(:upload_file, :csv_error, message: "File has no data") if first_row_empty?
+  end
+
+  def validate_header
+    return if @data.nil?
```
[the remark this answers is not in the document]
abdullaha1985: I tested uploading no file, the result is all validations are ran meaning @data is nil before ‘TypeError’ is raised by CSV.read.

e.g in upload_form.rb.
```
...
    @data ||= CSV.read(upload_file)
  rescue TypeError
    errors.add(:upload_file, :csv_error, message: "Choose a file before uploading")
```

app/models/upload_form.rb
```diff
@@ -4,26 +4,65 @@ class UploadForm
 
   validate :validate_reading_csv
   validate :validate_csv
+  validate :validate_data_present
+  validate :validate_header
+
+  HEADER_COLS = ["Address line 1", "Address line 2", "Address line 3", "City", "County", "Postcode", "IP address"].freeze
```
koetsier, line 10: Maybe just add some constants?

something like
```
  module INDICES
    ADDRESS_LINE1 = 0
    ADDRESS_LINE2 = 1
    ADDRESS_LINE3 = 2
    CITY = 3
    COUNTY = 4
    POSTCODE = 5
    LAST = 5
  end
```

app/models/upload_form.rb
```diff
@@ -4,26 +4,65 @@ class UploadForm
...
 
   def validate_reading_csv
-    @data ||= CSV.read(upload_file, col_sep: "\t")
+    @data ||= CSV.read(upload_file)
```
koetsier, line 13: I just read up a bit on `CSV` on the ruby docs, and now we're using headers you could use `headers: true` here, which then returns a `CSV::Table` object which may make some of the code a bit simpler. 

There are a few more options described here: https://ruby-doc.org/stdlib-3.0.0/libdoc/csv/rdoc/CSV.html which may be useful? What do you think

koetsier (changes requested):
Could you also squash the commits? (Or sometimes it is even better to have multiple tidied-up commits that are easier to understand)

app/models/upload_form.rb
```diff
@@ -4,26 +4,66 @@ class UploadForm
...
 
   def validate_csv
     errors.add(:upload_file, :csv_error, message: "Invalid data structure") unless csv_valid?
   end
 
+  def validate_data_present
+    errors.add(:upload_file, :csv_error, message: "File has no data") if first_row_empty?
+  end
+
+  def validate_header
```
koetsier, line 31: This validation is different from the one in line 23 and 27 - they both use a private function to determine if the form is valid. Maybe we can make them look the same?

lib/bulk_upload/address.rb
```diff
@@ -0,0 +1,17 @@
+module BulkUpload
+  module Address
+    ADDRESS = {
```
koetsier, line 3: This constant should probably be named something else - it is not an address.

lib/bulk_upload/bulk_upload.rb
```diff
@@ -3,9 +3,11 @@
 module BulkUpload
   module BulkUpload
     def self.add_location_data(data, organisation)
-      data.each do |row|
-        location = organisation.locations.new(address: row[0], postcode: row[1])
-        row.slice(2..).each do |ip_address|
+      data_without_header = data.drop(1)
```
koetsier, line 6: It may be more obvious to remove the header in UploadForm. We don't do anything with the headers after reading them in anyway

koetsier (changes requested):

lib/bulk_upload/bulk_upload.rb
```diff
@@ -2,10 +2,12 @@
...
 module BulkUpload
   module BulkUpload
+    include AddressHeaders
+
     def self.add_location_data(data, organisation)
       data.each do |row|
-        location = organisation.locations.new(address: row[0], postcode: row[1])
-        row.slice(2..).each do |ip_address|
+        location = organisation.locations.new(address: UploadForm.build_address(row), postcode: row[POSTCODE]&.strip)
+        row.slice(IP_ADDRESS..).reject(&:blank?).each do |ip_address|
```
koetsier, line 10: maybe you could `strip` the ip address as well, seeing how all the other fields are

koetsier (changes requested):
One small comment - could you clean up the commits and put some descriptions in them?

koetsier (approved):
Ok looks good now!

PR #1784, opened 2022-11-22
fix delete functionality and refacto

iram-shehzadi:
### What
Fix broken feature to remove organisations added to our allow list.

### Why
Feature to remove organisations added to our allow list does not work. When you try to remove an organisation from the allow list, you get a ‘Page not found' and the removal doesn’t happen.

https://technologyprogramme.atlassian.net/browse/GW-406


spec/features/super_admin/allowlists/organisation_names/delete_custom_organisation_spec.rb
```diff
@@ -16,15 +16,15 @@
     end
 
     it "removes the correct organisation name from the list" do
-      click_on "Yes, remove this organisation"
+      click_on "Yes, remove DummyOrg1 from the allow list"
```
koetsier, line 19: I think we should have a test here that would have failed before your fix and passes after.

koetsier (approved):
Looks good. One comment

PR #1799, opened 2022-12-20
Gw 290 blank error location uploads

abdullaha1985:
### What
Describe the change

Fix for Locations Upload error.
Currently system does not allow CSV with only one Location Address to be uploaded.

### Why
I changed the validations in upload_form.rb to allow CSV with one row of data.
It will only throw error when no data exist.

https://technologyprogramme.atlassian.net/browse/GW-290


koetsier (approved):
lgtm

PR #1814, opened 2023-02-06
Fix elasticsearch client to version 7

koetsier:
We're currently running Elasticsearch version 7 and the version 8 client is incompatible.


iram-shehzadi (approved):
lgtm

PR #1822, opened 2023-03-08
fix bug for the filtering of logs using the success status

iram-shehzadi:
### What
Filter for the logs partially works

### Why
We’re not able to filter logs using the three options

We’re not able to see filter status selected 

https://technologyprogramme.atlassian.net/browse/GW-777


koetsier (approved):
👍

PR #1848, opened 2023-06-14
create validation to stop admins from being locked out of their account

iram-shehzadi:
### What
Introduce enhancements to the administrator management process. The changes include allowing modifications to administrators only if there are two or more administrators present within an organisation. It also incorporates the addition of confirmations, reminders, and error summaries to provide a more secure and reliable experience.

### Why
The objective of these enhancements is to address the following goals:

Security: By requiring a minimum of two administrators, we ensure that organisations are not left without an admin in case of any unforeseen circumstances, such as one admin leaving or getting locked out of their account.

Reliability: By incorporating confirmations, reminders, and error summaries, we improve the user experience and help administrators make informed decisions while managing the administrator accounts. This reduces the risk of errors and improves the overall reliability of the system.

By implementing these improvements, we aim to create a more robust and user-friendly environment for administrators and organisations. The changes will enhance the overall security posture and operational continuity, reducing the likelihood of account lockouts and ensuring organisations have appropriate contacts for support.

https://technologyprogramme.atlassian.net/browse/GW-624


koetsier (changes requested):
A few comments. Let me know what you think!

app/views/memberships/edit.html.erb
```diff
@@ -1,15 +1,18 @@
...
-<%= link_to "Back", :back, class: "govuk-back-link" %>
 <%= render "confirm_remove_team_member" if params[:remove_team_member] %>
 
-<h1 class="govuk-heading-l govuk-!-margin-top-2 govuk-!-margin-bottom-3"><%= @membership.user.name %></h1>
+  <%= render "confirm", locals: { membership: @membership, permission_level_data: @permission_level_data } if params[:confirm] %>
+
+<h1 class="govuk-heading-l govuk-!-margin-top-2 govuk-!-margin-bottom-3">Manage permissions for <%= @membership.user.name %></h1>
 <p class="govuk-body govuk-!-margin-bottom-4"> <%= @membership.user.email %> </p>
 
+<%= render "memberships/minimum_two_administrators_required" %>
```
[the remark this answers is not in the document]
iram-shehzadi: In the Figma its always there. Will confirm with Crispin and leave for now

app/views/memberships/edit.html.erb
```diff
@@ -1,19 +1,21 @@
...
 <%= render "confirm_remove_team_member" if params[:remove_team_member] %>
 
-<h1 class="govuk-heading-l govuk-!-margin-top-2 govuk-!-margin-bottom-3"><%= @membership.user.name %></h1>
+<%= render "confirm_change_permissions", membership: @membership, permission_level_data: @permission_level_data if params[:confirm] %>
+
+<h1 class="govuk-heading-l govuk-!-margin-top-2 govuk-!-margin-bottom-3">Manage permissions for <%= @membership.user.name %></h1>
 <p class="govuk-body govuk-!-margin-bottom-4"> <%= @membership.user.email %> </p>
 
+<%= render "memberships/minimum_two_administrators_required" %>
+
```
koetsier, line 11: I think you can simplify the code a bit by letting the form below submit a 'get' request to the `edit_membership_path` path and add a `hidden_field` with `confirm =true` That way, the code is similar to the confirmation of removing a user and you can remove the `if-then-else` in the `update` method of the controller

koetsier, line 11: Maybe `confirm` could be called something more descriptive to distinguish it from confirming removing a user?

koetsier, line 11: I think you can simplify the code a bit by letting the form below submit a 'get' request to the `edit_membership_path` path and add a `hidden_field` with `confirm =true` That way, the code is similar to the confirmation of removing a user and you can remove the `if-then-else` in the `update` method of the controller

app/views/memberships/_confirm_change_permissions.html.erb
```diff
@@ -0,0 +1,15 @@
+<% content_for :page_title, "Change permissions - #{yield :page_title}", flush: true %>
+
+<% if @membership.can_manage_team_as_admin_with_limitations? %>
+  <%= render "memberships/cannot_change_permissions" %>
```
koetsier, line 4: I think you probably have to put a conditional on the `update` method in the controller as well and not only in the view. Otherwise you could in theory bypass this by sending a custom update request.Maybe also add a request spec to verify that

PR #1851, opened 2023-06-29
Vs code dev container setup

A1Robson:
### What
Added folder and files for vscode related container setup

### Why
Enable simple vscode setup and configuration. 
Compose individual services in a way which supports vscode debugging.

### Expected impact
Zero impact intended - should only affect vscode development.


koetsier (approved):
LGTM

PR #1852, opened 2023-06-29
Gw 404 change organisations default sorting behaviour

A1Robson:
### What
Made the initial page view of Organisations (as super admin) sort the Orgs by Name in ascending alphabetic order instead of the previous sorting by creation date.
Also changed the sort order of the other columns so that they show the Orgs with the highest Ip count or most recently signed MOU etc. Thinking being that more recent is more interesting and bigger is better, so show those on first click.

### Why
The initial sort order for the Organisations being based on creation date didn't make sense based on the usual use case when viewing organisations.


koetsier (approved):
lgtm. Could you squash commits?

A1Robson:
tidied this and created pr on another branch: https://github.com/alphagov/govwifi-admin/pull/1853

PR #1853, opened 2023-06-30
Changes to Organisations default sort order

A1Robson:
### What
Made the initial page view of Organisations (as super admin) sort the Orgs by Name in ascending alphabetic order instead of the previous sorting by creation date.
Also changed the sort order of the other columns so that they show the Orgs with the highest Ip count or most recently signed MOU etc. Thinking being that more recent is more interesting and bigger is better, so show those on first click.

### Why
The initial sort order for the Organisations being based on creation date didn't make sense based on the usual use case when viewing organisations.


app/helpers/organisations_helper.rb
```diff
@@ -2,7 +2,13 @@ module OrganisationsHelper
   def sort_link(column, title = nil)
     title ||= column.titleize
     css_class = column == sort_column ? "current #{sort_direction}" : nil
-    direction = column == sort_column && sort_direction == "asc" ? "desc" : "asc"
+
+    direction = if column == "name"
```
koetsier, line 6: It is a little unclear to me what this change actually does. Can this be simplified? I tried the original versions and the tests still pass so I'm not quite clear what it is supposed to do

koetsier (changes requested):

A1Robson:
Will have a look...
--
Aaron

app/helpers/organisations_helper.rb
A1Robson: Hmm, my tests fail if I revert the helper only. The intent of the change to the helper code is that: when creating a link for a column which is _not_ the currently sorted column, the default direction is ascending for the name column but is descending for the other columns.

koetsier (approved):
👏

PR #1854, opened 2023-07-01
Ruby version upgrade and tidy

A1Robson:
### What
Upgrading Ruby version to 3.2.2 and bundler to 2.4.10

### Why
Keep things up to date


koetsier (approved):
👍

PR #1855, opened 2023-07-01
Minor updates

A1Robson:
### What
Deleted the old docker compose concourse file as don't believe is used (please confirm).
Added an extra linting target for vscode

### Why
Tidy


koetsier (approved):
👍

PR #1862, opened 2023-07-06
feat: add commit sha as image tag and include all tags when pushing

jlhco:
### What
Tag the image with the commit sha and include all tags when pushing the image up

### Why
So we can view this tag in the pipeline

Link to JIRA / Trello card (if applicable): https://technologyprogramme.atlassian.net/browse/GW-819


szd55gds (approved):
LGTM

PR #1879, opened 2023-07-24
create validation to stop admins from being locked out of their account 

iram-shehzadi:
### What
Introduce enhancements to the administrator management process. The changes include allowing modifications to administrators only if there are two or more administrators present within an organisation. It also incorporates the addition of confirmations, reminders, and error summaries to provide a more secure and reliable experience.

### Why
The objective of these enhancements is to address the following goals:

Security: By requiring a minimum of two administrators, we ensure that organisations are not left without an admin in case of any unforeseen circumstances, such as one admin leaving or getting locked out of their account.

Reliability: By incorporating confirmations, reminders, and error summaries, we improve the user experience and help administrators make informed decisions while managing the administrator accounts. This reduces the risk of errors and improves the overall reliability of the system.

By implementing these improvements, we aim to create a more robust and user-friendly environment for administrators and organisations. The changes will enhance the overall security posture and operational continuity, reducing the likelihood of account lockouts and ensuring organisations have appropriate contacts for support.

https://technologyprogramme.atlassian.net/browse/GW-624

### DIFF
--- a/app/controllers/memberships_controller.rb
+++ b/app/controllers/memberships_controller.rb
@@ -1,7 +1,7 @@
 class MembershipsController < ApplicationController
   before_action :set_membership, only: %i[edit update destroy]
   before_action :validate_can_manage_team, only: %i[edit update destroy]
-  before_action :validate_has_two_confirmed_admin_users, only: %i[update destroy]
+  before_action :validate_preserve_admin_permissions, only: %i[update destroy]
   skip_before_action :redirect_user_with_no_organisation, only: %i[destroy edit update]
 
   def edit
@@ -98,7 +98,7 @@ private
     end
   end
 
-  def validate_has_two_confirmed_admin_users
+  def validate_preserve_admin_permissions
     raise ActionController::RoutingError, "Not Found" if @membership.preserve_admin_permissions?
   end
 end
--- a/spec/features/team/remove_a_team_member_spec.rb
+++ b/spec/features/team/remove_a_team_member_spec.rb
@@ -1,39 +1,48 @@
 describe "Remove a team member", type: :feature do
-  let!(:organisation) { create(:organisation) }
-  let!(:user) { create(:user, organisations: [organisation]) }
-  let!(:another_user) { create(:user, organisations: [organisation]) }
+  let(:organisation) { create(:organisation) }
+  let(:user) { create(:user, organisations: [organisation]) }
+  let(:second_user) { create(:user, organisations: [organisation]) }
+  let(:third_user) { create(:user, organisations: [organisation]) }
 
   before do
+    user.membership_for(organisation).update!(can_manage_team: true)
+    second_user.membership_for(organisation).update!(can_manage_team: true)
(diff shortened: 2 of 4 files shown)


app/controllers/memberships_controller.rb
```diff
@@ -1,6 +1,7 @@
 class MembershipsController < ApplicationController
   before_action :set_membership, only: %i[edit update destroy]
   before_action :validate_can_manage_team, only: %i[edit update destroy]
+  before_action :validate_has_two_confirmed_admin_users, only: [:update]
```
koetsier, line 4: There probably needs to be a check on 'destroy' as well right?

app/controllers/memberships_controller.rb
```diff
@@ -95,4 +97,8 @@ def validate_can_manage_team
       raise ActionController::RoutingError, "Not Found"
     end
   end
+
+  def validate_has_two_confirmed_admin_users
+    raise ActionController::RoutingError, "Not Found" if @membership.cannot_change_permissions?
```
koetsier, line 102: I'm not sure this works. If you have 2 admins and remove one, this validation runs before the admin is removed and therefore passes. Also, suppose, you have one administrator user and one view_only user and you want to change the view_only user to administrator this validation will I think not allow you to do this.

Usually, with this kind of thing, the model is changed in memory and then validated before persisting to the database. Maybe we should have think on how to do this here because it may end up being quite difficult to do.

koetsier (changes requested):
This is turning out to be a bit more difficult to do than previously thought. I think we should rethink on how to approach this logic - maybe we can pair on this on Monday if you like.

spec/requests/memberships/update_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:first_team_member) { create(:user, organisations: user.organisations) }
+  let(:membership) { first_team_member.memberships.first }
+
+  before do
+    https!
+    sign_in_user(user)
+  end
+  context "when the membership has been confirmed, the user is an administrator and there is one admin user" do
+    it "redirects" do
+      membership.update!(confirmed_at: Time.zone.now)
```
koetsier, line 13: this can be replaced by membership.confirm!

spec/features/team/remove_a_team_member_spec.rb
```diff
@@ -13,21 +13,41 @@
       click_on "Remove user from GovWifi admin"
     end
 
-    it "shows the remove a team member confirmation box" do
-      expect(page).to have_content("Are you sure you want to remove")
-    end
+    context "when there are more than three admin users in the organisation" do
+      let!(:second_user) { create(:user, :super_admin, organisations: [organisation]) }
```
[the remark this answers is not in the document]
koetsier: The users probably should not be super admins

koetsier (changes requested):
Some suggestions to improve the tests. Maybe the tests can also be made a bit more explicit by setting various attributes explicitly without relying on defaults which may make them a bit easier to read.

PR #1892, opened 2023-08-02
Add Link to download admin user data

koetsier:
A link was added to download a CSV file containing data of all the admin users in the system.

It records the email, the name, current sign in date. last sign in date, the number of sign ins and the organisation name

If there are multiple organisations for a user, there will be duplicate rows, with only the organisation name being different

image

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?selectedIssue=GW-953


A1Robson (approved):
lgtm

PR #1901, opened 2023-08-16
Gw 626 content design the 2 admin flow for new organisation

iram-shehzadi:
### What
Introduce a new registration journey page that enables users to invite an additional admin member to their organisation. This enhancement also involves implementing error messages and restrictions on the IPs page, ensuring these measures remain in effect until a new admin is successfully added.

### Why
The motivation behind this development stems from the current limitation wherein some organisations are restricted to a single admin. This constraint poses significant challenges both for the affected organisations and the broader GovWifi initiative.

For instance:

Mitigating 2FA device challenges: Addressing situations where an admin loses access to their two-factor authentication (2FA) device is often a protracted process. Streamlining this procedure is vital for maintaining efficient admin operations.

Minimising impact of admin leaving: In scenarios where the sole admin leaves, the absence of a designated contact point can paralyse GovWifi operations, particularly during critical processes like certificate rotation.

https://technologyprogramme.atlassian.net/browse/GW-626

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -19,11 +19,10 @@ class Organisation < ApplicationRecord
       .order(sort_column => sort_direction)
   }
 
-
   def meets_invited_admin_user_minimum?
     memberships.count(&:administrator?) >= 2
   end
-    
+
   def meets_admin_user_minimum?
     memberships.count { |membership| membership.administrator? && membership.confirmed? } > 2
   end


A1Robson:
So after some testing, it seems like it's still possible to add a location when there is only 1 administrator. This doesn't seem to be restricted in the same way as it is for an IP ?

app/views/users/confirmations/invite_admin.html.erb
```diff
@@ -0,0 +1,39 @@
+<% content_for :page_title, "Invite a team member to be an administrator" %>
+
+<%= form_with model: resource, as: resource_name, url: invitation_path(resource_name) do |f| %>
+  <% if @errors %>
```
koetsier, line 4: I don't think the `@errors` check is necessary if you instantiate a new User model in the invite_admin method.

app/views/users/confirmations/invite_admin.html.erb
```diff
@@ -0,0 +1,39 @@
...
+          <li>add or remove team members</li>
+        </ul>
+      </p>
+
+      <h1 class="govuk-heading-l">Invite a team member</h1>
+      <div class="govuk-grid-column-full govuk-!-padding-left-0">
+        <%= f.fields_for :user do |user_fields| %>
+          <%= user_fields.govuk_email_field :email, label: { text: "Email address of team member" } %>
+        <% end %>
+        <%= hidden_field_tag :permission_level, "administrator" %>
```
koetsier, line 31: Maybe better use the form model for the hidden field tags; ` <%= f.hidden_field :permission_level, value: "administrator" %>`

app/views/users/confirmations/invite_admin.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <ul class="govuk-list govuk-list--bullet">
+          <li>view locations and IPs, team members, and logs</li>
+          <li>manage locations and IPs</li>
+          <li>add or remove team members</li>
+        </ul>
+      </p>
+
+      <h1 class="govuk-heading-l">Invite a team member</h1>
+      <div class="govuk-grid-column-full govuk-!-padding-left-0">
+        <%= f.fields_for :user do |user_fields| %>
```
koetsier, line 28: I don't think `fields_for` is necessary here. The resource should be a 'User' model

spec/features/locations/bulk_upload_spec.rb
```diff
@@ -1,14 +1,28 @@
 describe "Bulk upload locations and IPs", type: :feature do
   let(:user) { create(:user, :with_organisation) }
   let(:organisation) { user.organisations.first }
+  let!(:another_administrator) { create(:user, organisations: [user.organisations.first]) }
+
+  context "when there is only one administrator for the organisation" do
+    before do
+      user.organisations.first.users.delete(another_administrator)
```
koetsier, line 8: Would it not be better to not create the 'another_administrator'  in the first place?

koetsier (changes requested):
Some changes. Let me know what you think

app/controllers/users/invitations_controller.rb
```diff
@@ -6,7 +6,7 @@ class Users::InvitationsController < Devise::InvitationsController
   def create
     if user_is_invalid? || user_belongs_to_current_organisation?
       self.resource = invite_resource
-      render :new
+      render(params[:user][:source] == "invite_admin" ? :invite_second_admin : :new)
```
koetsier, line 9: I don't think there's a test for this yet.

app/views/layouts/application.html.erb
```diff
@@ -46,7 +46,7 @@
     <%= render "shared/header" %>
 
     <div class="govuk-width-container">
-      <% if user_signed_in? %>
+      <% if user_signed_in? && request.path != "/users/invitations/invite_second_admin" %>
```
koetsier, line 49: Not too sure about this comparison with a specific URL. We could have a method like 'show_navigation_bars' or whatever in the application_controller and overload it as necessary in the right controller

spec/features/ips/add_ip_to_existing_location_spec.rb
```diff
@@ -1,6 +1,19 @@
 describe "Adding an IP to an existing location", type: :feature do
   let(:user) { create(:user, :with_organisation) }
   let(:location) { create(:location, organisation: user.organisations.first) }
+  let!(:another_administrator) { create(:user, organisations: [user.organisations.first]) }
+
+  context "when there is only one administrator for the organisation" do
+    before do
+      user.organisations.first.users.delete(another_administrator)
```
koetsier, line 8: Might it be better not to add  the administrator in the first place?

koetsier (changes requested):

app/controllers/application_controller.rb
```diff
@@ -37,6 +38,10 @@ def redirect_user_with_no_organisation
 
 protected
 
+  def show_navigation_bars
```
koetsier, line 41: I like this better, but there is this bit of logic in there that is very local to one controller. I think that, if you didn't know this was here, figuring out why the layout changes for this particular action would be really tricky.

I was thinking you could change this method back to:
```
def show_navigation_bars
  user_signed_in?
end
```
and then in the invitations controller you can overload this method:
```
  def show_navigation_bars
    return false if action_name=="invite_second_action"
    super
  end
```
That way, when you're looking at the `invite_second_admin` action, you can easily see why the navigation bars don't show.

What do you think?

koetsier (approved):
One comment 👍  

Can you sort out the commits before merging?

PR #1913, opened 2023-08-29
Install and run prettier linting

A1Robson:
### What
Add the Prettier package into the system

### Why
So that the style formatting which is no longer done by stylelint is now handled by prettier. 

https://technologyprogramme.atlassian.net/browse/GW-996


koetsier (approved):
lgtm

PR #1934, opened 2023-10-09
Enable EmailGatewayStub use in development env

A1Robson:
### What

Created a stub implementation of EmailGateway which just prints out info and does not use notify or send actual email

### Why

db migration was attempting to use notify and failing since dev notify was not fully configured. Easier to not need notify in some localdev circumstances

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -16,7 +16,6 @@ require_relative "../lib/gateways/email_gateway"
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-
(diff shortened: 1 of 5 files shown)


koetsier (approved):
Looks good. Do you think it would be useful to do for the Test environment as well?

A1Robson:
So there are some tests which assume there is a Notifications::Client embedded within the email gateway. As a later step we can probably rejig the structure so that we always inject our own Notification interface / abstract class and then we'd be able to use the stub email gateway in test too.

PR #1939, opened 2023-10-13
Fix permission level selection in edit form

iram-shehzadi:
### What

Fix permission level selection in edit form

### Why


koetsier (approved):
👍

PR #1941, opened 2023-10-16
format invite admin page

iram-shehzadi:
### What

Format invite admin page 

### Why


koetsier (approved):
looks good

PR #1964, opened 2023-12-21
add functionality to view CBA enrolled user details in admin logs

iram-shehzadi:
### What

Functionality to view CBA in logs with most minimal of changes as user research has not been done on this prototype. We need to do further testing in January to see how people are currently using the logs and what their processes are

### Why

We are introducing CBA but there is not a way to view the logs and it will help admins see any people logging in and identify them as a CBA authenticator, rather than it being blank

https://technologyprogramme.atlassian.net/browse/GW-1267


koetsier (changes requested):
Looks good 👍 

I think it would be easier though to split this into two PRs, one for the CBA flag and one for the logging feature. That'll make it easier to remove the CBA flag at a later date too.

PR #2010, opened 2024-03-05
Cba_gui_changes

A1Robson:
### What

Admin site changes to deal with Certificates

https://technologyprogramme.atlassian.net/browse/GW-1299


app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,81 @@
+class CertificatesController < ApplicationController
+  before_action :authorise_manage_certificates
+  before_action :authorise_manage_current_certificate, only: %i[show edit destroy]
+
+  def index
+    @pagy, @certificates = pagy(Certificate.where(organisation: current_organisation))
```
koetsier, line 6: Is pagy really necessary here? I don't think there'll be that many certificates right?

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    if uploaded_file
+      flash.clear
+      raw_cert = uploaded_file.read
+      mime_type = Marcel::MimeType.for(uploaded_file)
+
+      if mime_type.to_s == "application/x-x509-ca-cert"
+        @certificate = Certificate.new(organisation: current_organisation, name: cert_name)
+        begin
+          @certificate.import_from_x509_content(raw_cert)
+        rescue RuntimeError => e
```
koetsier, line 42: Maybe better to catch all errors at the end of this method?

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+        end
+      else
+        flash[:alert] = "Unsupported file type. Certificate should be a pem."
+      end
+    else
+      flash[:alert] = "No Certificate file selected. Please choose a file to try again."
+    end
+
+    begin
+      saved = @certificate&.save
```
koetsier, line 54: if you catch errors at the end of the method, you don't need to check if `@certificate` is `nil`. I also think you need save! (with bang) to make it throw exceptions.

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    rescue ActiveRecord::RecordNotUnique
+      flash[:alert] = "A certificate with the same serial number already exists!"
+    rescue ActiveRecord::ActiveRecordError => e
+      flash[:alert] = e.message
+    end
+
+    if saved
+      Services.certificate_repository.store_certificate(current_organisation.id, @certificate.name, raw_cert, @certificate.is_root_cert)
```
koetsier, line 62: What if this fails? Could combine this with the save in a transaction so that there is not inconsistency between the db and the s3 bucket

app/views/certificates/index.html.erb
```diff
@@ -0,0 +1,69 @@
...
+            </tr>
+          </thead>
+          <tbody class="govuk-table__body">
+            <% @certificates.each do |certificate| %>
+              <tr class="govuk-table__row">
+                <td class="govuk-table__cell govuk-!-width-one-quarter">
+                  <%= link_to certificate.name, certificate_path(certificate), class: "govuk-link govuk-link--no-visited-state" %>
+                </td>
+                <td class="govuk-table__cell">
+                  <%= certificate.has_oversize_serial_number ? "#{certificate.get_truncated_serial_number}..." : certificate.serial_number %>
```
koetsier, line 44: could make this line a method in certificate or create a helper

config/application.rb
```diff
@@ -10,8 +10,8 @@
 require "action_view/railtie"
 require "sprockets/railtie"
 require "active_storage/engine"
-# require "rails/test_unit/railtie"
 require_relative "../lib/gateways/email_gateway"
+require_relative "../lib/gateways/certificate_repository_s3"
```
koetsier, line 14: is this needed? Does the autoloader not load it?

config/application.rb
```diff
@@ -49,5 +49,6 @@ class Application < Rails::Application
     config.active_model.i18n_customize_full_message = true
 
     config.email_gateway = Gateways::EmailGateway
+    config.certificate_repository = Gateways::CertificateRepositoryS3
```
koetsier, line 52: Is this necessary? It doesn't seem to be overloaded in any of the other environments

lib/gateways/certificate_repository_s3.rb
```diff
@@ -0,0 +1,46 @@
...
+      @s3_trusted_cert_path = s3_trusted_cert_path
+      @s3_root_cert_path = s3_root_cert_path
+      @client = Aws::S3::Client.new(config)
+    end
+
+    def store_certificate(organisation_id, cert_name, raw_data, is_root_cert)
+      organisation_cert_name = construct_organisation_cert_name(organisation_id, cert_name)
+      key = construct_s3_cert_path(organisation_cert_name, is_root_cert)
+
+      @client.put_object(body: StringIO.new(raw_data), bucket: @s3_cert_bucket, key:)
```
koetsier, line 14: Do we not need to upload this to two buckets? One in Ireland and one in London

app/controllers/certificates_controller.rb
A1Robson: yeah that would be great - didn't know that was a possibility in rails.

app/controllers/certificates_controller.rb
A1Robson: i guess not - i think there will end up being a few iterations of the gui, so can probably stay until next time...

config/application.rb
A1Robson: was in there for the email gateway already, so assumed needed

config/application.rb
A1Robson: Might be nice to swap it out for another implementation in a similar way to the email gateway - but the whole s3 repo could potentially go, so maybe not needed..

app/views/certificates/index.html.erb
A1Robson: yup would make sense

app/models/certificate.rb
```diff
@@ -0,0 +1,55 @@
...
+
+class Certificate < ApplicationRecord
+  belongs_to :organisation
+  validates :name, presence: true, uniqueness: { case_sensitive: false, scope: :organisation_id }
+  validates :thumbprint, presence: true, uniqueness: { scope: :organisation_id }
+  validates :subject, :issuer, :valid_from, :valid_to, :serial_number, :content, presence: true
+
+  MAX_LEN = 16
+
+  def import_from_x509_file(file_path)
```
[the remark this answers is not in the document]
A1Robson: i guess, but functionality seems to sit comfortably with the class...

app/controllers/certificates_controller.rb
A1Robson: took another look and found a much neater way, thx.

app/controllers/certificates_controller.rb
A1Robson: done

app/controllers/certificates_controller.rb
A1Robson: left this one for you :)

PR #2037, opened 2024-05-07
Ensure uploaded/deleted certificate have valid chains

koetsier:
### What

When uploading or deleting certificate, all known certificates should have a valid chain to a self signed root certificate.
Uploading a root certificate is always allowed
When uploading an intermediate certificate, ensure an issuing certificate is present.
When deleting a certificate, ensure there are no known issued certificates.

### Why

To make sure we have all certificates in place to be able to verify a leaf certiicate.

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)
](https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251?assignee=557058%3Add853f03-c9ea-4b86-a72e-1f14eb7f1aaf&selectedIssue=GW-1608)

image

image


app/controllers/certificates_controller.rb
```diff
@@ -17,8 +17,12 @@ def edit
 
   def destroy
     @certificate = Certificate.find(params[:id])
-    @certificate.destroy!
-    redirect_to(certificates_path, notice: "Successfully removed Certificate: #{@certificate.name}")
+    if @certificate.has_child?
+      redirect_to(certificates_path, alert: "Cannot remove a certificate with issued certificates. Please remove the issued certificates first.")
```
leonaAtkins, line 21: Is this to do with intermediates or roots, maybe more clearer wording like, "unable to remove intermediate certificate without removing the client first" or something like that.

koetsier: it depends, it can be any certificate that has direct children, so either a root or any intermediate. In x509 speak, people seem to not use 'child' certificate so much and 'issuing' seems to be the word? Happy to change it to something else.

leonaAtkins: Ok, I understand, ya that's tricky if there isn't standard thermology,

config/locales/en.yml
```diff
@@ -41,6 +41,7 @@ en:
               taken: "This certificate has already been uploaded"
               blank: "No Certificate file selected. Please choose a file to try again."
               invalid_certificate: "Certificates must be in a .PEM format"
+              no_parent: "The certificate does not have a known issuing certificate. Please add its issuing certificate first."
```
leonaAtkins, line 44: To be consistant, you may want to move some of the other messages, like those in certificate_controller into there?

leonaAtkins (approved):

config/locales/en.yml
koetsier: 👍

PR #2041, opened 2024-05-14
Add a 'serve-no2fa' option to the Makefile

koetsier:
This allows a developer to start the server without having to worry about using 2FA to test a feature


leonaAtkins:
This is deniably needed, just want to check, it can only be applied on dev/local, so someone couldn't invoke this on other envs?

leonaAtkins (approved):

PR #2051, opened 2024-05-29
Export certificates

koetsier:
### What

Add a rake task to export certificates to an S3 bucket
### Why

So that the FreeRadius server can pick them up during the safe restart and start accepting valid EAP-TLS connection.

https://technologyprogramme.atlassian.net/browse/GW-1683


spec/use_cases/certificate_exporter_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    intermediate_ca = create(:certificate, issuing_key: root_key, issuing_subject: "/CN=rootca", organisation:)
+
+    UseCases::CertificateExporter.export
+
+    Zip::File.open(certficate_file_path) do |files|
+      expect(files.count).to eq(1)
+      expect(files.read(files.first)).to eq(intermediate_ca.content)
+    end
+  end
+end
```
leonaAtkins, line 35: It think it would be good to have a test to handle s3 (AWS) exceptions, you can mock the s3 resource and make it return an exception, which is expected and caught, something like.
https://stackoverflow.com/questions/26665520/rspec-how-to-stub-a-3rd-party-exception

PR #2054, opened 2024-06-03
Gw 816 make some accessibility changes to the admin tool

iram-shehzadi:
### What

There are some accessibility things we must improve on our website

Page titles aren’t all correct - including upload multiple locations, switch organisation, add IP address, admin details (super admin), user details (super admin)

Change admin permissions (both invite new and edit) - with a screen reader, the 3 hint text lines below each permission line get merged together. We could add <span> to break it up, add a full stop to break them into 3 sentences or maybe something else…

Give context to ‘view logs' link on /IPS page - the ‘view logs’ link doesn’t have any context as a screen reader for which logs I could look at. If you look at the ‘remove’ link it does. We just need to do similar for view logs. Otherwise as a screen reader user, I can’t tell what the view logs is for. Everything else on the page appears to do this fully.

### Why

WCAG updates from 2.1 to 2.2

https://technologyprogramme.atlassian.net/browse/GW-816

### DIFF
--- a/app/controllers/memberships_controller.rb
+++ b/app/controllers/memberships_controller.rb
@@ -10,29 +10,29 @@ class MembershipsController < ApplicationController
         value: "administrator",
         text: "Administrator",
         hint: <<~HINT.html_safe,
-          View locations and IPs, team members, and logs<br>
-          Manage locations and IPs<br>
-          Add or remove team members<br>
-          View, add and remove certificates
+          <span>View locations and IPs, team members, and logs</span><br>
+          <span>Manage locations and IPs</span><br>
+          <span>Add or remove team members</span><br>
+          <span>View, add and remove certificates</span>
         HINT
       ),
       OpenStruct.new(
         value: "manage_locations",
         text: "Manage Locations",
         hint: <<~HINT.html_safe,
-          View locations and IPs, team members, and logs<br>
-          Manage locations and IPs<br>
-          Cannot add or remove team members<br>
-          View, add and remove certificates
+          <span>View locations and IPs, team members, and logs</span><br>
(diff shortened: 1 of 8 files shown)


leonaAtkins (commented):
On the Invite Team member page, at the Permission level section, I am able to navigate using the down arrow, read the page and select radio options, however if I wanted to navigate via the headings, there are none, (see attached screenshot), which doesn't seem to correspond to the page content (think it's picking up the footer headings).
Also instead of using spans, a full stop (.) would have been enough to make the screen reader pause and read lines by lines. (I tried this in developer tools just to make sure :) ).
On the organizations page, listing IP's and Show logs, when I navigate to show logs I hear.
'lists of 1 items,  213.186.185.17, view logs for 213.186.185.17  Momentum Centre, London, SE1 0SX, link, row 3 col 11, which could be too verbose, maybe only the IP is needed?  It also shows a large red box (I think ZoomText is picking up the span tag), maybe aria-label would be better here). I was trying to get a screen shot of that but it wouldn't let me.
Also if there is no logs, it just says 'blank', would it be better to say 'no logs available' or something?
image
The search box on user and admin pages, are not doing what I think is expected, (it's only reading out the label, not the description),  here is an example of how it should be laid out by W3C, the hint text comes after the input form, with 'approach 2' being the most compatible with assistive technologies. 
https://www.w3.org/WAI/tutorials/forms/instructions/
Hope this is ok, happy to talk about any aspects or provide any assistance or additional testing.
For Reference I'm using Fusion (JAWS + ZoomText) in Chrome on Windows 11.

leonaAtkins (approved):

PR #2069, opened 2024-06-24
Decouple the ruby application running state from the GovWifi environment

koetsier:
### What
The ruby application running state needs to be decoupled from the GovWifi environment it’s running in.

### Why
Leaving things as they are means that spinning new GovWifi environments from scratch take extra time, and this delays the ability to bring the service back online in the event of a serious incident.

See: https://github.com/alphagov/govwifi-terraform/pull/877

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251


config/environments/production.rb
```diff
@@ -76,7 +76,7 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
-  config.action_mailer.default_url_options = { host: "admin.wifi.service.gov.uk" }
+  config.action_mailer.default_url_options = { host: ENV["FQDN"] }
```
leonaAtkins, line 79: With reference to the PR on the Terraform PR, if we could use more readable names it would help document the code.
```suggestion
  config.action_mailer.default_url_options = { host: ENV["FULL_DOMAIN_NAME"] }
```

koetsier: fixed it to FULLY_QUALIFIED_DOMAIN_NAME. hope that is clearer. (it should include the hostname as well)

PR #2072, opened 2024-07-01
notify users when user removed

iram-shehzadi:
### What

When a superadmin deletes an account, the user is informed

### Why

So that superadmins don’t have to manually write out a message to users when removing them

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)](https://technologyprogramme.atlassian.net/browse/GW-1597)

### DIFF
--- a/app/controllers/super_admin/wifi_user_searches_controller.rb
+++ b/app/controllers/super_admin/wifi_user_searches_controller.rb
@@ -36,13 +36,13 @@ private
   end
 
   def notify_user
-    if @wifi_user.contact.include?("@")
-      AuthenticationMailer.notify_user_account_removed(
+    if @wifi_user.mobile?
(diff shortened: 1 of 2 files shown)


app/controllers/super_admin/wifi_user_searches_controller.rb
```diff
@@ -33,4 +34,18 @@ def destroy
   def search_term_params
     params.fetch(:search_form, {}).permit(:search_term)
   end
+
+  def notify_user
+    if @wifi_user.contact.include?("@")
```
koetsier, line 39: could move this to the model

koetsier, line 39: Could move this logic to the model, so you can do `@wifiuser.email?` or similar

koetsier (approved):
👍 one suggestion

PR #2073, opened 2024-07-01
Rename CBA button

koetsier:
The button to enable/disable CBA for an organisation used to be used to enable/disable logging for CBA only, but is now used to enable all CBA functionality and so the button is renamed accordingly

image
image


seb-szy (approved):
LGTM

seb-szy (approved):
LGTM

PR #2077, opened 2024-07-08
set up sms gateway for notify

iram-shehzadi:
### What

Set up a notify sms gateway so that we are able to send sms messages to users

### Why

We need to notify wifi users when their account has been removed, currently we only have and email gateway set up.

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)](https://technologyprogramme.atlassian.net/browse/GW-1776)


koetsier (approved):
looks good

PR #2140, opened 2024-10-15
Accessibility Audit - Incorrect Attributes

mel-am:
Scope removal and role has been updated.

### What
Incorrect Attributes

### Why

Attributes have been used incorrectly.
This affects screen reader users, as the intended purpose, to mark cells in the 'Username' column as being table headers for the row, which would then inform screen reading software to announce both the row and column table headers to screen reader users whilst they navigate the table, is not being achieved, due to the invalid application of the attribute.

[GW-1915](https://technologyprogramme.atlassian.net/browse/GW-1915)

[GW-1915]: https://technologyprogramme.atlassian.net/browse/GW-1915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


koetsier (approved):
🥇

PR #2145, opened 2024-10-21
GW-1917 row context added

mel-am:
### What

Table header was not present.

### Why

On the 'Organisations allow list' page, there is a table of registered organisations which includes a 'Remove' action button in the third column for each organisation row. However, this column does not have a table header to introduce the content within it. These links lack DAC | Accessibility Report 48 sufficient context to differentiate which 'Remove' button removes which organisation, when screen reader users are browsing out of context. This affects screen reader users, as they cannot tell from the link text alone what the purpose or destination of the link might be. It is not clear which action applies to which organisation when browsing out of context.

[GW-1917](https://technologyprogramme.atlassian.net/browse/GW-1917)

[GW-1917]: https://technologyprogramme.atlassian.net/browse/GW-1917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


leonaAtkins (commented):
While I can confirm I now hear 'remove test' (I set up a test org), I wonder if it should be a little more verbose and say 'remove test org from allow list', as those without sight don't have the visual context.

mel-am:
@koetsier- Please can you advise on the comment from Leona, should the message be amended? _While I can confirm I now hear 'remove test' (I set up a test org), I wonder if it should be a little more verbose and say 'remove test org from allow list', as those without sight don't have the visual context._

leonaAtkins (approved):
Works perfectly, well done.

PR #2152, opened 2024-10-29
GW-1913- Accessibility audit radio buttons

mel-am:
### What
Ensure that inputs have a unique and descriptive programmatically associated label.
Further, as detailed by GOV.UK Design System: Radios (How it works), related radio buttons
should be grouped within a

element and be provided with a descriptive
(which should be the first direct child of the , as defined by the HTML standard, otherwise it might not be correctly picked up by assistive technology).

### Why

On the 'Memorandum of understanding' page, there are radio buttons that, due to a lack of
programmatic association between the label and input, lack an accessible name.
This affects screen reader users, as the radio buttons are not labelled and, therefore, their
purpose is not programmatically determinable, and it also affects voice activation users, as
there is no programmatic association between the visible label and the input for the user to
select the input via its visible label ('click I will sign the MOU').

[GW-1913](https://technologyprogramme.atlassian.net/browse/GW-1913)

[GW-1913]: https://technologyprogramme.atlassian.net/browse/GW-1913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/mous/show_options.html.erb
+++ b/app/views/mous/show_options.html.erb
@@ -12,11 +12,6 @@
   <p class="govuk-body">Anyone can view or share the <%= link_to "GovWifi MOU (opens in another tab).", "https://www.wifi.service.gov.uk/memorandum-of-understanding/", target: "_blank", class: "govuk-body govuk-link govuk-link--no-visited-state", rel: "noopener" %>
   </p>
 
-  <fieldset class="govuk-fieldset" aria-describedby="sign-mou-hint">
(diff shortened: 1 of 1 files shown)


leonaAtkins:
@mel-am  are you aware this build has failed the linter?

leonaAtkins (approved):
Tested by tabbing through the page, when landing on the radio button I now hear "Who will sign the MOU", before this was not announced, fixed, well done.

PR #2162, opened 2024-11-05
Allow signing out while setting up 2FA

koetsier:
Previously, it was impossible to sign out while setting up 2FA clicking the sign out link, would take the user back to the 2FA screen.

Now, signing out takes the user back to the sign in page while setting up 2FA

https://technologyprogramme.atlassian.net/jira/software/projects/GW/boards/251


leonaAtkins (approved):
LGTM.

PR #2168, opened 2024-11-08
Fix unconfirmed memberships

koetsier:
### What
Ensure only confirmed organisation memberships can be used

### Why
Unconfirmed organisations should not be shown in the UI and
should not be able to be modified

### DIFF
--- a/spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
+++ b/spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
@@ -45,7 +45,7 @@ describe "Lookup wifi user contact details", type: :feature do
   end
 
   context "with a phone number search term" do
-    let(:search_term) { "+44 7891 234567" }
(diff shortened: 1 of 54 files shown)


spec/features/super_admin/lookup_wifi_user_contact_details_spec.rb
```diff
@@ -45,7 +45,7 @@
   end
 
   context "with a phone number search term" do
-    let(:search_term) { "+44 7891 234567" }
+    let(:search_term) { "+447891234567" }
```
leonaAtkins, line 48: Should it still test phone numbers with spaces, to catch that scenario, ie ensure spaces are ignored/removed ?

koetsier: It won't handle spaces in phone numbers but I don't think it matters. This way it _does_ handle names with spaces in it

PR #2179, opened 2024-11-19
GW-1924 - Headings and labels - Update confirm rotate radius key

mel-am:
### What

Button text was insufficiently descriptive of their topic or purpose, whilst browsing out of
context.

### Why

On the 'Rotate secret key' page, neither the page title, error summary nor 'Yes, rotate this
RADIUS key' button identify which location the secret key is to be rotated for.
This affects all users, but the loss of context as to which location the secret key is to be
rotated, the context from the previous step, can particularly affect cognitive and screen
reader users by relying on their memory to remember which location the pronoun 'this' is
referring to.
[GW-1924
Screenshot 2024-11-19 at 12 53 06
](https://technologyprogramme.atlassian.net/browse/GW-xxxx)

### DIFF
--- a/app/views/ips/_confirm_rotate_radius_key.html.erb
+++ b/app/views/ips/_confirm_rotate_radius_key.html.erb
@@ -2,7 +2,7 @@
 
 <div class="govuk-error-summary" role="alert">
   <h2 class="govuk-error-summary__title">
-    Are you sure you want to rotate this RADIUS secret key?
+    Are you sure you want to rotate this RADIUS secret key for <%= @key_to_rotate.address %>?
   </h2>
   <div class="govuk-error-summary__body">
     <table class="govuk-table">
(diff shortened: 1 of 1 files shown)


app/views/ips/_confirm_rotate_radius_key.html.erb
```diff
@@ -17,5 +17,8 @@
 
     <%= button_to "Yes, rotate this RADIUS key", location_path(@key_to_rotate),
```
leonaAtkins, line 18: Hi Mel, this doesn't read continuously, what I hear is "Yes, rotate this key", then it stops, if I press down I then here "for <address>", feels a bit jarring, and this is because the 'span' is written outside of the button, and indeed form, which you can see if you look at the html output. 
I don't think 'button_to' has any aria features, so you may have to 'hand craft' the form and button like.
```
  <%= form_with(url: @key_to_rotate, method: :patch) do |form| %>
    <div>
      <button class="govuk-button red-button" type="submit" aria-label="Yes, rotate  RADIUS key for <%= @key_to_rotate.address %>"> Yes, rotate this RADIUS key</button>
    </div>
  <% end %>
```

leonaAtkins (approved):

mel-am:
Screenshot 2024-12-16 at 16 27 38

PR #2215, opened 2025-01-27
Remove Devise rememberable

koetsier:
The rememberable module in Devise allows users to stay signed in across browser sessions by setting a persistent remember_me cookie. This cookie securely stores a token linked to the user’s account, enabling automatic authentication when the user revisits the app, unless the cookie is manually cleared or expired.

We currently don't use this functionality and it can therefore be removed


mel-am (approved):
lgtm

PR #2221, opened 2025-02-04
Log Zendesk errors in Sentry

koetsier:
We have had a few Zendesk tickets informing us that Zendesk sometimes does not accept a support ticket sent from the Admin Portal

We need to figure out why and how often this happens so we can do something about it

https://technologyprogramme.atlassian.net/browse/GW-2075


mel-am (approved):
lgtm

PR #2228, opened 2025-02-11
Ensure that error pages render the correct error page

koetsier:
When entering an incorrect route in the admin portal, it returns a 500 error code with a message: 

```
500 Internal Server Error
If you are the administrator of this website, then please read this web application's log file and/or the web server's log file to find out what went wrong. 
```

This is an incorrect message and should instead be a message saying that the page does not exist with a 404 error code

https://technologyprogramme.atlassian.net/browse/GW-2100

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -49,6 +49,7 @@ class ApplicationController < ActionController::Base
   end
 
   def redirect_user_with_no_organisation
+    return if current_user.nil?
     return if current_user.is_super_admin? && current_organisation.present?
 
     if current_user.is_super_admin? && current_organisation.nil?


koetsier:
image

app/controllers/application_controller.rb
```diff
@@ -49,6 +49,7 @@ def super_admin?
   end
 
   def redirect_user_with_no_organisation
+    return if current_user.nil?
```
leonaAtkins, line 52: Could be a line liner? (untested)
```suggestion
    return if current_user.nil? || current_user.is_super_admin? && current_organisation.present?
```

koetsier: Could do, but I'm not sure if it would make the code easier to understand

leonaAtkins: the 2 returns could also be confusing.  No worries, all good.

leonaAtkins (approved):

PR #2302, opened 2025-04-23
Use new 2fa gem

koetsier:
### What

Use the new 2fa gem (https://github.com/GovWifi/govwifi_two_factor_auth)

### Why
The old 2fa gem (https://github.com/Houdini/two_factor_authentication) is no longer maintained and is not compatible with Rails 8. We decided to fork the gem and bring it up to date.

https://technologyprogramme.atlassian.net/browse/GW-2218


Gemfile.lock
```diff
@@ -1,3 +1,14 @@
+GIT
```
leonaAtkins, line 1: while I recognise this isn't my area of expertise, is this the best way to integrate this gem? Is there no gem repo we could submit it too, or could we pull and build in the docker build using [gemstash](https://guides.rubygems.org/run-your-own-gem-server/)?   Just concerned how we maintain this going forward

koetsier: We could use rubygems to upload our gem, but that requires an account that we can all administer and we don't have one. This solution just gets the gem straight from the source in git and doesn't require any extra configuration.

ripan:
@koetsier LGTM 👍

leonaAtkins (approved):

PR #2305, opened 2025-04-28
review sort link functionality

iram-shehzadi:
### What

Clean up the code.

Remove CSS tags as they are not used.

Add test to ensure the sort link functionality works as expected.

Ensure the sort direction (ascending/descending) toggles correctly when clicking column headers.

### Why

Previously the sort_link method added CSS classes to table headers but did not seem to properly handle the sorting behaviour. This clean up ensures that sortable columns function correctly  and are tested.

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)
](https://technologyprogramme.atlassian.net/browse/GW-1364)

### DIFF
--- a/app/controllers/super_admin/organisations_controller.rb
+++ b/app/controllers/super_admin/organisations_controller.rb
@@ -39,7 +39,7 @@ private
   end
 
   def sortable_columns
-    %w[name created_at locations_count ips_count mous.created_at last_sign_in_at email sign_in_count certificates_count]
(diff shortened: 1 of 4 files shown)


koetsier (approved):
I think the spec is a duplicate: https://github.com/GovWifi/govwifi-admin/blob/master/spec/features/super_admin/organisations/sort_organisation_list_spec.rb. otherwise 👍

PR #2306, opened 2025-04-28
Update location

mel-am:
### What

Describe the change

### Why


mel-am:
Screenshot 2025-04-30 at 10 51 49

app/views/locations/edit.erb
```diff
@@ -10,6 +10,9 @@
       <%= f.govuk_text_field :address,
                              hint: { text: "For example, 10 High Street, London" } %>
       <%= f.govuk_text_field :postcode %>
+      <%= f.govuk_collection_select :organisation_id, Organisation.all, :id, :name,
```
koetsier, line 13: I think the choice should be limited to all orgs the user is a member of, instead of all orgs

spec/features/locations/update_location_spec.rb
```diff
@@ -1,7 +1,8 @@
 describe "Update location", type: :feature do
-  let(:organisation) { create(:organisation, locations: [location]) }
-  let(:user) { create(:user, :confirm_all_memberships, organisations: [organisation]) }
   let(:location) { create(:location) }
+  let(:organisation) { location.organisation }
```
koetsier, line 3: looks good, but maybe add a request spec, so we know that a user can't set the location to an org they're not a member of

app/views/locations/edit.erb
koetsier, line 13: This should probably be limited to all orgs the user is a member of and not all orgs. This should also be enforced in the controller

koetsier: maybe make an exception for a super admin?

koetsier (changes requested):
I like the solution, but some comments

spec/controllers/locations_controller_spec.rb
```diff
@@ -78,4 +81,46 @@
       end
     end
   end
+
```
koetsier, line 84: These tests should ideally be request specs. Controller specs are deprecated

spec/controllers/locations_controller_spec.rb
```diff
@@ -78,4 +81,46 @@
...
+      context "when location's organisation doesn't belong to signed_in user organisation" do
+        it "render exit page with flash alert" do
+          put :update_location, params: { location_id: location.id, location: { organisation_id: other_organisation.id, address: location.address, postcode: location.postcode } }
+          expect(flash[:alert]).to match(/You can't update the location to an organisation you're not a member of/)
+          expect(response).not_to redirect_to(ips_path)
+        end
+      end
+    end
+  end
```
koetsier, line 125: There are a number of other constraints we need to take into account when moving locations:

1. The user needs to have the right permissions on both the source and destination organisation.
2. The destination organisation must have at least 2 administrators

app/controllers/locations_controller.rb
```diff
@@ -155,4 +156,14 @@ def authorise_ip_actions
       redirect_to ips_path
     end
   end
+
+  def check_user_organisations
+    unless current_user.is_super_admin?
+      organisation = Organisation.find(params.dig(:location, :organisation_id))
+      unless current_user.organisations.include? organisation
+        flash[:alert] = "You can't update the location to an organisation you're not a member of"
+        render :edit
```
koetsier, line 165: This should redirect because otherwise it won't have a location_id set. Not being a member of the organisation you want to move a location to is also not part of the normal flow so can be handled a bit abruptly

spec/controllers/locations_controller_spec.rb
```diff
@@ -78,4 +81,46 @@
...
+
+  describe "POST /update_location" do
+    context "when sign_in user is super_admin" do
+      before do
+        sign_in super_admin
+      end
+
+      context "when location's organisation belongs to signed_in user organisation" do
+        it "update location and redirect to ips_path " do
+          put :update_location, params: { location_id: location.id, location: { organisation_id: user.organisations.first.id, address: location.address, postcode: location.postcode } }
```
koetsier, line 93: It would be a good idea to check if the location has actually moved after doing this.

koetsier (changes requested):

oisinmulvihill (approved):
That looks good to me.

🆗 🚢

PR #2325, opened 2025-05-08
Change the ip address from a partial to a helper

koetsier:
### What

Change the ip address from a partial to a helper


koetsier:
> @koetsier You fixed a bug here if i am not mistaken 👏
> 
> ```ruby
> ip = "121.22.33.44"
> 
> ip.split(".")[0...-1] # returns ["121", "22", "33"]
> ip.scan(/\d{1,3}/) # returns ["121", "22", "33", "44"]
> ```

It ends with

`ip.split(".").last`

Not very clear I agree

PR #2334, opened 2025-05-13
Use govuk components gem for logs index

koetsier:
### What

Use govuk components gem for logs index

### Why

Using the gem makes writing design system compatible frontend code easier

https://technologyprogramme.atlassian.net/browse/GW-2246


leonaAtkins (approved):
Deployed to alpaca, all worked as expected.

PR #2336, opened 2025-05-14
reset smoke test users

iram-shehzadi:
### What

Create a Rake task to reset the smoke test users details to the expected state, which we can run manually should the smoke tests fail.

### Why

The smoke tests can break if ran concurrently, which can leave the admin site in an inconsistent state and the smoke tests will fail with, invalid user/password, or unable to add / remove user. This happens because one of the steps in the smoke tests change the password of a user and back again.

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)](https://technologyprogramme.atlassian.net/browse/GW-2195)

### DIFF
--- a/lib/tasks/reset_smoke_test_users.rake
+++ b/lib/tasks/reset_smoke_test_users.rake
@@ -1,0 +1,24 @@
+namespace :reset do
+  desc "Reset smoke test users to the passwords stored in AWS"
+  task smoke_test_users: :environment do
+    users = {
+      ENV["GW_USER"] => ENV["GW_PASS"],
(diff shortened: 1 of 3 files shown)


lib/tasks/reset_user_for_smoke_tests.rake
```diff
@@ -0,0 +1,25 @@
+namespace :smoke_test do
+  desc "Reset smoke test users to the passwords defined in environment variables"
+
+  task reset_users: :environment do
```
koetsier, line 4: Could you put the logic in a class and then write a few small tests?

PR #2355, opened 2025-05-27
Use govuk component gem for certificate views

koetsier:
This allows for simpler, more compliant UI code


leonaAtkins (approved):
Tested on Alpaca, LGTM

PR #2361, opened 2025-05-30
refactor filter to search for organisations

iram-shehzadi:
### What

Updated the JavaScript used for filtering the organisations table to work with the GOV.UK Design System table component. Replaced class based filtering logic with generic table row and cell checks.

### Why

The previous filter stopped working after switching to the GOV.UK table component, which doesn't use the .result-row or .filter-by classes the old script depended on.

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)](https://technologyprogramme.atlassian.net/browse/GW-2277)

### DIFF
--- a/app/views/shared/_filter_search.html.erb
+++ b/app/views/shared/_filter_search.html.erb
@@ -5,12 +5,13 @@
 
   function filterResults() {
     var filterValue = filterInput.value.toUpperCase();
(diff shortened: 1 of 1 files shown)


szd55gds (approved):
Looks ok to me, merge when ready, test in Staging please

PR #2362, opened 2025-06-02
Update error pages

koetsier:
Add View Component Gem and use ErrorSummary view component

The View Component Gem is used by the GOV.UK components gem so does not really introduce
a new dependency. This gem is very convenient for encapsulating design system
compliant HTML.

Error summary markup is currently duplicated across the project and often deviates from the design system. This component centralises that logic, ensuring consistency and compliance.

It can now be reused via a simple view helper.


leonaAtkins (approved):
Deployed to Dev, tested, LGTM

PR #2368, opened 2025-06-04
Update the footer with the GovUK Component gem

koetsier:
This ensures better compatibility with the design system

Also, the design of the footer will be updated soon, which means that when the gem is updated, out footer will stay up-to-date


leonaAtkins (approved):
Deployed to Alpaca, tested and working, looks good to me.

PR #2375, opened 2025-06-11
add status page link to locations

iram-shehzadi:
### What

Add GovWifi status link to the ‘home’ locations page.

### Why

Admins are looking to check status more often and we don't link clearly to status page.

It will reduce support ticket requests.

[[GW-xxxx](https://technologyprogramme.atlassian.net/browse/GW-xxxx)](https://technologyprogramme.atlassian.net/browse/GW-181)


koetsier (approved):
you could use the new `govwifi_link_to` helper so you don't need the 'govuk-link' class

PR #2395, opened 2025-07-10
remove status page

iram-shehzadi:
### What
Remove /status route, page and code from the admin portal

### Why
This route and page exists, but does not work and has no links to it and so should be removed

(https://technologyprogramme.atlassian.net/browse/GW-2350)


koetsier (approved):
Looks good

PR #2414, opened 2025-08-01
GW-2319 new ruby updater workflow

leonaAtkins:
### What

Add dependency on previous job

### Why

Else the labeler runs in the wrong order.

[GW-2319](https://technologyprogramme.atlassian.net/browse/GW-2319)


leonaAtkins:
for some reason, git has broken this PR, will recreate.

PR #2488, opened 2025-12-18
Upgrade User DB Version To MYSQL 8.4

camdesgov:
### What

Upgrade User DB Version To MYSQL 8.4

### Why

This is to match the version we are now running in production. It relates to: 
https://github.com/GovWifi/govwifi-terraform/pull/983

Remove "--==default-authentication-plugin==" option as it is no longer supported in MYSQL 8.4 and causes an error, when testing locally.

https://technologyprogramme.atlassian.net/browse/GW-2549


oisinmulvihill (approved):
LGTM

🆗 🚢

iram-shehzadi (approved):

PR #2636, opened 2026-05-29
enforce IP location referential integrity with cascading deletes

iram-shehzadi:
### What

Enforce IP location referential integrity with cascading deletes

### Why

Investigation identified that the ips table does not currently enforce referential integrity with the locations table. The table only contains an index on location_id and does not have a foreign key constraint.

As a result, orphaned IP records exist within the database:

2 IP records reference a deleted/non-existent location (location_id = 580)

1 IP record has a NULL location_id

The Rails associations are already configured with:

Ip model: belongs_to :location

Location model: has_many :ips, dependent: :destroy

This indicates that application-level cleanup is already implemented correctly when records are deleted through Rails. However, database-level referential integrity is not enforced, meaning orphaned IP records can still be created when operations bypass Rails (e.g. direct SQL operations, scripts, imports, or use of delete instead of destroy).

Changes to be implemented:

Check whether these IPs are being used (done)

Remove existing orphaned/invalid IP records (staging and prod)

Add a foreign key constraint between ips.location_id and locations.id

Configure ON DELETE CASCADE to automatically remove associated IPs when a location is deleted

Consider making location_id non-nullable if every IP must belong to a location

These changes will ensure referential integrity is enforced at the database level and prevent orphaned IP records from being created in future.

https://technologyprogramme.atlassian.net/browse/GW-2762

### DIFF
--- a/db/migrate/20260529103602_add_foreign_key_to_ips.rb
+++ b/db/migrate/20260529103602_add_foreign_key_to_ips.rb
@@ -1,0 +1,6 @@
+class AddForeignKeyToIps < ActiveRecord::Migration[8.0]
+  def change
+    change_column_null :ips, :location_id, false
+    add_foreign_key :ips, :locations, on_delete: :cascade
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_12_23_164928) do
+ActiveRecord::Schema[8.0].define(version: 2026_05_29_103602) do
   create_table "active_storage_attachments", charset: "utf8mb3", force: :cascade do |t|
     t.string "name", null: false
     t.string "record_type", null: false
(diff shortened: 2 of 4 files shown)


ethancane-digital-co:
There exists a task (assigned to your good self) for this topic in the "Define account health rules" epic:

`Consider making location_id non-nullable if every IP must belong to a location`

https://technologyprogramme.atlassian.net/browse/GW-2762

It might be worth cross-referencing this PR with that ticket to add further context when it comes time to work on it?

Otherwise, all great work!

ethancane-digital-co:
Accidentally closed PR. Apologies.

ethancane-digital-co (approved):
Everything looks good.

oisinmulvihill (approved):
This looks good to me!

🆗🚢

iram-shehzadi:
> There exists a task (assigned to your good self) for this topic in the "Define account health rules" epic:
> 
> `Consider making location_id non-nullable if every IP must belong to a location`
> 
> https://technologyprogramme.atlassian.net/browse/GW-2762
> 
> It might be worth cross-referencing this PR with that ticket to add further context when it comes time to work on it?
> 
> Otherwise, all great work!
This is the work for that

PR #2646, opened 2026-06-12
Gw 2739 how many orgs

ethancane-digital-co:
### What

Return a count of the number of current organisations and publish to S3 and metrics API.

### Why

To record stats and push onwards to Tableau dashboards.

[GW-2739](https://technologyprogramme.atlassian.net/browse/GW-2739)


oisinmulvihill (approved):
I'm not sure why Iram's changes are showing up in here as a difference to master. You might need to resolve that.

The other work looks good to me!

🆗🚢

PR #2650, opened 2026-06-19
bump govuk-components to 6.2.0 and view_component to 4.9.0

iram-shehzadi:
### What
Combined two Dependabot updates:
govuk-components 5.10.0 -> 6.2.0 and view_component 3.21.0 -> 4.9.0

govuk-components 6.2.0 requires view_component >= 4.9, < 4.10, so the two updates are mutually consistent. Verified bundle install resolves cleanly inside the Docker image build.

### Why
The failures came from the dependabot bumps broke in two ways:
1. view_component 4.9.0 - ErrorSummaryComponent#initialize called
bare super, which forwarded title:/list: to ViewComponent: :Base#initialize (no longer accepts arguments - ArgumentError: wrong number of arguments (given 1, expected 0).
- Fixed by changing super -
super () in app/components/error_summary_component. rb: 5
2. govuk-components 6.2.0 - the template called govuk_list as if
it were a component instance method; it's now only available through the view-helper proxy → NoMethodError: undefined method 'govuk_list'.
- Fixed by changing govuk_list - helpers-govuk_list in app/components/error_summary_component.html.erb:8

https://technologyprogramme.atlassian.net/browse/GW-2854


oisinmulvihill (approved):
This looks good to me!

🆗🚢

PR #2677, opened 2026-07-17
How many Organisations have dormant administrators?

oisinmulvihill:
### What

- Adds a new rake task, `metrics:publish_orgs_with_dormant_admins_count`, that counts organisations relying on a single dormant admin
- Publishes the daily count to S3 (`account-health-orgs-with-dormant-admins-count/orgs-with-dormant-admins-count-<date>`) and posts it to the Metrics API, following the same pattern as `publish_organisation_count`
- Includes spec coverage for the counting logic, edge cases, S3 output, and Metrics API publishing

### Why

- Account health reporting needs visibility into organisations at risk of losing admin access, where the only qualifying admin hasn't signed in for over a year
- This mirrors an existing, already-scheduled metric so it slots into the current dashboard/reporting pipeline with no new infrastructure pattern

- [GW-2741](https://technologyprogramme.atlassian.net/browse/GW-2741)

### DIFF
--- a/lib/tasks/publish_orgs_with_dormant_admins_count.rake
+++ b/lib/tasks/publish_orgs_with_dormant_admins_count.rake
@@ -1,0 +1,11 @@
+require "logger"
+logger = Logger.new($stdout)
+
+namespace :metrics do
+  desc "Publish orgs with dormant admins count to S3 and post to metrics API"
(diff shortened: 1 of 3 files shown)


ethancane-digital-co (approved):
LGTM

PR #2682, opened 2026-07-22
How many Organisations have less than two admins?

oisinmulvihill:
### What

- Adds `metrics:publish_orgs_with_less_than_two_admins_count`, following the same S3 + Metrics API publish pattern as `metrics:publish_orgs_with_dormant_admins_count`
- Counts organisations with fewer than two users holding admin permissions (`can_manage_locations` and `can_manage_team` both true), no sign-in-recency filtering
- Adds full spec coverage: admin-count boundaries, permission edge cases, multi-org aggregation, S3 payload shape, Metrics API POST body, and API-failure resilience

### Why

- Organisations with fewer than two admins risk losing management access entirely if their sole admin leaves
- This is an account-health signal worth tracking daily alongside the existing dormant-admins metric
- Feeds the existing Tableau/metrics dashboard pipeline via the same S3 + Metrics API path as other account-health metrics

- [GW-2738](https://technologyprogramme.atlassian.net/browse/GW-2738)

### DIFF
--- a/lib/tasks/publish_orgs_with_less_than_two_admins_count.rake
+++ b/lib/tasks/publish_orgs_with_less_than_two_admins_count.rake
@@ -1,0 +1,11 @@
+require "logger"
+logger = Logger.new($stdout)
+
+namespace :metrics do
+  desc "Publish orgs with less than two admins count to S3 and post to metrics API"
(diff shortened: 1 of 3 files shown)


ethancane-digital-co (approved):
LGTM

camdesgov (approved):

PR #2708, opened 2026-08-10
GW-2645: Fix site logs timeout — MAC query + local dev migrations

ethancane-digital-co:
### What

Two changes to resolve the site logs MAC search timeout:

1. **Fix the MAC query** — Normalise the MAC address in Ruby before querying, so the SQL uses a plain `WHERE mac = ?` equality that can hit the existing index, instead of wrapping the column in `UPPER(REPLACE(...))` functions that prevented index usage.

2. **Fix local read replica setup** — Replace the stale `rr_schema.sql` (a manual DROP/CREATE snapshot that had drifted from production) with a migration runner that clones the logging API repo and runs the real Sequel migrations. This guarantees the local schema matches production, including the mac index needed to verify the query fix.

### Why

The MAC search on the site logs page was timing out because the query couldn't use the existing index on the `mac` column. The local development environment also couldn't reproduce or verify the fix because the read replica schema was missing the index entirely.

### Link to JIRA card

[GW-2645](https://technologyprogramme.atlassian.net/browse/GW-2645)


mysql/migrate.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -euo pipefail
+
+LOGGING_API_REPO="https://github.com/GovWifi/govwifi-logging-api.git"
```
leonaAtkins: not keen on hard coded urls, the repo may change, is there a way this could be passed in?  We may change org again, then this would break.

db/seeds.rb
```diff
@@ -56,11 +56,18 @@
 ip = FactoryBot.create(:ip, address: Faker::Internet.unique.public_ip_v4_address,
                             location: location_one,
                             created_at: Time.zone.now - 10.days)
+
+# Helper to format MAC to dashes (the standard for our app)
+def format_mac(mac)
+  mac.gsub(/[-:]/, "").upcase.scan(/../).join("-")
```
oisinmulvihill: Can you move this into one place that is shared with login_search_form.rb. That way its easy to maintain and test.

lib/gateways/sessions.rb
```diff
@@ -46,5 +44,14 @@ def self.organisation_usage_stats
         0
       end
     end
+
+  private
+
+    def normalize_mac(mac)
+      # Remove all non-alphanumeric characters and uppercase
+      clean = mac.gsub(/[^0-9A-Fa-f]/, "").upcase
```
oisinmulvihill: see earlier comment about this being in one place for maintainability

mysql/migrate.sh
oisinmulvihill: I second this, it should come from an environment variable. It could use a default value which is this

oisinmulvihill (changes requested):
See comments inline. The main thing is to have one place for MAC string sanitiser. Then use this everywhere instead of copy and pasting the regex and other actions on one line.

oisinmulvihill (approved):
Excellent, good job. Thanks for making the changes.

🆗🚢

PR #2736, opened 2026-09-21
Remove /status route

pixeltrix:
### What

Removes the `resources :status, only: %i[index]` line from routes.rb

### Why

It doesn't have a controller which gets logged by Sentry unlike routing errors.

**Note:**
I noticed that errors all return a 200 response instead of the original error code whilst writing a spec for this - is this intentional?


ethancane-digital-co:
@pixeltrix Thanks for creating this PR. Most appreciated.

However the root cause goes a little deeper than expected. I have created a separate PR found here:

https://github.com/GovWifi/govwifi-admin/pull/2738

Which fully resolves this issue and would not have been possible without your contribution.

PR #2737, opened 2026-09-22
Fix unvalidated choice param controlling template render in HelpController

iram-shehzadi:
## What was wrong

The "Contact us" support form on the admin site (`/help`) doesn't require you to be signed in and anyone can submit it, including automated bots and scripts, not just people using the actual website.

Behind the scenes, the form has a hidden field that tells the app which version of the page to show again if something's wrong with your submission (e.g. you forgot to fill in a box). The app trusted whatever value was in that hidden field completely, without checking it was actually one of the real options. Normally a person's browser only ever sends sensible values here, because that's all the visible form allows — but someone bypassed the real form and sent requests straight to the server with made-up, broken values instead.

This showed up in Sentry as a wave of crash errors on this form, including one attempt using `<%=7*7%>` in the email field and a well-known test string security scanners use to check if a website will run code that's typed into a form. Nothing was successfully broken into, but it confirmed someone was deliberately testing the form for weaknesses, not just users making typos.

## What this PR changes

The form now only ever shows one of the four real pages it's supposed to (technical support, admin account, user support, signed in), no matter what value is sent in that hidden field. If something unexpected is sent, it just falls back to a safe default page instead of trying to use it directly. This closes off the crashes and removes the opening that was being probed.

## How to see the difference

**Before this fix (on `master`):**
1. Open the support form, e.g. `/help/new/technical_support`
2. Open your browser's DevTools (right-click → Inspect) and find the hidden field named `support_form[choice]`
3. Change its value to something made up, like `banana`
4. Leave the message box empty and submit
5. Result: the site crashes with a server error page

**After this fix (this branch):**
- Repeat the exact same steps
- Result: the form just reloads with a normal "please fill this in" message - no crash

## Testing done

- Confirmed the normal, everyday use of the form still works exactly as before (fill it in, submit, get a confirmation).
- Confirmed a broken/invalid submission now fails gracefully instead of crashing.
- Directly replayed the three crash scenarios seen in Sentry (a made-up hidden field value, a broken character in that value, and a spoofed request header) and confirmed none of them cause an error anymore.


ethancane-digital-co (approved):
LGTM

